PR #167, opened 2016-03-29
Remove easter egg

kysel:
I think this type of text shouldn't be in public source files nor in inline documentation.


quigamdev:
You are absolutely right, we have to remove it. I asked author of the commit (exyi), but he told me he did not write this. Thanks.

PR #184, opened 2016-05-17
Support for Knockout extenders (client converters)

darilek:
Basic support for knockout extenders.
Steps to activate extender:

1. decorate specified property with `ClientExtenderAttribute`
2. create and register Knockout extender in javascript
3. use extender when declare Knockout binding ie. `data-bind='value: PropertyName.ExtenderName'`

This patch is primarily intended for component developers (Knockout wrappers). It should be easy to expand support to the DotVVM binding syntax


exyi:
Thank you, it seems to be pretty cool.
Maybe it may be useful to have more than one extender on a single property or to pass a parameter to the extender (not just [true](https://github.com/riganti/dotvvm/pull/184/files#diff-8259ae5d80e11312ec1132b3d2c325abR104)), so we can probably transfer the entire object in options like `{myExtender1: "parameter value", anotherExtender: true}` and the attribute should have a property for the parameter

I think we can merge it, @tomasherceg?

darilek:
I did not tried apply more then one extender on observable property

darilek:
I will try to prepare a patch which covers the points discussed above

tomasherceg:
Awesome. When it's ready, we can merge it and write some tests for this.

darilek:
Hello, I've prepared second patch which covers @exyi requests :-)

Small example:
```c#
// C# view model class
public class ViewModel 
{
   // atribute has parameters: name, parameter value (optional) and order (optional)
   [ClientExtender("extender1", "hello world", 2]
   [ClientExtender("extender2", true, 1]
   public string PropertyName { get; set; }
}
```
Client side serialializer has ability chain declared extenders by specified order
```js
// result of executed code on client side
this.PropertyName = ko.observable(value).extend({ extender2 : true}).extend({extender1 : "hello world"});
```

darilek:
I've added the first (very simple) client extender sample - password strength indicator

exyi:
Sorry it took so long time, but finally merged. Thank you.

Issue #176, opened 2016-05-12
Double postback prevention on a button


tomasherceg:

tomasherceg:
Will be solved in #283.

PR #189, opened 2016-06-08
added Disable on postback postback handler and helper property for button

exyi:
should solve #176


exyi:
Do you have an alternative solution for this?

djanosik:
Yes, #176 is solved by concurrency modes. And I think we need more generic solution that will allow to disable button, add CSS class, etc. when postback is in progress.

exyi:
On Thu, 2017-12-07 at 18:43 +0000, Dusan Janosik wrote:
> Yes, #176 is solved by concurrency modes.
It is solved only partially, you can't (easily) set a disable attribute
when the postback on **this** button is running.
> And I think we need more
> generic solution that will allow to disable button, add CSS class,
> etc. when postback is in progress.
I agree, maybe we could make a postback handler that will have a
"Attributes" and "CssClasses" property groups?

djanosik:
I am not sure. Would it be possible to set properties in postback handler? Setting attributes will not be good enough.
```
<Button class="whatever">
    <UpdateProgressHandler Button.Enabled="false" class-progress="true" />
</Button>
```

PR #199, opened 2016-06-28
renamed Globalize global variable to dotvvm_Globalize

exyi:


darilek:
It seems this PR is not contained in the RTM version. Will be it merged to the nearest maintain version?

exyi:
I just left it here for review and forgot about it, sorry about that. Merged now, hope package will be released soon.

darilek:
Thanks

PR #200, opened 2016-06-30
bundling and minification added

juratime:
bundling and minification using web.optimization

### DIFF
--- a/src/DotVVM.Compiler/App.config
+++ b/src/DotVVM.Compiler/App.config
@@ -21,6 +21,14 @@
         <assemblyIdentity name="Microsoft.CodeAnalysis.CSharp" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-1.1.0.0" newVersion="1.0.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-8.0.0.0" newVersion="8.0.0.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="WebGrease" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.5.2.14234" newVersion="1.5.2.14234" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
 </configuration>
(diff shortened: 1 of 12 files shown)


djanosik:
Interesting solution, but I don't think this PR should be merged. It's really bad idea to bundle and minify resources at runtime. Instead, we need to find a way to build bundles at compile time (MSBuild task? see [Bundler & Minifier](http://www.nuget.org/packages/BuildBundlerMinifier/)). It may require some changes to DotVVM resource management.

Also... WebGrease is a deprecated solution and I doubt it will ever be ported to .NET Core.

martinsura:
Bundling resources in runtime is nice feature in development mode. (Less, Typescript compile on the fly)

But for production is better @djanosik solution.

exyi:
@djanosik Dotvvm already contains infrastructure for bundled scripts, or similar processing of resources. However it is documented only at [ResourceManagement/BundlingResourceProcessor.cs](https://github.com/riganti/dotvvm/blob/12e5e60dab0c9a2d495f5e4d21ec0ad1a288378e/src/DotVVM.Framework/ResourceManagement/BundlingResourceProcessor.cs). I thought I had a sample for that, but I can't find it, it was probably lost when we were creating the new samples with UI tests :(. You just register which resources are in which bundle and when you reference at least one resource from the bundle, the entire bundle is loaded.

I also like compile-time processing of resources, in best case recompiled on save so I don't have to recompile the project to refresh one script, just save and reload page. WebEssentials could do that in past, now you have to use grunt or something I'm not familiar with :(

@juratime Could you please publish as separate assembly, in separate repository and as separate NuGet package, since I don't want to make DotVVM dependent on WebGrease and Antlr. Just PR some interfaces if you'd need them. Or ask for them, maybe they are already present ;).

djanosik:
@exyi Yeah, I know about that :) You can use the mentioned Bundler & Minifier. It's a [VS extension](https://visualstudiogallery.msdn.microsoft.com/9ec27da7-e24b-4d56-8064-fd7e88ac1c40) + [MSBuild task](http://www.nuget.org/packages/BuildBundlerMinifier/). We can do it in similar way.

tomasherceg:
We can bundle and minify resources at runtime provided we'll cache them accordingly, and I think that there are plenty of users who don't want to install any other extensions, use grunt/gulp or other tools, they just want the framework to handle it somehow. 
Ideally the runtime bundling will be a separate DotVVM package - something like DotVVM.Utils.RuntimeBundler.

On the other hand, there must be an option to give DotVVM the bundled and minified resources manually, e.g. using MSBuild or other tooling.

We won't merge this PR as is, but I'll go though it with @juratime and we'll modify it accordingly.

martinsura:
> and I think that there are plenty of users who don't want to install any other extensions, use grunt/gulp or other tools, they just want the framework to handle it somehow. 

+1 Exactly my usecase

PR #222, opened 2016-07-31
Customized parameter types

parmap:
Just check the solution please. By me, there wasn't need to create special wrapper and to have back compatibility, the parameter types is optional for route.


exyi:
PR #208 already solves this. I personally like much more the collection in DotvvmConfiguration than ServiceLocator, because when you'd want to extend it you'd have to get, add and register which is much more painful.

Issue #216, opened 2016-07-25
Authorization filters for custom presenters


CristianT:
My application has a global authorization filter like this:

`config.Runtime.GlobalFilters.Add(new AuthorizeAttribute());`

As a user I understand that the authorization applies to any request handled by dotvvm but that is not the case of custom presenters:

`config.RouteTable.Add("UpdateValue", "UpdateValue/{Id?}", null, null, () => new UpdateValuePresenter());`

The attribute Authorize does not work either.

```
    [Authorize]
    public class UpdateValuePresenter : IDotvvmPresenter
```

So I have to implement the authorization myself.
If this is the intended behavior at least a comment in the documentation would be nice.

tomasherceg:
Thanks for the feedback, it is a very good point. 
I think we should support the filters even on the presenters.

exyi:
Done in #261

PR #261, opened 2016-11-05
Refactored Action Filters

exyi:
* ActionFilterAttribute is splitted into interfaces - request filter, viewModel filter and command filter
* Now they also work when applied to IDotvvmPresenter and global request filters are applied to requests to any presenter not only the default one

Review it please.


src/DotVVM.Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
```diff
@@ -61,14 +63,28 @@ private bool TryParseGooglebotHashbangEscapedFragment(string queryString, out st
...
             }
             catch (DotvvmInterruptRequestExecutionException)
             {
                 // the response has already been generated, do nothing
             }
+            catch(Exception exception)
+            {
+                foreach (var f in filters)
+                {
+                    await f.OnPageExceptionAsync(context, exception);
```
djanosik: The method is now called also for `DotvvmControlException` and `DotvvmHttpException` rethrown in `DotvvmPresenter` (L66, L221). Is this intended?

src/DotVVM.Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
```diff
@@ -61,14 +63,28 @@ private bool TryParseGooglebotHashbangEscapedFragment(string queryString, out st
...
+                    if (context.IsCommandExceptionHandled) context.InterruptRequest();
```
djanosik: Shouldn't we check for `IsPageExceptionHandled` instead?

djanosik:
I like this PR, but we should probably consider more changes:

-  Rename `IRequestActionFilter` to `IResourceActionFilter` and methods to `OnResourceLoadingAsync`, etc. Custom presenters will not always be used to load pages.
- Remove `ActionFilterAttribute` base class from global action filters that are not meant to be used like attributes (eg. `AuthorizeFilterAttribute`).
- Update `Authorize` filters to work on custom presenters (it will fix #216).

What do you think?

src/DotVVM.Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
exyi: DotvvmPresenter rethrows only `HttpException`, `DotvvmControlException` should be handled like any other. I'll fix it.

exyi:
@djanosik I have just fixed the little issues you reported, thank you for review. 
I would not rename the `IRequestActionFilter` since it contains filter applied per request. The `Page...` should also stay as `OnPageExceptionAsync` was there before and I'd not change it. I also think that everyone will understand what `Page` means.

djanosik:
@exyi I see, it'd be a breaking change. That's unfortunate, the names feel a bit confusing. Couldn't we keep `OnPageExceptionAsync` on `ActionFilterAttribute` as `Obsolete` for some time? Maybe it's not worth it.

exyi:
@djanosik It is not a solution because it is an interface user will implement :(. And I actually don't think that the name is confusing, there is only better alternative.

djanosik:
@exyi Ok then. :)

exyi:
@djanosik Authorize attribute should be working correctly on `IDotvvmPresenter`s.

PR #271, opened 2016-11-24
Refactored Resource Management

exyi:
* Splitted resource type (script/style) and resource location type (local file, local embedded, remote url, local generated)
* All local resources are served by one middleware and have nice & cache-friendly urls (`dotvvmResource/{hash}/{resourceName}`)
* Supports loading fallback - when one location fails, then it can be loaded from another one
* When url is remote and local location is also specified, `integrity` attribute is rendered ([see Subresource Integrity at MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity))
* Added few extensibility interfaces, so you can change the way how local resource are loaded - you can even automatically publish them to CDN or IPFS and load them from it ;)

### Refactoring overview

I have changed the resource hierarchy a bit:
* `IResource` - just lighter interface than the old `ResourceBase` - has dependencies and Render method
    - `ResourceBase` - abstract implementation of that interface
* `ILinkResource : IResource` - adds GetLocations, used to obtain url or content from the resource
    - `LinkResourceBase` - abstract implementation, has utility for rendering resource, integrity hashes
* ScriptResource/StylesheetResource/NullResource - similar as before

And added new class hierarchy for representing resource location:
* `IResourceLocation` - GetUrl()
* `ILocalResourceLocation : IResourceLocation` - LoadResource() - gets the content
* `RemoteResourceLocation : IResourceLocation` - loaded from the url
* `EmbededResourceLocation: ILocalResourceLocation` - loaded from assembly embeded resource
* `LocalFileResourceLocation : ILocalResourceLocation` - loaded from file

### Backwards compatibility
This is the main problem - the resource registration API is not compatible, but you only have to apply cosmetic changes to the code, no huge architecture refactoring is required on your side. Please let me know in comments how problematic would it be for you.

### Content Security Policy?
Originally I wanted to add automatic CSP, but it is not that simple since resources can be loaded dynamically after postback. We could enforce that everything is loaded from our server and you can add your own locations on your own, but I'm not sure if it is good idea to enforce. Any ideas? @ridercz? Or another security guy?

### Images?
We have nice cache-friendly content serving mechanism - what about to use it for images (video, audio and other content)? Maybe we could use special url that will point to the resource, like `resource://my-image` or `~~my-image` and it will be translated to resource's location.

We could also add virtual resources for entire image folder to avoid registration hell - could be used like `images:delete-icon.png`.

Do you have any other use cases/ideas that the new resource system could support? Let me know until we release it :)

EDIT:
We have currently a problem with source maps. They are looked up in relative path to the script file, so they will not be found in the /dotvvmResource/{hash} directory. Maybe we could turn off our resource middleware when run on debug.


darilek:
@exyi nice work. I will close PR #211 as obsolete :-) I have some questions/suggestions:

- special url for image resources? YES! ... and ability for dynamic resource name - somethink  like `resource://{ViewModel.PropertyName}` would be great :-)
- how about some sort of group definitions for scripts, styles etc (for local resources)? instead of refrencing each required resource in the code, you would have wrote only resource group name and framework will merge resources in group into one stream (with hash). May be grouping is needless due to resource dependencies, but merging into one file would be nice

exyi:
@darilek I'm glad you like it :)
* I'm not sure if the resource name could be dynamic, as the url can't be resolved on client side, server rendered binding should be possible, although I'd not expect the binding for attribute part very soon.
* I think that this issue is already solvable by
    - Dependencies, as you said. Also any control that you use can have dependencies, so you don't have to care about scripts required by the controls
    - `IResourceProcessor`s - this interface is present for a long time. It can replace any resource with something else. We have `BundlingResourceProcessor` that replaces specified resources by links to the bundle where they are located.
We can not create the bundles, but I don't want to have this functionality in the FW, we can maybe create another package for that. I would be cool to automatically bundle all required resources into one, just by saying `config.UseAutomaticBundling()`.

I have just realized that we have a problem with source maps ;(

djanosik:
I don't use DotVVM Resource Management anywhere, but I've tried to review the PR. Just a few notes:

- I would prefer to have shortcuts for registering resources. Like `.RegisterLocalFileStylesheet(...)`. But we can surely leave this to users to implement whatever extensions they want.
- I think it isn't a good idea to auto-generate the CSP header. There is an [NWebsec](https://github.com/NWebsec/NWebsec) middleware that makes it easy enough. It reminds me that DotVVM is not CSP compliant. I will file an issue.
- I have no idea how to handle source maps. But I think you can't disable RM middleware in debug mode, because you still need it to load embedded resources.
    - Can we accept that source maps will not work with embedded resources?
    - Can we disable RM middleware just for local file resources?
- I would like to see OOB solution for minification and bundling. It can live in other project.
    - It should not handle precompilation of `less`, `TypeScript`, etc.
    - It should bundle all resources into one file. Maybe we can skip this on HTTP/2.
    - It should minify all not-yet-minified resources.
    - It should not do anything in debug mode.

exyi:
Btw, why we need CDN fallbacks? It looks like some CDNs are sometimes blocked in China and maybe in other countries with censored internet: https://news.ycombinator.com/item?id=13076295

CZ is coming soon, as all gambling pages are using jQuery from google CDN.

exyi:
I have just implemented few improvements for development and debugging of the scripts:
* Source maps are loaded when they are in the same directory as the script/style.
* Resources refreshed on each request, so you can edit them without app restart.
* You can specify embedded resource's real path, so both above features work on them. 🍻 

These all apply only when `Configuration.Debug = true` so it will not break your application's performance nor security.

The source maps have to be in the same directory as the resource, because they are located by the resource hash a and filename. When resource is at `/resource/<hash>/<name>` the source map is requested at `/resource/<hash>/Whatever.js.map`. Dotvvm finds the resource file location by the hash and then finds `.map` file in the scripts folder.

djanosik:
@exyi Looks ok, but we should try it on real project before 1.1 goes out :) When specifying embedded resource's real path, am I supposed to use relative or absolute path? We should describe what path is expected in XML comments.

- `debugFileName` param in `EmbededResourceLocation` should be `debugFilePath`
- why are there multiple classes / interfaces in one file?

darilek:
For embedded resource real path: will it work also for resources, compiled in library project (control repository) and referenced in main solution (web) project?

exyi:
@darilek of course, that is the reason why am I implementing it ;) You only have to specify path on your filesystem.

@djanosik There are multiple classes in one file because I was lazy and it makes sense for me in some cases, but I'll fix as there is too many people that don't agree with me :)

PR #289, opened 2016-12-19
Error page font size changes

darilek:
I did small changes to error page font sizes. IMHO they brings better readability and shows more text on page without scrolling


exyi:
You are probably right that it looks better, thank you.

PR #339, opened 2017-03-17
Literal server rendering renders both value and binding

exyi:
I'm not sure, if it is always ok, but in a lot of cases it's better to render both, because the value can be updated after it was rendered.


djanosik:
It would also be nice to render both in client rendering when possible.

exyi:
This is too old to merge, however we'll discuss how it's possible to render both the binding and value. Unfortunately it's not possible to switch the rendering right away, as it's a quite ugly breaking change :/

PR #358, opened 2017-04-28
Implemented support for recursive markup controls

exyi:
* view compiler, resolver, ... have to return control's outer API (type, data context) and *lazy* compiled builder to prevent stack overflow exceptions.
* added sample tree control with postbacks, binding to root, bindings to parent


src/DotVVM.Framework/Compilation/DefaultControlBuilderFactory.cs
```diff
@@ -68,18 +68,30 @@ private IControlBuilder CreateControlBuilder(MarkupFile file)
                 var namespaceName = GetNamespaceFromFileName(file.FileName, file.LastWriteDateTimeUtc);
                 var assemblyName = namespaceName;
                 var className = GetClassFromFileName(file.FileName) + "ControlBuilder";
-                try
-                {
-                    return ViewCompilerFactory().CompileView(file.ContentsReaderFactory(), file.FileName, assemblyName, namespaceName, className);
-                }
-                catch (DotvvmCompilationException ex)
+                void editCompilationException(DotvvmCompilationException ex)
```
tomasherceg: Local function in the middle of a function? Wouldn't it be better to leave there at the end of the method?

src/DotVVM.Framework/Compilation/DefaultViewCompiler.cs
```diff
@@ -35,7 +35,7 @@ public DefaultViewCompiler(DotvvmConfiguration configuration)
         /// <summary>
         /// Compiles the view and returns a function that can be invoked repeatedly. The function builds full control tree and activates the page.
         /// </summary>
-        public virtual CSharpCompilation CompileView(string sourceCode, string fileName, CSharpCompilation compilation, string namespaceName, string className)
+        public virtual (ControlBuilderDescriptor, Func<CSharpCompilation>) CompileView(string sourceCode, string fileName, CSharpCompilation compilation, string namespaceName, string className)
```
tomasherceg: I don't have a problem with tuples returned inside non public methods, but this method is public and virtual. 
We should return an object here, because someone may need to override this function and return an object with additional properties (if the method returns object, he can return something that extends it). Using tuple in virtual methods breaks the extensibility totally.

src/DotVVM.Framework/Compilation/IControlBuilderFactory.cs
```diff
@@ -1,9 +1,9 @@
+using System;
+
 namespace DotVVM.Framework.Compilation
 {
     public interface IControlBuilderFactory
     {
-        
-        IControlBuilder GetControlBuilder(string virtualPath);
-
+        (ControlBuilderDescriptor descriptor, Lazy<IControlBuilder> builder) GetControlBuilder(string virtualPath);
```
tomasherceg: Same situation as with the virtual methods. This breaks the open/closed principle.

src/DotVVM.Framework/Compilation/DefaultViewCompiler.cs
exyi: If you'd ever need additional properties, use the ControlBuilderDescriptor or attached property on CSharpCompilation. I don't see any reason to do that, but you still can, the function of the tuple is only to distinguish two steps of compilation - first phase, when tree is parsed and direstives read and second phase when the view is compiled. Only in case when you'd like to add additional state, it could potentially help you - but it wont, because the interface's consumer could not use that possibility anyway. And in case you'd also implement the consumer, you can add method to the interface or add another interface, as you would basically bypass this implementation.

src/DotVVM.Framework/Compilation/IControlBuilderFactory.cs
exyi: I don't agree with open/closed principle

Issue #224, opened 2016-08-03
Route Group


parmap:
Very nice feature is route grouping. The advantage of this feature is ability to define e.g. filters in one place, which will apply for all routes inside group. Nested groups would be fine. Of course route settings should override group settings.

What could be able to specificy: filters, route prefix, view virtual path

exyi:
You mean something like:
```C#
routes.AddGroup(name: "Admin", prefix: "admin", virtualPath: "Views/Administration", content: adminRoutes =>
{
    adminRoutes.Add("Users", "users", "Users.dothtml"); // name = Admin_Users, path = "admin/users", virtualPath = "Views/Administration/Users.dothtml"
});
```
I agree it would be nice.

PR #363, opened 2017-06-03
Added route groups

Rumbeer:
Route groups can be added to a RouteTable, described in https://github.com/riganti/dotvvm/issues/224. Group has its prefixes for route name, url and virtual path. The group can be retrieved and works as a RouteTable.


exyi:
Could you just add you email address (433545@mail.muni.cz) to github, so we can see, that the commit was made by you? The settings it at https://github.com/settings/emails

Rumbeer:
Thank you for the reply, the email should be added now.

Issue #369, opened 2017-06-20
Gather timestamps during the execution of the request


tomasherceg:
We should add a `TraceData` Dictionary<string,object>in the `DotvvmRequestContext` class and DotVVM should store how long each phase of the request processing took:

* BeginRequest
* RouteMatched
* ViewInitialized
* ViewModelCreated
* InitCompleted
* ViewModelDeserialized
* LoadCompleted
* CommandExecuted
* PreRenderCompleted
* ViewModelSerialized
* OutputRendered
* EndRequest

Ideally, we should use `System.Diagnostics.Stopwatch` singleton object. 

We can use this data in the VS Extension diagnostics window (it is being planned right now), send them to the Application Insights or another service.

Issue #370, opened 2017-06-20
DotVVM.Tracing.ApplicationInsights


tomasherceg:
We should create a new Nuget package which reports exceptions and request information data to the Application Insights. 

It would be registered in the `UseDotvvm` method, for example this way:

```
var dotvvmConfiguration = app.UseDotVVM<DotvvmStartup>(applicationPhysicalPath, options: options =>
{
    options.AddApplicationInsightsTracing()
});
```

1) This would register an implementation of `IRequestTracingReporter` in the `DotvvmConfiguration.Runtime.Tracing.Reporters` collection which we can add. 

When the requst ends, DotVVM should call the `TraceRequestAsync(context)` method on all configured reporters. 

2) It would also report all exceptions from commands or requests which occured during the request execution.

tomasherceg:
@Rumbeer 
I like the implementation, but we should do several modifications to this:

In **OWIN**, the registration is simple: 
`options.AddApplicationInsightsTracing();`

In **ASP.NET Core**, the `TelemetryClient` is registered separately, and the reporter is registered in the `Configure` phase,because we don't have the `DotvvmConfiguration` object in the `ConfigureServices` method yet.
I think that we can add the `options.AddApplicationInsightsTracing` method to ASP.NET Core too. It should only register the `TelemetryClient` and `ApplicationInsightsReporter` in the `IServiceProvider`.
The `UseDotVVM` method shoul look for all implementations of `IRequestTracingReporter` and add them to the `Reporters` collection when the `DotvvmConfiguration` object is ready.

And finally, to make things consistent, I'd register the `ApplicationInsightsReporter` in DI in the OWIN version too.

tomasherceg:
TBD

- [ ] Test it in several real projects.
- [ ] Check that the documentation is current.
- [ ] Check that is is merged to 1.1 too.

quigamdev:
Already solved. The implementation was moved to https://github.com/riganti/dotvvm-tracing

PR #372, opened 2017-06-21
Gather timestamps during the execution of the request

Rumbeer:
Added gathering of timestamps during the exectuion of the request described in https://github.com/riganti/dotvvm/issues/369. It uses the IStopwatch interface to register a System.Diagnostics.Stopwatch singleton object.

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
@@ -82,7 +82,7 @@ namespace DotVVM.Framework.Hosting
         {
             var stopwatch = context.Configuration.ServiceLocator.GetService<IStopwatch>();
 
-            long lastStopwatchState = AddTraceData(stopwatch.GetElapsedMiliseconds(), "BeginRequest", context, stopwatch);
+            long lastStopwatchState = AddTraceData(stopwatch.GetElapsedMiliseconds(), RequestTracingConstants.BeginRequest, context, stopwatch);
             if (context.HttpContext.Request.Method != "GET" && context.HttpContext.Request.Method != "POST")
             {
                 // unknown HTTP method
@@ -100,7 +100,7 @@ namespace DotVVM.Framework.Hosting
             var page = DotvvmViewBuilder.BuildView(context);
             page.SetValue(Internal.RequestContextProperty, context);
             context.View = page;
-            lastStopwatchState = AddTraceData(lastStopwatchState, "ViewInitialized", context, stopwatch);
+            lastStopwatchState = AddTraceData(lastStopwatchState, RequestTracingConstants.ViewInitialized, context, stopwatch);
 
             // locate and create the view model
             context.ViewModel = ViewModelLoader.InitializeViewModel(context, page);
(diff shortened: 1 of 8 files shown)


src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -249,6 +262,14 @@ public class DotvvmPresenter : IDotvvmPresenter
             }
         }
 
+        private long addTraceData(long lastStopwatchState, string processName, IDotvvmRequestContext context)
```
tomasherceg: Please rename the method to `AddTraceData`.
And the rename the `processName` to `eventName`.

src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -249,6 +262,14 @@ public class DotvvmPresenter : IDotvvmPresenter
...
+        {
+            var stopwatch = context.Configuration.ServiceLocator.GetService<IStopwatch>();
```
tomasherceg: We should get the stopwatch instance in the constructor so we do not call `GetService` for every traced event.

src/DotVVM.Framework/Hosting/DotvvmRequestContext.cs
```diff
@@ -77,6 +77,11 @@ public class DotvvmRequestContext : IDotvvmRequestContext
         /// </summary>
         public ModelState ModelState { get; private set; }
 
+        /// <summary>
+        /// Gets the data gathered during execution of the request.
+        /// </summary>
+        public Dictionary<string, object> TraceData { get; set; }
```
tomasherceg: Make the setter private.

Rumbeer:
Added commit with requested changes (AddTraceData gets instance of IStopwatch through constructor, rename addTraceData to AddTraceData and processName to eventName, the TraceData setter is now private)

Rumbeer:
Added constants containing event names used during exection of a request.

exyi:
Thank you for the PR, I like this feature but there are few things I’d do a bit differently:
• I don’t understand what is the purpose of IStopwatch interface. Do you think it would be a good idea to implement custom clock?
• Instead, I’d like to inject a tracer, that would collect these events and log them somewhere.
• I don’t like the another property on an already bulky IDotvvmRequestContext. What about to make a request scoped service or a property on HttpRequest (in the Items dictionary)?
• Why is the event id a string? IMO it would be much better to have an enum for them.

tomasherceg:
That's mainly my fault because I worked with @Rumbeer on this.

1) I wanted an abstraction over Stopwatch,. mainly because registering Stopwatch instance in the service locator is not the best idea (any other framework can do the same and they may need another lifetime of the instance which would break us).

2) That's solved in another issue, we are doing the Application Insights tracer. https://github.com/riganti/dotvvm/issues/370 
The purpose of this is only to gather data for the further processing. The tracer will be able to get this collection and do whatever it wants with it. And you'll be able to have multiple tracers.

3) I don't like it either but we should review everything that is on the context and decide what to keep and what to move elsewhere. I am sure we will find more properties that should not be there. 

4) I'd like to use this collection for more things in the future. The enum would be changed very soon. 
And the application developers may want to add their own data in this collection, so the tracer can attach them to the HTTP request information. 

I have merged this PR so we can continue working on the App Insights integration.

Issue #151, opened 2016-02-19
Markup controls packed into dlls.


exyi:
Currenty markup controls can only be a file, it should be possible to include in a dll (as a resource?)

exyi:
resolved in #373

PR #373, opened 2017-06-24
Embedded resource controls

JozefPastucha:
- EmbeddedMarkupFileLoader for loading markup files from embedded resources

- AggregateMarkupFileLoader, so it should be possible to add new markup file loaders.

- Issue #151


src/DotVVM.Framework/Hosting/EmbeddedMarkupFileLoader.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Gets the markup file for the specified virtual path.
+        /// </summary>
+        public MarkupFile GetMarkup(DotvvmConfiguration configuration, string virtualPath)
+        {
+            if (virtualPath.IndexOf('.') == -1 || virtualPath.IndexOf('.') == 0)
+            {
+                return null;
+            }
+
+            string assemblyName = virtualPath.Substring(0, virtualPath.IndexOf("."));
```
tomasherceg: This won't work when the assembly name contains dots. You are looking for the first dot but actually you cannot know what is the name of the assembly. 
I think that the better strategy will be to look for all assemblies loaded in the current AppDomain and try to find the resource in all assemblies where the embedded resource name starts with the assembly name.
If you find multiple resources which satisfy the condition, an exception should be thrown.

src/DotVVM.Samples.Common/DotvvmStartup.cs
```diff
@@ -94,6 +94,7 @@ private static void AddControls(DotvvmConfiguration config)
             config.Markup.AddMarkupControl("sample", "Localization_Control", "Views/FeatureSamples/Localization/Localization_Control.dotcontrol");
             config.Markup.AddMarkupControl("sample", "ControlCommandBinding", "Views/FeatureSamples/MarkupControl/ControlCommandBinding.dotcontrol");
             config.Markup.AddMarkupControl("sample", "ControlValueBindingWithCommand", "Views/FeatureSamples/MarkupControl/ControlValueBindingWithCommand.dotcontrol");
+            config.Markup.AddMarkupControl("sample", "EmbeddedResourceControlsButton", "Views/FeatureSamples/EmbeddedResourceControls/MyButton.dotcontrol");
```
tomasherceg: This actually loads the control from the filesystem. The purpose of this feature is to load the control from the embedded resource.

src/EmbeddedResourceControls/EmbeddedResourceControls.csproj
```diff
@@ -0,0 +1,49 @@
...
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="MyButton.dotcontrol" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+</Project>
```
exyi: I'm afraid, that this project can't be compatible with the AspNetCore samples, could you please replace it with .NET Core version?

src/DotVVM.Framework/Hosting/EmbeddedMarkupFileLoader.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// Gets the markup file for the specified virtual path.
+        /// </summary>
+        public MarkupFile GetMarkup(DotvvmConfiguration configuration, string virtualPath)
+        {
+            if (!virtualPath.StartsWith("embedded://"))
+            {
+                return null;
+            }
+
+            string resourceName = virtualPath.Remove(0, 11);
```
exyi: Just a tip: `11` is a magic constant and it is not very obvious what does it mean and you had to count the letters which is annoying and it's easy to make a mistake. So although it is longer I prefer to use `"embedded://".Length`, but you can just copy-paste the string instead of counting characters ;)

PR #377, opened 2017-06-27
Add IRefreshableGridViewDataSet.OnloadingDataAsync

ENgateman:
According to @martindybal, IRefreshableGridViewDataSet seriously lacks an asychronous version of the OnLoadingData property because it is often used for accessing databases. 

He asked me to implement this new OnLoadingDataAsync property and implement it fast.  It seems to be rather urgent.

There is just one little problem. I have very limited idea how should I approach this problem as my experience with both async-await and databases is well... limited... to a point that its very existence is arguable. 

The code I wrote is most likely rubbish, please advise.


src/DotVVM.Core/Controls/GridViewDataSet.cs
```diff
@@ -194,94 +293,44 @@ protected virtual void FillDataSet(GridViewDataSetLoadedData data)
...
-        /// Gets or sets the value of PrimaryKeyPropertyName property for the row that is currently edited.
-        /// </summary>
-        [Bind(Direction.None)]
-        [Obsolete("Use RowEditOptions.EditRowId instead. This property will be removed in future versions.")]
-        public object EditRowId { get => RowEditOptions.EditRowId; set => RowEditOptions.EditRowId = value; }
+        private async Task<GridViewDataSetLoadedData<T>> LoadData()
+        {
+            if (OnLoadingData != null && OnLoadingDataAsync != null)
+            {
+                //TODO: What type of exception should be here?
```
tomasherceg: No exception should be here - neither of the delegates is required.

src/DotVVM.Samples.Common/ViewModels/ControlSamples/DataPager/DataPagerViewModel.cs
```diff
@@ -11,7 +11,7 @@ public class DataPagerViewModel : DotvvmViewModelBase
         public GridViewDataSet<Data> DataSet { get; set; }
         public override Task Init()
         {
-            DataSet = GridViewDataSet.Create(GetData, pageSize: 3);
```
tomasherceg: Do not delete the previous working use case, it should work too. We need to cover both options with the test.

src/DotVVM.Core/Controls/GridViewDataSet.cs
```diff
@@ -1,25 +1,67 @@
...
     public class GridViewDataSet<T> : IGridViewDataSet
     {
-
         public GridViewDataSet()
         {
             IsRefreshRequired = true;
         }
 
         /// <summary>
+        /// Gets or sets the value of PrimaryKeyPropertyName property for the row that is currently edited.
```
tomasherceg: For version 1.2, we can safely remove the obsolete members in this class.

src/DotVVM.Core/Controls/GridViewDataSet.cs
ENgateman: But if they are both set at once? Which one should be used then?

tomasherceg: Oh, sorry, my bad. In this case. We don't have anything better than DotvvmHttpException for this situation.

ENgateman: Well that one just happens to be in DotVVM.Framework and therefore I can't reference it in DotVVM.Core without creating a circular dependency.

Maybe it wouldn't be such a bad idea to move DotvvmHttpException to DotVVM.Core.

src/DotVVM.Core/Controls/GridViewDataSet.cs
ENgateman: Sure, I'll do that in master after I merge these changes into 1.1.0.

PR #382, opened 2017-07-14
Application Insights request filtering

Rumbeer:
Added implementation of ITelemetryProcessor, that filters requests tracked by application insights.


exyi:
moved to #384

PR #384, opened 2017-07-19
Added request tracing and MiniProfiler tracer

adamjez:
Reworked IRequestTracingReporter to IRequestTracing with TraceEvent method which is called when that event happen not at the end of every request.

Added invocation of IConfigureOptions<DotvvmConfiguration> to be able to register services to DI and register dotvvm controls or filters in one method call e.g. AddMiniProfilerEventTracing. Method Configure in IConfigureOptions<DotvvmConfiguration> services is called after default DotvvmConfiguration is created and before user configuration.

Added projects MiniProfiler.AspNetCore and MiniProfiler.Owin to provide integration with MiniProfiler library (https://github.com/MiniProfiler/dotnet).
Reworked DotVVM.Tracing.ApplicationInsight to support IRequestTracing and added dotvvm control for appinsight tracing javascript

Added sample projects to show usage of MiniProfiler intergration for AspNetCore and Owin.

This pull request contains  pull request #382


src/DotVVM.Tracing.ApplicationInsights/ApplicationInsightsJavaScript.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (!telemetryConfiguration.DisableTelemetry && 
+                !string.IsNullOrEmpty(telemetryConfiguration.InstrumentationKey))
+            {
+                string additionalJS = string.Empty;
+                IIdentity identity = httpContextAccessor?.HttpContext.User?.Identity;
+                if (enableAuthSnippet &&
+                    identity != null &&
+                    identity.IsAuthenticated)
+                {
+                    string escapedUserName = identity.Name.Replace("\\", "\\\\");
```
exyi: I think, that this will not replace everything malicious, for example `"` or `'` will get propagated to the script and cause problems. It would be best to use `JsonConvert.ToString(identity.Name)` for string escaping.

src/DotVVM.Tracing.ApplicationInsights/ApplicationInsightsJavaScript.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.telemetryConfiguration = telemetryConfiguration;
+            this.httpContextAccessor = httpContextAccessor;
+            this.enableAuthSnippet = serviceOptions.Value.EnableAuthenticationTrackingJavaScript;
+        }
+
+        protected override void RenderControl(IHtmlWriter writer, IDotvvmRequestContext context)
+        {
+            if (!telemetryConfiguration.DisableTelemetry && 
+                !string.IsNullOrEmpty(telemetryConfiguration.InstrumentationKey))
```
exyi: And what about to disable the telemetry is the DoNotTrack header is present? It will get blocked anyway by any ad blocker or by Privacy Badger and it will look better if the site will appear tracker-free for these people.

src/DotVVM.Framework/Runtime/Tracing/RequestTracingExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using DotVVM.Framework.Hosting;
+
+namespace DotVVM.Framework.Runtime.Tracing
+{
+    public static class RequestTracingExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
adamjez: Ok, I didn't know that.

src/DotVVM.Samples.MiniProfiler.Owin/DotVVM.Samples.MiniProfiler.Owin.csproj
```diff
@@ -0,0 +1,174 @@
...
+    </VisualStudio>
+  </ProjectExtensions>
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+</Project>
```
adamjez: It would be better to convert old project system in all Owin samples at once. BasicSamples, ApplicationInsight and MiniProfiler.

src/DotVVM.Tracing.ApplicationInsights/ApplicationInsightsJavaScript.cs
adamjez: good idea, I will implement that

src/DotVVM.Tracing.ApplicationInsights/ApplicationInsightsJavaScript.cs
adamjez: I will replace it with [applicationinsight-aspnetcore](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/v2.1.1/src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs#L67) implementation which in version 2.1.1 is using JavaScriptEncoder.

Issue #315, opened 2017-01-30
Better error info for "Invalid command invocation"


exyi:
The FindCommand procedure usually fail because one of DataContext path, validation path or binding id gets wrong, so listing all command on the page sorted by simmilarity would help a lot
command on the page sorted by simmilarity would help a lot. The error also occurs when command inside Repeater is invoked but the collection is not populated, so stating "Nothing was found in found inside specified DataContext, check if ViewModel is populated" could help.

exyi:
resolved by #386, but if you'd have more ideas to make this error message more useful, feel free to comment.

PR #386, opened 2017-07-20
Better exception information on invalid command invocation

Rumbeer:
Added properties with information about commands on the page to exception thrown when invoking invalid command. Properties contain key which describes the group of bindings and command bindings that did not match path, binding id or validation path with information why the validation failed.

Problem is described in https://github.com/riganti/dotvvm/issues/315.

invalid-command

### DIFF
--- a/src/DotVVM.Framework/Runtime/Commands/EventValidator.cs
+++ b/src/DotVVM.Framework/Runtime/Commands/EventValidator.cs
@@ -71,7 +71,8 @@ namespace DotVVM.Framework.Runtime.Commands
                     // find bindings of current control
                     var bindings = control.GetAllBindings()
                         .Where(b => b.Value is CommandBindingExpression);
-                    List<string> exceptionPropertyKeys = new List<string>();
+                    List<string> wrongExceptionPropertyKeys = new List<string>();
+                    List<string> correctExceptionPropertyKeys = new List<string>();
                     foreach (var binding in bindings)
                     {
                         StringBuilder infoMessage = new StringBuilder();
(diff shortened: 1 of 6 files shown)


src/DotVVM.Framework/Runtime/Commands/EventValidator.cs
```diff
@@ -99,54 +146,121 @@ private FindBindingResult FindControlCommandBinding(string[] path, string comman
...
 
             return new FindBindingResult
             {
-                Property = resultProperty,
+                ErrorMessage = bindingInPath ? errorMessage : "Nothing was found in found inside specified DataContext, check if ViewModel is populated",
+                CandidateBindings = candidateBindings,
                 Binding = resultBinding,
-                Control = resultControl
+                Control = resultControl,
+                Property = resultProperty
```
exyi: It would be much better to deduplicate this huge piece of to a separate function.

Rumbeer: Can I move FindControlCommandBinding and FindCommandBinding to one and give them a bool parametr for choosing between them?

Rumbeer:
Would this error page be good enough?

invalid-command-invocation

exyi:
Yes, I think that now it's quite clear what is wrong with the bindings, thanks.

Maybe it would be a bit nicer if the header would include what is same in the binding, not only what is different, like: "Command bindings with the same data context and validation path but different binding id: "
It would also be better show the command binding list separated by new lines instead of a semicolon, in the form of html `ul > li`.

src/DotVVM.Framework/Runtime/Commands/EventValidator.cs
exyi: That's a possibility, but I think that the `FindControlCommandBinding` method with `targetControl` parameter set to null will behave the same as `FindCommandBinding` method ;) Or maybe just a minor tweak would be required.

Rumbeer:
The bindings are now showed with both correct and wrong properties and individual bindings are in html list.

Issue #412, opened 2017-07-26
UpdateProgress control - dont show on short postbacks


MichalTichy:
UpdateProgress should have option to set delay to before its shown

PR #413, opened 2017-07-27
Added delay attribute to UpdateProgress control

Rumbeer:
Added delay attribute to UpdtateProgress control

https://github.com/riganti/dotvvm/issues/412


src/DotVVM.Framework/Controls/UpdateProgress.cs
```diff
@@ -21,6 +36,11 @@ protected override void AddAttributesToRender(IHtmlWriter writer, IDotvvmRequest
         {
             writer.AddKnockoutDataBind("dotvvm-UpdateProgress-Visible", "true");
 
+            if (Delay != 0)
```
adamjez: What if the delay is a negative number? https://stackoverflow.com/questions/8430966/is-calling-settimeout-with-a-negative-delay-ok

exyi: Then the browser minimum time ​(~4ms) will be used, I actually think that having the possibility to force min time delay by `-1` is a good feature. And IE 8 will crash earlier :P

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -872,23 +872,57 @@ class DotVVM {
...
+                var running = false;
+
+                var show = () => {
+                    running = true;
+                    if (delay == null) {
+                        element.style.display = "";
+                    } else {
+                        timeout = setTimeout(e => {
+                            element.style.display = "";
+                        }, delay);
```
exyi: By the way, the property could be bindable, if you'd read the "data-delay" attribute on each run and render an `attr` binding in case there is a binding (which can be simply done by assigning the `GetValueRaw(DelayProperty)` to `this.Attributes["data-delay"]`). But I don't see any use case for this...

src/DotVVM.Framework/Controls/UpdateProgress.cs
adamjez: Ok, But that's somehow strange that if you specify 0 ms delay you get **NO** delay and if you specify negative delay, for example -1, you get the minimum delay (**4 ms**).

src/DotVVM.Framework/Controls/UpdateProgress.cs
```diff
@@ -13,6 +14,20 @@ namespace DotVVM.Framework.Controls
     /// </summary>
     public class UpdateProgress : HtmlGenericControl
     {
+
+        /// <summary>
+        /// Gets or sets the delay after which the content inside UpdateProgress control is shown
```
adamjez: There should be hinted in which a time unit (s, **ms**, ns) is the value expected

src/DotVVM.Framework/Controls/UpdateProgress.cs
djanosik: I think we should not abuse negative values to specify a default value.

MichalTichy: We should throw exception when invalid (negative number) value is entered.

src/DotVVM.Framework/Controls/UpdateProgress.cs
Rumbeer: Good point I will add it.

Issue #311, opened 2017-01-23
CheckBox - CheckedValue type can differ from CheckedItems list type producing odd results


kiraacorsac:
Given `Strings` is a list of `string`  and `Number` an `int`, I can have
```
<dot:CheckBox Text="{value: Name}" CheckedItems="{value: Strings}" CheckedValue="{value: Number}" />
```
creating nonsensical results in less trivial examples.

I would expect an exception.

exyi:
There is a control usage validator that should now throw a compile error: https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Controls/CheckBox.cs#L119

PR #415, opened 2017-07-27
CheckedValue and CheckedItem types validation in checkable controls

Rumbeer:
`CheckedItem` or `CheckedItems` must now have matching types with `CheckedValue` for `RadioButton` and `Checkbox` or an exception is thrown.

Issue https://github.com/riganti/dotvvm/issues/311


src/DotVVM.Framework/Controls/CheckBox.cs
```diff
@@ -23,7 +24,17 @@ public IEnumerable CheckedItems
...
+
+        protected internal override void OnLoad(IDotvvmRequestContext context)
+        {
+            if (CheckedItems != null && CheckedValue != null && 
+                CheckedItems.GetType().GetGenericArguments()[0] != CheckedValue.GetType())
+            {
+                throw new DotvvmControlException(this,
+                    $"Type of items in CheckedItems \'{CheckedItems.GetType().GetGenericArguments()[0].FullName}\' must be same as CheckedValue type \'{CheckedValue.GetType().FullName}\'.");
+            }
+            base.OnLoad(context);
```
exyi: I'm sorry, but bindings have a `ResultType` property that is determined compile time and it should be used to determine if the value types are matching - this will not work in client-side templates, in case the value is null for another reason and will cause unnecessary binding evaluation (5 of them actually).

Another problem is that checking only the generic arguments is not correct at all - it will break on pure arrays, types implementing only `System.Collections.IEnumerable` and types that implement `IEnumerable<T>` but do not have type parameter itself. It is also possible that the type parameter means something completely different. Anyway, we have a `ReflectionUtils.GetEnumerableType` helper method that does exactly what you need here.

PR #424, opened 2017-08-03
Code refactoring - Catch by type.

MSRiganti:
No functional change.

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
@@ -344,16 +344,16 @@ namespace DotVVM.Framework.Hosting
                     await resultTask;
                 }
             }
+            catch (DotvvmInterruptRequestExecutionException ex)
+            {
+                throw new DotvvmInterruptRequestExecutionException("The request execution was interrupted in the command!", ex);
+            }
+            catch (TargetInvocationException ex)
+            {
+                context.CommandException = ex.InnerException;
+            }
             catch (Exception ex)
             {
-                if (ex is TargetInvocationException)
-                {
-                    ex = ex.InnerException;
-                }
-                if (ex is DotvvmInterruptRequestExecutionException)
-                {
-                    throw new DotvvmInterruptRequestExecutionException("The request execution was interrupted in the command!", ex);
-                }
                 context.CommandException = ex;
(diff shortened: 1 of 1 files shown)


exyi:
I'm sorry, but when the exception is of type `TargetInvocationException` with `DotvvmInterruptRequest` in the `InnerException` the current code will correctly interrupt the request, but after your change, it will take it as a command exception, which is incorrect.

MSRiganti:
You're not accurate and you're wrong. 
**Original**:

```
try
{
    result = action.Action();

    resultTask = result as Task;
    if (resultTask != null)
    {
        await resultTask;
    }
}
catch (Exception ex)
{
    if (ex is TargetInvocationException)
    {
        ex = ex.InnerException;
    }
    if (ex is DotvvmInterruptRequestExecutionException)
    {
        throw new DotvvmInterruptRequestExecutionException("The request execution was interrupted in the command!", ex);
    }
    context.CommandException = ex;
}
```
When an exception occurs:
1. Type is **TargetInvocationException**: it goes to **catch (Exception ex)**, 
then it goes in **if (ex is TargetInvocationException)** where **InnerException** replace **ex**,
then **ex** is inserted in **context.CommandException**,
so **InnerException** is inserted in **context.CommandException** just as it is below.
2. Type is **DotvvmInterruptRequestExecutionException**: it goes to  **catch (Exception ex)**, 
then in **if (ex is DotvvmInterruptRequestExecutionException)** where it throw **new DotvvmInterruptRequestExecutionException** and nothing more just as it is below.
3. Type is different than the previous above: It just insert **ex** in **context.CommandException** and nothing else does just as it is below in pull request.



**New in pull request**:
```
try
{
    result = action.Action();

    resultTask = result as Task;
    if (resultTask != null)
    {
        await resultTask;
    }
}
catch (DotvvmInterruptRequestExecutionException ex)
{
    throw new DotvvmInterruptRequestExecutionException("The request execution was interrupted in the command!", ex);
}
catch (TargetInvocationException ex)
{
    context.CommandException = ex.InnerException;
}
catch (Exception ex)
{
    context.CommandException = ex;
}
```

So if the exception is of type **TargetInvocationException** with **DotvvmInterruptRequestExecutionException** in the **InnerException** it will insert **InnerException** into **context.CommandException** like before.

PR #428, opened 2017-08-08
Js postback refactoring

exyi:


src/DotVVM.Framework/Resources/Scripts/DotVVM.PostBackHandlers.ts
```diff
@@ -15,8 +25,21 @@ class ConfirmPostBackHandler extends DotvvmPostBackHandler {
     }
 }
 
-class DotvvmPostBackHandlers {
-    public confirm = (options: any) => new ConfirmPostBackHandler(<string>options.message);
+class PostbackOptions {
+    constructor(public readonly postbackId: number, public readonly sender?: HTMLElement, public readonly args : any[] = [], public readonly viewModel?: any, public readonly viewModelName?: string, public readonly validationTargetPath?: any) {}
+}
+
+class ConfirmPostBackHandler2 implements DotvvmPostbackHandler2 {
```
tomasherceg: Shouldn't it be DotvvmPostBackHandlerEx?

djanosik: I think there is no need to have two separated interfaces just because you need to overload some arguments. BTW: @exyi Why do you use types instead of interfaces here?

djanosik: But if we need another interface, I would call it `AsyncDotvvmPostbackHandler`.

exyi: Unfortunately, the postback handler api is documented and I don't see any way how to make it backward compatible, so I have created the second interface, although I don't like the solution very much. Maybe we could break the compatibility, but it'll result only in a runtime error if you don't use typescript (and upgrade the definitions...).

And I use `type` instead of `interface` because as far as I know, the type is more powerful - I don't think you can do the discriminated union in an interface. But I'm not a Typescript expert, do you think it would better to use interface?

djanosik: One important difference is that interfaces can be extended, while type aliases can't. You should always use interface when it's possible. BTW: Can we change type names while maintaining runtime compatibility?

exyi: Ok, but extending this interface will break things, so I don't see any benefit.

exyi: We can't change the type name, because it's a class ;(

djanosik: Why should it break things? Of course, this example is pointless, but why to disallow that?
```
interface MySuperPostbackHandler extends DotvvmPostbackHandler2 {
      name: string;
}
```

exyi: I thought that you wanted to extend the interface using `interface DotvvmPostbackHandler2 { ... more props... }`, that would "work" with interfaces and don't with type. Extending using `extends` works fine, your example actually works. And you could also do `type MyPostbackHandler = DotvvmPostbackHander & { name: string }` ;)

djanosik: Ok, you have convinced me. But it still feels unusual.

PR #433, opened 2017-08-11
Feature: FormControls.Enabled

Sejsel:
# FormControls.Enabled
_proposed in #296_

Adds a FormControls.Enabled property to all HTML elements which disables controls inside that are not explicitly enabled.

### DotVVM

- Button (ButtonBase)
- CheckBox (CheckableControlBase)
- ComboBox (SelectorBase)
- DataPager
- LinkButton (ButtonBase)
- ListBox (SelectorBase)
- RadioButton (CheckableControlBase)
- TextBox

### Bootstrap (not yet implemented)

- Button
- CheckBox
- DataPager
- RadioButton
- TextBoxGroup

### Business pack (not yet implemented)

- Button
- Calendar
- CheckBox
- CheckBoxList
- ColorPicker
- ComboBox
- DataPager
- DateTimePicker
- DateTimeRangerPicker
- DropDownList
- MultiSelect
- NumericUpDown
- RadioButton
- RadioButtonList
- RangeCalendar
- RangeSlider
- Rating
- Slider
- TextBox

ListView (Business pack) does not have an Enabled property
DateTimePicker (Bootstrap) does not have an Enabled property
DropDownButton (Bootstrap) does not have an Enabled property

Example
---
```html
<div FormControls.Enabled="{value: false}">
    <dot:Button Text="I am disabled" />
    <dot:Button Text="I am enabled" Enabled="{value: true}" />
</div>
```

### DIFF
--- a/src/DotVVM.Framework/Binding/CustomDefaultProperty.cs
+++ b/src/DotVVM.Framework/Binding/CustomDefaultProperty.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DotVVM.Framework.Controls;
+
+namespace DotVVM.Framework.Binding
+{
+    public class CustomDefaultProperty : DotvvmProperty
+    {
+        /// <summary>
+        /// Gets the property whose value this property will default to if it is not set
+        /// </summary>
+        public DotvvmProperty DefaultProperty { get; protected set; }
+
+        /// <summary>
+        /// Gets whether the value of the default property when can be inherited from the parent controls.
+        /// </summary>
+        public bool DefaultPropertyInherit { get; protected set; }
+
+        /// <summary>
+        /// Gets the value of the property.
+        /// </summary>
+        public override object GetValue(DotvvmBindableObject control, bool inherit = true)
+        {
+            object value;
+            if (control.properties != null)
+            {
+                if (control.properties.TryGetValue(this, out value))
+                {
+                    return value;
+                }
+            }
+            if (IsValueInherited && inherit && control.Parent != null)
+            {
+                return GetValue(control.Parent);
+            }
+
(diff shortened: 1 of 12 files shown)


src/DotVVM.Framework/Binding/CustomDefaultProperty.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DotVVM.Framework.Controls;
+
+namespace DotVVM.Framework.Binding
+{
+    public class CustomDefaultProperty : DotvvmProperty
```
djanosik: When I see this name I can't imagine what is it good for. Is `DotvvmPropertyWithFallback` better?

Sejsel: To be honest, I don't find either of those names very good and can't think of a better alternative.

djanosik: Hmm. Will try to find better name.

Sejsel: How about `DefaultToPropertyDotvvmProperty` or `PropertyDefaultDotvvmProperty`?

djanosik: Or `DelegatedDotvvmProperty`? Because it delegates call to another property when not set.

Sejsel: I asked @exyi and @tomasherceg and it's going to be `DotvvmPropertyWithFallback`.

djanosik:
All controls in Business Pack will support `FormControls.EnabledProperty` in v2.0.

PR #435, opened 2017-08-14
Feature diagnostics reporting

Mrnda:
Add reporting of diagnostics infomation (viewmodels, HTTP headers, exception stacktrace,...).


src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        private DiagnosticsServerConfiguration configuration;
+
+        public string DiagnosticsServerHostname
+        {
+            get
+            {
+                if (configuration == null)
+                    LoadConfiguration();
```
Mrnda: Forgotten that from earlier. Will remove that as it does not make sense.

src/DotVVM.Framework/Diagnostics/Models/HttpHeaderItem.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public string Key { get; set; }
+        public string Value { get; set; }
+
+        public static HttpHeaderItem FromKeyValuePair(KeyValuePair<string, string[]> pair)
+        {
+            return new HttpHeaderItem
+            {
+                Key = pair.Key,
+                Value = pair.Value[0]
```
Mrnda: OK will add

src/DotVVM.Framework/Diagnostics/DiagnosticsRequestTracer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            this.informationSender = sender;
+        }
+
+        public Task TraceEvent(string eventName, IDotvvmRequestContext context)
+        {
+            if (eventName == RequestTracingConstants.BeginRequest)
+            {
+                stopwatch.Start();
+            }
```
Mrnda: OK I will add that to the DiagnosticsInformation and to the extension.

Mrnda: OK. Everything should be fixed by now in the latest commit.

Issue #405, opened 2017-07-24
RedirectToRoute should support query string parameters


tomasherceg:
We should add an option to specify query string parameters in the `RedirectToRoute` methods.

exyi:
resolved in #439

Issue #404, opened 2017-07-24
Query-* property group in RouteLink


tomasherceg:
The `RouteLink` control should have the `Query-*` property group which would be able to add or modify query string parameters. 

The behavior should be the same as with route parameters, so if the parameter is already present in the query string of the current page, it is copied to the new page (unless it is set to empty string explicitly).

exyi:
resolved in #439

PR #439, opened 2017-08-18
Query string parameters to RedirectToRoute, Query-* property group in RouteLink

Rumbeer:
Added option to specify query string parameters in `RedirectToRoute` method.

Added UI test for this feature.

Issue https://github.com/riganti/dotvvm/issues/405

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
@@ -7,6 +7,8 @@ using System.Net;
 using DotVVM.Framework.Controls;
 using DotVVM.Framework.Storage;
 using System.Diagnostics;
+using System.Reflection;
+using System.Text.RegularExpressions;
 using DotVVM.Framework.Hosting.Middlewares;
 using DotVVM.Framework.ViewModel.Serialization;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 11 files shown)


src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -83,14 +85,37 @@ public static void RedirectToUrl(this IDotvvmRequestContext context, string url,
         /// <summary>
         /// Returns the redirect response and interrupts the execution of current request.
         /// </summary>
-        public static void RedirectToRoute(this IDotvvmRequestContext context, string routeName, object newRouteValues = null, bool replaceInHistory = false, bool allowSpaRedirect = true, string urlSuffix = null)
+        public static void RedirectToRoute(this IDotvvmRequestContext context, string routeName, object newRouteValues = null, bool replaceInHistory = false, bool allowSpaRedirect = true, string urlSuffix = null, object queryStringParameters = null)
```
exyi: I'd rather use a shorter name because this parameter is going to be used as a named one, what about `query`? I think it's quite clear what it means the context of redirection.

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -83,14 +85,37 @@ public static void RedirectToUrl(this IDotvvmRequestContext context, string url,
...
+        private static string BuildUrlSuffix(string urlSuffix, object queryStringParameters)
+        {
+            var hashIndex = urlSuffix.IndexOf('#');
+            var resultSuffix = hashIndex < 0 ? urlSuffix : urlSuffix.Substring(0, hashIndex);
+            if (queryStringParameters is IEnumerable<KeyValuePair<string, string>>)
+            {
+                foreach (KeyValuePair<string, string> item in (IEnumerable<KeyValuePair<string, string>>)queryStringParameters)
+                {
+                    resultSuffix = resultSuffix + (urlSuffix.LastIndexOf('?') < 0 ? "?" : "&") + Regex.Escape(item.Key) +
+                                "=" + Regex.Escape(item.Value);
```
exyi: `Regex.Escape` IMO does not make any sense for escaping urls, did you wanted to use `Uri.EscapeDataString` instead?

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -83,14 +85,37 @@ public static void RedirectToUrl(this IDotvvmRequestContext context, string url,
...
+            var url = route.BuildUrl(context.Parameters, newRouteValues) + BuildUrlSuffix(urlSuffix ?? "", queryStringParameters);
 
             context.RedirectToUrl(url, replaceInHistory, allowSpaRedirect);
         }
 
+        private static string BuildUrlSuffix(string urlSuffix, object queryStringParameters)
+        {
+            var hashIndex = urlSuffix.IndexOf('#');
+            var resultSuffix = hashIndex < 0 ? urlSuffix : urlSuffix.Substring(0, hashIndex);
+            if (queryStringParameters is IEnumerable<KeyValuePair<string, string>>)
```
exyi: I hope you could use C# 7 pattern matching feature, it would make the code a bit more repetitive in this case ;) https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching

Rumbeer:
Added `Query-*` property group to `RouteLink` that use the same method for building the suffix when the control is rendered on server.

Added UI test for this feature.

Issue https://github.com/riganti/dotvvm/issues/404

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -755,6 +755,28 @@ class DotVVM {
...
+            hashSuffix = urlSuffix.substring(urlSuffix.indexOf("#"));
+        } else {
+            resultSuffix = urlSuffix;
+            hashSuffix = "";
+        }
+        for (var property in query) {
+            if (query.hasOwnProperty(property)) {
+                if (!property) continue;
+                var queryParamValue = ko.unwrap(query[property]);
+                if (!queryParamValue) continue;
```
exyi: This will break when the query would contain number `0` or boolean `false`, you should probably use `queryParamValue != null` condition. It will `continue;` for `null` and `undefined` and will include the parameter for `0`, empty string and `false`

Issue #368, opened 2017-06-19
GridViewTextColumn needs a TextChanged command.


siwicgrey:

exyi:
In general, there is quite a lot of controls that only wrap another control, so it would be cool to support some easy way to access the inner control properties to eliminate the need to copy all of them and then handle issues (like this one) that we forgot about one.

Maybe it could be used like properties: `<dot:GridViewTextColumn TextBox.TextChanged=... />` and then copied using a simple `foreach` loop in the column.

Or the control could be accessible as a inner element property: `<dot:GridViewTextColumn><TextBox TextChanged=... /> </doty:GridViewTextColumn />`. The GridViewTextColumn.cs would look like:

```C#
[MarkupOptions(InnerElement)]
public TextBox TextBox { get; set; }

override OnInit() {
    TextBox.SetValue(TextBox.TextProperty, this.GetValueRaw(TextProperty));
    ... few other most common properties
    Children.Add(TextBox)
}

cc @djanosik

tomasherceg:
For now, let's just add the property to the column and set it to the TextBox when the column is in the edit mode.

PR #443, opened 2017-08-25
Added Changed property to GridViewTextColumn

Rumbeer:
Issue https://github.com/riganti/dotvvm/issues/368

I chose the name `Changed` for the property to be consistent with other properties with similar functionality.


djanosik:
I think the name should be `ChangedBinding` and its type should be `ICommandBinding`. But I am not sure about that.  @exyi

exyi:
@djanosik I agree with that because it's more consistent but I think that it does not matter much in this case. In this case, there is actually no reason why would the value be disallowed except [this error](https://github.com/riganti/dotvvm/pull/443/files#diff-ab0255b92599e438c1cfeb716182f1b6R96) (although I haven't seen value usage of `Command` property...)

djanosik:
It's more about convention. What properties on "wrapping" controls like GridViewColumn should have the `Binding` suffix? All props that are copied to wrapped control?

Rumbeer:
Ok, so I will change the property to `ChangedBinding`.

PR #447, opened 2017-09-07
Binding compiler: Added support for semicolon

exyi:
* It basically works like a binary expression which returns the right expression after execution of the left one
* expressions without right node like `expr;` return void


djanosik:
@exyi @tomasherceg I suppose this change is related to REST API bindings, right? Should we merge it? What is the status of RES API bindings?

exyi:
On Thu, 2017-12-07 at 17:30 +0000, Dusan Janosik wrote:
> @exyi @tomasherceg I suppose this change is related to REST API
> bindings, right?
Yes it is, it is the main reason we need this, but it may be used in
other bindings too.

> Should we merge it?
It is not accepted... It it related to compiler - @Mylan719 would you
like to have a look at it?

> What is the status of RES API bindings?

The stuff that is implemented is implemented... It's untested, and I
don't khow how the tests should look like.

tomasherceg:
@Mylan719 Did you have time to look at it? 
I would like to merge this so we can finalize the REST API bindings. Without the semicolon operator, the feature is not very useful.

exyi:
I'm merging it, CI seems to be happy and you had enough time for protests...

PR #448, opened 2017-09-08
remove jQuery from Debug mode

symbiogenesis:
This is an attempt at removing jQuery. Don't really know if it works.


EmbraceUnity:
This is more to get the discussion going. How could I go about testing this?

exyi:
Hi, thank you for the PR, it looks basically correct, I have just fixed few minor mistakes. The only functional change I see is the change of `hide` and `show` methods to `setTimeout` which does not do the animation. Maybe we could do it using CSS, or simply drop it, as it's not very necessary feature ;)

You can test it by running dotvvm samples (project `DotVVM.Samples.BasicSamples.AspNetCore`) and creating an error in a command or launching a command that is client-side interrupted (due to validation, for example). Check out `FeatureSamples/Validation/SimpleValidation` and `/FeatureSamples/ActionFilterErrorHandling/ActionFilterErrorHandling` samles (Try Unhanded Error button add new task that is empty)

PR #450, opened 2017-09-09
Implemented TextBox.UpdateTextOnInput

### LINKED ISSUE #430: TextBox UpdateAfterKeyDown change doesn't fire when the user clicks on the clear icon in the textbox
In some browsers, the `input` controls have the clear icon (cross) on the right side, but when you click on it, the change event is not fired even when `UpdateAfterKeyDown` is set to `true`.

exyi:
The `TextBox.UpdateTextAfterKeyDown` is now obsolete, but behaves the same as `UpdateTextOnInput` - although it is "incorrect", IMO it's better than the previous way it worked

The `UpdateTextOnInput` property can be inherited, so you may turn it on for an entire form

I have unified the `FormatString` property and calling `ToString` method in the binding because the `FormatString` property did not support update on keydown/input

resolves #430


djanosik:
Moving to Version 2.1. @quigamdev We need to decide how to handle UI validation when we remove the `dotvvm-textbox-text` binding handler. https://github.com/riganti/dotvvm/blob/1919736a47f4928956db6b7b5e409d0e5094ce85/src/DotVVM.Framework/Resources/Scripts/DotVVM.ts#L1180-L1188

tomasherceg:
I have merged changes from master and added the `TextInput` event to the control (with disclaimer mentioning its effects to the user experience).
The only failed test on ASP.NET Core concerns GitHub API and it's an issue with the test, so I am merging this.

PR #457, opened 2017-09-19
Allow manual integrity hash and add tests

symbiogenesis:
Seems that auto-computing the integrity hash, as is done now, wouldn't necessarily be the right thing, given that it would only ensure that the resource hasn't changed while the instance of Dotvvm is running.

If you restart your instance, and the resource has become malicious, then it will just hash the malicious resource and write that as the integrity SRI.

Most good CDNs provide the SRI, and it should never change. So if you have it, you should be able to use it.

This patch, as is, doesn't remove the auto-computing, it only adds the ability to add a manual hash, just to be conservative for now, if there was another use case I hadn't considered (development?).

But you may wish to remove the auto-compute entirely. Then you could remove the boolean VerifyResourceIntegrity property, as well.


exyi:
This would not happen, as the integrity hash is only computed from the LocalFileResourceLocation not UrlResourceLocation, so when you have UrlResourceLocation as primary and local or embedded resource location as a fallback it will ensure that the CDN will never send a different resource than what you have downloaded and added to project.

The implementation looks OK, I'm just not sure if we need this feature but I think that is may be useful when you don't want to include the entire resource in the project

PR #464, opened 2017-10-02
Add Async and Defer boolean attributes for script resources

symbiogenesis:


src/DotVVM.Framework/Controls/IHtmlWriter.cs
```diff
@@ -18,6 +18,12 @@ public interface IHtmlWriter
         /// <param name="appendSeparator">The separator that will be used when <paramref name="append"/> is true and when the attribute already has a value.</param>
         void AddAttribute(string name, string value, bool append = false, string appendSeparator = null);
 
+		/// <summary>
+        /// Adds the specified boolean attribute to the next HTML element that is being rendered. Boolean attributes such as 'async' and 'defer' have no value. They are true when present and false otherwise
+        /// </summary>
+        /// <param name="name">The name of the HTML attribute.</param>
+		void AddBooleanAttribute(string name);
```
djanosik: I think this method should not be defined on this interface. You can create an extension method in your project.

exyi:
Actually, I don't think that the `Defer` and `Async` api should exist in this form, as setting these properties on one script resource would break all dependencies. It also pretty useless to load the `script`s asynchronously when they are already on the end of the page, and styles are not synchronous by default.

symbiogenesis:
I wanted get a good score on Google PageSpeed, even if there is no other benefit. It is an SEO thing.

Currently, PageSpeed says:

> None of the above-the-fold content on your page could be rendered without waiting for the following resources to load. Try to defer or asynchronously load blocking resources, or inline the critical portions of those resources directly in the HTML.
> 
> Remove render-blocking JavaScript:
> 
>     https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js

I also wanted to load scripts via CDN with integrity hashes, and had put the script tag directly in the dotmaster.

However, now I am able to specify the Integrity hash, I could probably load all of my scripts via the registration, and wire up the dependencies correctly.

So, perhaps this isn't necessary as long as one is consistent in using the script registration.

exyi:
On Thu, 2017-12-28 at 20:14 +0000, symbiogenesis wrote:
> I wanted get a good score on Google PageSpeed, even if there is no
> other benefit. It is an SEO thing.

Ok, I think that we might render `defer` attribute by default for all
scripts. I'm pretty sure it would not break anything, could help with
the (stupid) statistics and it will certainly not harm the performance
(maybe it would get measurably better...). If you'd like that, we can
create a separate issue for that, it should be a easy fix.

> I also wanted to load scripts via CDN with integrity hashes, and had
> put the script tag directly in the dotmaster.

Do you have any info about integrity and performance? I actually think
that browsers *could* use that for caching (as content adressed blobs),
but I have not seen anything about that. It would mean that it would be
profitable to render these integrity checks even for local resources by
default.

symbiogenesis:
Given that the entire point of DotVVM is to minimize the need to use javascript, it makes sense that we should have predictable control over how it is used and in that case we may be in a position to always defer javascript. Async probably not, although I'm not clear on what it truly means. I'll open a separate issue on that.

> Do you have any info about integrity and performance? I actually think that browsers *could* use that for caching (as content adressed blobs), but I have not seen anything about that. 

That is a brilliant idea, although I don't know how local caching works in browsers. May vary by browser. That would be a great question to ask Mozilla, Google, etc.

symbiogenesis:
The response from WHATWG is 

> As for identifier-based caching, the problem is that this exposes, across origins, which resources you have visited. w3c/webappsec-subresource-integrity#22 has more details and pointers on this.

[https://github.com/whatwg/fetch/issues/301](url)

exyi:
> Given that the entire point of DotVVM is to minimize the need to use javascript
I'd be careful with this assumption :P Although dotvvm is definitely designed to help you with that.

`async` is problematic, as it does not guarantee the order of execution, and I don't think it will have significantly better performance than `defer`.

And thanks for the info about the subresource integrity, I have also found an interesting study of the problem (https://hillbrad.github.io/sri-addressable-caching/sri-addressable-caching.html). It just seems that nothing is easy on web :)

I'm closing this PR in favour of #528, thank you.

Issue #432, opened 2017-08-11
IHttpRequest is missing some properties which are present in Url


tomasherceg:
The `Host` property is missing from `IHttpRequest`, but the other properites like `QueryString` are there.
We should either remove all of them and keep `Uri` only (which is a breaking change), or add the missing ones.

Rumbeer:
Is there a reason why `Path`, `PathBase`, `Query`... properties in Owin implementation are set in the constructor and not from `OriginalRequest` like .Net core implementation or other properties in this class?

exyi:
Done in #465

PR #465, opened 2017-10-03
Removing some IHttpRequest properties

Rumbeer:
Made `IHttpRequest` properties `Path`, `Sheme` and `QueryString` obsolete.

Replaced `Path` with `Url.AbsolutePath`, `Scheme` with `Url.Scheme` and `QueryString` with `Url.Query`. I left the property `PathBase` because i did not know what to replace it with.

issue https://github.com/riganti/dotvvm/issues/432


adamjez:
I reverted part of this PR in https://github.com/riganti/dotvvm/commit/f7e81eaca50137213a42425972bb3cf05f5d2b2d.
There isn't good replace for property `Path`. 

You suggest `Url.AbsolutePath` **but** it includes **virtual path** which `Path` doesn't include. Without property `Path` there isn't a way to retrieve Url path without virtual path which leads to routing completely not working.

Issue #156, opened 2016-03-10
ComboBox - add ValueMatchBinding property


tomasherceg:
The viewmodel looks like this:

```
public List<Country> Countries { get; set; }
public Country SelectedCountry { get; set; }
```

The view looks like this:

```
<dot:ComboBox DataSource="{value: Countries}" SelectedValue="{value: SelectedCountry}" />
```

However, the binding doesn't work really well because when deserializing the JSON from the server, the deserializer doesn't know that the SelectedCountry and e.g. Countries[2] are the same instances.
Therefore we need a property to detect, which instances are equal. 

The usage would be like this:

```
<dot:ComboBox DataSource="{value: Countries}" SelectedValue="{value: SelectedCountry}"
    ValueMatchBinding="{value: Id}" />
```

martinsura:
Btw is planned use ValueMatchBinding="{value:Id}" for mulit-select ComboBox?

tomasherceg:
We were also thinking about the following names:

* `KeyBinding`
* `ItemKeyBinding`
* `ItemCompareBinding`
* `EqualityBinding`

tomasherceg:
To be consistent with the Business Pack naming, we will use `ItemKeyBinding`.

exyi:
There are actually more problems than just matching values by id when you have the same instance multiple times in the same view model. For example, the client-side deserialization changes both the selected object and the object in the collection even though you have changed only one on the server. And it does much crazier things if you change both of them... And it gets even worse when you have the collection bound to the ComboBox, although I'm not sure exactly why.

I think that we basically can't fix this issue, without changing the serialization/deserialization process significantly (to something like [lave](https://github.com/jed/lave)), and the only thing we can do is to warn users in the docs. Maybe we could throw an exception or write a warning to console when one object is used multiple times in the view, but I don't know how to efficiently detect it in JS (no reference-based hashtable)

djanosik:
Well, we must make sure we don't use the same instance on multiple places in view model. That's what we do in BP and the `ItemKeyBinding` works quite well. Of course, you shouldn't have multiple objects with the same key.

exyi:
@djanosik But the knockout puts the original object from the collection to selected value by reference, which causes that the object is multiple times in the view model. Do you deep copy all the things in BussinessPack?

djanosik:
Exactly, we make deep copies of objects added to selected values. Because its very similar to what happens when you set SelectedValue in postback. It's deserialized as another instance.

[...]

PR #470, opened 2017-10-04
Removed obsoletes

### LINKED ISSUE #466: Remove obsolete properties and methods
There are some obsolete properties and methods in framework (eg. in GridViewDataSet) created in 1.x releases. We should remove them in 2.0.

exyi:
This was pushed to master by @Leeemons, but it's hard for me to do a review with some expected feedback for me, so I'm reopening it as a Pull Request. Sorry for complications.

Resolves #466

> There are some obsolete properties and methods in framework (eg. in GridViewDataSet) created in 1.x releases. We should remove them in 2.0.


src/DotVVM.Framework/Binding/BindingHelper.cs
```diff
@@ -19,9 +19,6 @@ public static partial class BindingHelper
     {
         public static T GetProperty<T>(this IBinding binding, ErrorHandlingMode errorMode = ErrorHandlingMode.ThrowException) => (T)binding.GetProperty(typeof(T), errorMode);
 
-        [Obsolete]
-        public static string GetKnockoutBindingExpression(this IValueBinding binding) =>
-            JavascriptTranslator.FormatKnockoutScript(binding.KnockoutExpression);
```
exyi: This method was made obsolete in dotvvm 2.0, and I think that we should not remove, since it's used at a lot of places. Could you please add it back?

src/DotVVM.Framework/Controls/GridViewTextColumn.cs
```diff
@@ -40,18 +40,6 @@ public ICommandBinding ChangedBinding
...
-        /// Gets or sets the type of value being formatted - Number or DateTime.
-        /// </summary>
-        [MarkupOptions(AllowBinding = false)]
-        public FormatValueType ValueType
-        {
-            get { return (FormatValueType)GetValue(ValueTypeProperty); }
-            set { SetValue(ValueTypeProperty, value); }
-        }
-        public static readonly DotvvmProperty ValueTypeProperty =
-            DotvvmProperty.Register<FormatValueType, GridViewTextColumn>(t => t.ValueType);
```
exyi: This property was also removed in version 2.0 and it's also used in the dothtml markup, which does not give you any way to remove all obsolete usages and will not report an error when the property is misused. I think that we should also add this one back.

src/DotVVM.Framework/Controls/KnockoutHelper.cs
```diff
@@ -34,12 +34,6 @@ public static class KnockoutHelper
             }
         }
 
-        [Obsolete("Use the AddKnockoutDataBind(this IHtmlWriter writer, string name, IValueBinding valueBinding, DotvvmControl control) or AddKnockoutDataBind(this IHtmlWriter writer, string name, string expression) overload")]
-        public static void AddKnockoutDataBind(this IHtmlWriter writer, string name, IValueBinding valueBinding)
-        {
-            writer.AddKnockoutDataBind(name, valueBinding.GetKnockoutBindingExpression());
-        }
```
exyi: This is also v2.0 obsoleted method.

src/DotVVM.Framework/Controls/Literal.cs
```diff
@@ -42,19 +42,6 @@ public string FormatString
...
-        /// </summary>
-        [MarkupOptions(AllowBinding = false)]
-        [Obsolete("ValueType property is no longer required, it is automatically inferred from compile-time type of Text binding")]
-        public FormatValueType ValueType
-        {
-            get { return (FormatValueType)GetValue(ValueTypeProperty); }
-            set { SetValue(ValueTypeProperty, value); }
-        }
-        public static readonly DotvvmProperty ValueTypeProperty =
-            DotvvmProperty.Register<FormatValueType, Literal>(t => t.ValueType);
```
exyi: Same as TextBox.ValueType

src/DotVVM.Framework/Controls/SelectorBase.cs
```diff
@@ -26,32 +26,7 @@ public bool Enabled
...
-        /// Gets or sets the name of property in the DataSource collection that will be passed to the SelectedValue property when the item is selected.
-        /// </summary>
-        [Obsolete("Use ItemValueBinding property instead")]
-        public string ValueMember
-        {
-            get { return (string)GetValue(ValueMemberProperty); }
-            set { SetValue(ValueMemberProperty, value); }
-        }
-        public static readonly DotvvmProperty ValueMemberProperty =
-            DotvvmProperty.Register<string, SelectorBase>(t => t.ValueMember, "");
```
exyi: Also v2.0 ones. And these are also used in the markup, so we could not remove these anyway.

src/DotVVM.Framework/Controls/TextBox.cs
```diff
@@ -105,24 +105,9 @@ public bool UpdateTextAfterKeydown
...
-        /// <summary>
-        /// Gets or sets the type of value being formatted - Number or DateTime.
-        /// </summary>
-        [MarkupOptions(AllowBinding = false)]
-        [Obsolete("ValueType property is no longer required, it is automatically inferred from compile-time type of Text binding")]
-        public FormatValueType ValueType
-        {
-            get { return (FormatValueType)GetValue(ValueTypeProperty); }
-            set { SetValue(ValueTypeProperty, value); }
-        }
```
exyi: Same as other `ValueType`s

src/DotVVM.Framework/Hosting/DotvvmRequestContext.cs
```diff
@@ -104,9 +104,6 @@ public class DotvvmRequestContext : IDotvvmRequestContext
         /// </summary>
         public bool IsInPartialRenderingMode => DotvvmPresenter.DeterminePartialRendering(HttpContext);
 
-        [Obsolete("Get the IViewModelSerializer from IServiceProvider")]
-        public IViewModelSerializer ViewModelSerializer => Services.GetRequiredService<IViewModelSerializer>();
```
exyi: I'm not sure, but this is IMO also a v2.0 obsolete.

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -48,24 +48,6 @@ public static void ChangeCurrentCulture(this IDotvvmRequestContext context, stri
         }
 
         /// <summary>
-        /// Gets the current UI culture of this HTTP request.
-        /// </summary>
-        [Obsolete("This just returns CultureInfo.CurrentUICulture")]
-        public static CultureInfo GetCurrentUICulture(this IDotvvmRequestContext context)
```
exyi: Certainly a v2.0 obsolete https://github.com/riganti/dotvvm/commit/078092050ed620b224efb5f1abb7362779047960#diff-e0edace51ac65a3259596be0d7a09773

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -48,24 +48,6 @@ public static void ChangeCurrentCulture(this IDotvvmRequestContext context, stri
...
-        {
-            return CultureInfo.CurrentUICulture;
-        }
-
-        /// <summary>
-        /// Gets the current culture of this HTTP request.
-        /// </summary>
-        [Obsolete("This just returns CultureInfo.CurrentCulture")]
-        public static CultureInfo GetCurrentCulture(this IDotvvmRequestContext context)
```
exyi: Same as above.

src/DotVVM.Framework/ResourceManagement/InlineScriptResource.cs
```diff
@@ -12,11 +12,6 @@ namespace DotVVM.Framework.ResourceManagement
     /// </summary>
     public class InlineScriptResource : ResourceBase
     {
-        [Obsolete("Code parameter is required, please provide it in the constructor.")]
-        public InlineScriptResource(ResourceRenderPosition renderPosition = ResourceRenderPosition.Body) : base(renderPosition)
```
exyi: I have seen this one a lot of times, I'd also add it back.

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -29,33 +29,6 @@ private static IEnumerable<string> GetRoutesForFile(string fileName)
             else yield return pureName;
         }
 
-        [Obsolete("Use IRoutingStrategy.")] // TODO: make this a helper for IRoutingStrategy
```
exyi: Is the TODO comment fulfilled?

src/DotVVM.Framework/Compilation/Javascript/JavascriptCompilationHelper.cs
```diff
@@ -10,23 +10,6 @@ public static class JavascriptCompilationHelper
...
-        public static ParametrizedCode AddIndexerToViewModel(ParametrizedCode script, object index, bool unwrap = false) =>
-            AddIndexerToViewModel(script, new JsLiteral(index), unwrap);
-        [Obsolete]
-        public static ParametrizedCode AddIndexerToViewModel(ParametrizedCode script, JsExpression indexer, bool unwrap = false)
-        {
-            return indexerCode.AssignParameters(o =>
-                o == indexerTargetParameter ? new CodeParameterAssignment(script) :
-                o == indexerExpressionParameter ? CodeParameterAssignment.FromExpression(indexer) :
-                default(CodeParameterAssignment));
-        }
```
exyi: This one is a v2.0 obsolete, but I'd keep it removed, as it's not a very commonly used API.

djanosik:
@exyi Thanks for review. I will also check all the methods being removed. BTW: We should probably mark all issues introducing breaking changes with `breaking` tag.

src/DotVVM.Framework/Binding/BindingHelper.cs
djanosik: Yes, we should definitely keep this one. Why is it obsolete? Is there any alternative?

src/DotVVM.Framework/Controls/GridViewTextColumn.cs
djanosik: I am not sure here. Will it break anything if we remove it? It will, in this case, because bindable objects do not support HTML attributes. But in other cases it will just render an unwanted HTML attribute. @quigamdev What do you think?

src/DotVVM.Framework/Controls/SelectorBase.cs
djanosik: Ok, we will remove these in 3.0. BTW: Shouldn't we also add `ItemKeyBinding` to fix #156?

src/DotVVM.Framework/Hosting/DotvvmRequestContext.cs
djanosik: Yes, it's v2.0 obsolete. We should keep it.

src/DotVVM.Framework/ResourceManagement/InlineScriptResource.cs
djanosik: It's a v2.0 obsolete anyway. We should keep it.

src/DotVVM.Framework/Storage/IReturnedFileStorage.cs
```diff
@@ -9,12 +9,6 @@ public interface IReturnedFileStorage
         /// <summary>
         /// Stores the file and returns its unique ID.
         /// </summary>
-        [Obsolete("Don't bother with implementing this one, it is not used anymore")]
-        Task<Guid> StoreFile(byte[] bytes, ReturnedFileMetadata metadata);
```
djanosik: it's a v2.0 obsolete, but I would remove it.

src/DotVVM.Framework/Routing/RouteHelper.cs
djanosik: Its behavior is similar to `DefaultRouteStrategy ` isn't it?

exyi: Yes it is, but there should be an extension method that will apply the DefaultRouteStrategy

src/DotVVM.Framework/Controls/SelectorBase.cs
exyi: I don't like the `ItemKeyBinding` as it requires reimplementing knockout's `options` and `value` binding, doing a deep clone of the item (which may break validation errors IMO) and will match different items based on a ID, which may IMO cause even more problems. I'd at least wait for VDOM update with this one, as it removes at least some of the disadvatages.

src/DotVVM.Framework/Binding/BindingHelper.cs
exyi: Yes, it's just missing a proper description... You should use `binding.GetKnockoutBindingExpression(control)`.

src/DotVVM.Framework/Routing/RouteHelper.cs
djanosik: So... we need to update the extension method to use `DefaultRouteStrategy`. And the route strategy must accept the `pattern` parameter.

exyi:
I have resolved that in 706c13373a587ebba4e95d93d9931ad2bd11063b

PR #486, opened 2017-10-29
Added LocalizablePresenter

### LINKED ISSUE #330: Localization on .NET 4.6 and .NET Core
The localization solution as suggested in our [docs](https://www.dotvvm.com/docs/tutorials/basics-globalization/1-1) (calling `ChangeCurrentCulture` in VM or action filter) works on .NET 4.5 but won't work on .NET 4.6 or Core. It's because there was a breaking change in how current cultures are stored. MS can probably explain that better than me :)

> We had a change in setting the current cultures in 4.6. The change is setting the current cultures travel with the async operations. So if you set the culture during the async operation it will affect the current execution context. And when the operation is done and switch back to the original execution context, the culture will get reset to the culture in that context.

It means calling `ChangeCurrentCulture` will not change culture for the current thread or request anymore. Instead it will change culture for the current async method. There is no workaround, everything depends on where you change the culture. It's usually too late in VM or action filter. 

We can enable the old behavior for the whole application, but I don't think it's a good idea.

```CSHARP
AppContext.SetSwitch("Switch.System.Globalization.NoAsyncCurrentCulture", true);
```

Users on .NET Core can use the [RequestLocalization](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization#localization-middleware) middleware. It's already supported and it will override our `DotvvmConfiguration.DefaultCulture` setting. But there is no solution for .NET 4.6 and OWIN.

exyi:
It's a utility for localizing specific route based on route parameter or query string

Resolves #330. For localizing an entire application you can use standard [RequestLocalization](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization#localization-middleware) middleware, it's probably not going to be based on route parameters.

> But there is no solution for .NET 4.6 and OWIN.
It's called Asp.Net Core, you should migrate and this is one of the reasons.

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
@@ -30,12 +30,14 @@ namespace DotVVM.Framework.Hosting
         /// <summary>
         /// Changes the current culture of this HTTP request.
         /// </summary>
+        [Obsolete("This method only assigns CultureInfo.CurrentCulture, which is not preserved in async methods. You should assign it manually, or use RequestLocalization middleware or LocalizablePresenter.")]
         public static void ChangeCurrentCulture(this IDotvvmRequestContext context, string cultureName)
             => context.ChangeCurrentCulture(cultureName, cultureName);
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


exyi:
@djanosik  You are right.

PR #494, opened 2017-11-11
Better stack traces in the error page using Ben.Demystify

exyi:
Exception stack traces for the Error Page now come from `EnhancedStackTrace` from the Ben.Demystify library. Because it's a bit experimental, it falls back to the normal stack traces transparently, so it should not break anything.

Some screenshots (with highlighted cool stuff)

screenshot-2017-11-11 server error in application
screenshot-2017-11-11 server error in application 1

Note that it only works on `netstandard2.0` target, not `net451`, but it should not be a big deal as it's just a cosmetic change and you should be used to rough development experience if you are stuck to obsolete stuff :P


quigamdev:
Waits untill Ben.Demystifier is not strongly-named.

exyi:
What is the problem with strong naming?

quigamdev:
I updated Ben.Demystifier to newest version and I allowed usage in owin projects. However owin requires strongly-named assembly.

exyi:
> I updated Ben.Demystifier to newest version and I allowed usage in
> owin projects.
Ok, that's cool.

>  However owin requires strongly-named assembly.
I don't want to know why... Anyway, I'd merge it for Asp.Net Core and
maybe add the support for OWIN later rather than waiting for the
strong-named package.

PR #499, opened 2017-11-24
Added support for Dependency Injection into nested ViewModels

exyi:
It has to be explicitly enabled using Mapper Api:

```C#
mapper.Map(m => {
	// create from Asp.Net DI
	m.AllowDependencyInjection();
	// custom constructor
	m.SetConstructor(serviceProvider => SomehowCreateViewModel(serviceProvider))l
});
```

cc @Mylan719


src/DotVVM.Framework/ViewModel/Serialization/ViewModelMapperHelper.cs
```diff
@@ -22,6 +22,16 @@ public static IViewModelSerializationMapper Map(this IViewModelSerializationMapp
             return mapper;
         }
 
+        public static void SetConstructor(this ViewModelSerializationMap map, ObjectFactory factory)
+        {
+            map.SetConstructor(p => factory.Invoke(p, Array.Empty<object>()));
+        }
+
+        public static void AllowDependencyInjeciton(this ViewModelSerializationMap map)
```
djanosik: TYPO: AllowDependencyInjection :)

Issue #505, opened 2017-11-28
Remove obsolete properties from GridViewDataSet


djanosik:
Some properties in `GridViewDataSet` were marked as obsolete in v1.x. I think we can remove them in v2.0.

Issue #507, opened 2017-11-28
Add generic IGridViewDataSet interface


djanosik:
The generic interface will simplify some development scenarios (like adding extension methods). We should keep the non-generic interface as well.

Issue #506, opened 2017-11-28
Remove OnLoadingData delegates from GridViewDataSet


djanosik:
The `OnLoadingData` delegates have introduced deadlocks and other issues. We need to remove them in next major version. As part of that we should refactor functionality related to the `IsRefreshRequired ` property.

Issue #484, opened 2017-10-27
Find a new solution for temporarily fixed GridViewDataSet sync deadlocks


quigamdev:
https://github.com/riganti/dotvvm/pull/481 In this PR, there are temporarily fixed GridViewDataSet sync deadlocks. 
It is necessary to find a better solution than Task.Result.

exyi:
I think that the GridView should not load the data, the load operation should be initialized by ViewModel. This way, the data-set works only when GridView populates it, which means that the ViewModel is dependent on the View, which is pretty messy.

tomasherceg:
Yes, I also don't agree with the concept of `OnLoadingData` delegate, but once it is in the framework, we should keep it and make it work.

exyi:
Do you have any way to make it work correctly? I'm afraid that it's impossible without `Wait`ing the `Task`.

djanosik:
We have removed the `OnLoadingData` delgates from `GridViewDataSet` in #509.

PR #509, opened 2017-11-30
Simplified GridViewDataSet

djanosik:
I've made some changes to simplify the `GridViewDataSet`. It resolves issues #505, #506, #507 and #484.


src/DotVVM.Core/Controls/IPageableGridViewDataSet.cs
```diff
@@ -1,73 +1,39 @@
...
-        /// Navigates to the last page.
+        /// Navigates to the next page if possible.
         /// </summary>
-        Task GoToLastPageAsync();
-
-        /// <summary>
-        /// Navigates to the next page (if possible).
-        /// </summary>
-        [Obsolete("Use GoToNextPageAsync")]
         void GoToNextPage();
```
exyi: Do we need these helpers, when you can get the page from `PagingOptions` and then call `GoToPage`? What about to make this an extension method?

src/DotVVM.Core/Controls/Options/DistanceNearPageIndexesProvider.cs
```diff
@@ -4,21 +4,31 @@
...
+        /// <param name="pagingOptions">The settings for paging.</param>
         public IList<int> GetIndexes(IPagingOptions pagingOptions)
         {
-            return
-                Enumerable.Range(0, pagingOptions.PagesCount)
-                    .Where(n => Math.Abs(n - pagingOptions.PageIndex) <= distance)
-                    .ToList();
+            return Enumerable.Range(0, pagingOptions.PagesCount)
+                .Where(n => Math.Abs(n - pagingOptions.PageIndex) <= distance)
+                .ToList();
```
exyi: `Enumerable.Range(Max(pagingOptions.PageIndex - distance), Min(pagingOptions.PageIndex + distance, pagingOptions.PagesCount))` would look better to me... and it would not have so much sub-optimal complexity...

src/DotVVM.Core/Controls/Options/IPagingOptions.cs
```diff
@@ -24,28 +24,29 @@ public interface IPagingOptions
...
         /// </summary>
         IList<int> NearPageIndexes { get; }
 
         /// <summary>
-        /// Applies the paging settings to the IQueryable object.
+        /// Applies the paging options to the <paramref name="queryable"/> object.
         /// </summary>
+        /// <param name="queryable">The <see cref="IQueryable{T}" /> to modify.</param>
         IQueryable<T> ApplyToQueryable<T>(IQueryable<T> queryable);
     }
```
exyi: This interface is either redundant or misleading IMO. Either it will behave like a normal pager with Index + PageSize + TotalCount or it will do some magic in the `ApplyToQueryable` function and the `PageIndex`, `PageSize` will be completely irrelevant. I don't know if we should really remove it, but I don't like it at all.

src/DotVVM.Core/Controls/Options/SortingOptions.cs
```diff
@@ -22,8 +22,9 @@ public class SortingOptions : ISortingOptions
         public string SortExpression { get; set; }
```
exyi: It's quite unfortunate that this has to be a `string`, but what about to validate it the property exists in the setter.

src/DotVVM.Core/Controls/IPageableGridViewDataSet.cs
djanosik: Yes, we should keep these helpers. But they can be extension methods.

src/DotVVM.Core/Controls/Options/IPagingOptions.cs
djanosik: It shouldn't do any magic. But maybe we can move the logic back to `GridViewDataSet`?

src/DotVVM.Core/Controls/Options/SortingOptions.cs
djanosik: The expression is validated before `IQueryable` is being sorted. But we need to support sorting by multiple expressions. I would rather change this API, but I am not sure it's good to introduce another breaking change here.

src/DotVVM.Core/Controls/Options/IPagingOptions.cs
exyi: If it should only sort in the PageIndex/PageCount style, I think that there is no reason have a interface for that...

src/DotVVM.Core/Controls/Options/SortingOptions.cs
exyi: Hmm, it's just a string, we may introduce special format for multiple expressions...

tomasherceg: I was thinking about adding MultiCriteriaSortingOptions with a collection of additional sort expressions (and directions). Majority of users don't need it, so they can have null in this property.

src/DotVVM.Core/Controls/Options/IPagingOptions.cs
djanosik: Ok, I don't see any reason to keep it there.

src/DotVVM.Core/Controls/Options/SortingOptions.cs
tomasherceg: There are several options: 
1. Use SortingOptions for the first criteria and MultiCriteriaSortingOptions for the other ones (which would allow to use the DataSet even in controls which do not support multi-criteria sorting). We can create extension methods to work with the sort criteria in an unified way. 
2. Use MultiCriteriaSortingOptions for all criteria. In that case, SortingOptions must be null.
3. Or we might add AdditionalSortCriteria collection right in the SortingOptions and forget about MultiCriteriaSortingOptions. 

The first way seems unnecessarily complicated - I would go for 2 or 3. Option 3 means breaking the ISortingOptions interface, however I wouldn't mind so much - I guess nobody except us implemented that.

exyi: Extending the `ISortOptions` interface (3) seems the most sane option for me.

Btw, @tomasherceg you wanted a example where interface does not help with extensibility :P

tomasherceg: OOP sucks, go functional!

exyi: @tomasherceg So, finally? :P

PR #511, opened 2017-12-06
Added default polyfills and their lazy loading

### LINKED ISSUE #495: Include polyfill for ES6 Promise
We need to include polyfill for ES6 Promise in 2.0, because we still need compatibility with IE11. Is it good idea to also include polyfills for common ES6 functions?

adamjez:
Added few polyfills and javascript which detects missing features. If at least one feature is missing, polyfill bundle will be **loaded**. That way browsers, which support all required features (Edge, Firefox, Chrome ..) will not be affected at all (only feature detection). There is a nice article about this topic: https://philipwalton.com/articles/loading-polyfills-only-when-needed/

The PR adds these polyfills :
* **Promise** https://github.com/taylorhakes/promise-polyfill
* **Fetch** https://github.com/github/fetch
* **ParentNode.firstElementChild** [developers.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild) (for whatever reason this function is also missing in my IE 11 testing environment)

To add support for **Iterators** we would have to add a transpiler (http://gu.illau.me/posts/polyfilling-generators/), because the new iterator syntax would trigger syntax errors in a non-supporting browser.

The PR resolves #495


djanosik:
Hmm, `firstElementChild` should be supported in IE11. And I think we don't need polyfill for iterators.

src/DotVVM.Framework/Resources/Scripts/DotVVM.Polyfills.ts
```diff
@@ -0,0 +1,10 @@
+﻿(function () {
+    if (typeof Promise === 'undefined' || !self.fetch) {
+        var resource = <HTMLScriptElement>document.createElement('script');
```
exyi: The cast to `HTMLScriptElement` should not be required by TypeScript, it should know that form the `'script'` in parameter. Are you sure your TS version is up to date?

adamjez: You're right, the cast isn't required. I didn't know that.

adamjez:
@djanosik You're right. The `firstElementChild` is supported by IE11, but it's behavior is a little bit different than in other browsers.
In IE11 `document.firstElementChild` returns `undefined` but in for example Chrome, it returns `html` element. Should I remove the `firstElementChild` polyfill and change the `firstElementChild` call on `document` to `document.body` in `dotvvm.Debug.js` file?

djanosik:
Ok. Thank you!

BTW: I suggest you to not use WebCompiler and BundlerMinifier because they depend on Visual Studio. It's better to use some NPM tasks.

adamjez:
@exyi What do you think? Should I replace BundlerMinifier with NPM taks?

PR #513, opened 2017-12-08
Localizable presenter

exyi:
Added `LocalizablePresenter` class - a utility for localizing specific route based on route parameter or query string


src/DotVVM.Framework/Hosting/LocalizablePresenter.cs
```diff
@@ -0,0 +1,48 @@
...
+            var culture = this.getCulture(context);
+            if (culture != null)
+                CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = culture;
+            return this.nextPresenter(context);
+        }
+
+        public static Func<LocalizablePresenter> BasedOnParameter(string name)
+        {
+            var presenter = new LocalizablePresenter(
+                context => context.Parameters.TryGetValue(name, out var value) && !string.IsNullOrEmpty(value as string) ? new CultureInfo((string)value) : null,
```
adamjez: if `value` is not a valid culture name the statement `new CultureInfo((string)value)` will throw `CultureNotFoundException`. We should put the statement into a `try` block.

exyi: Hmm, I don't think so, silently eating the error may be pretty nasty. I think that displaying Error 500 when you access non-existent culture is not that bad option. Unfortunately, in the presenter it's too late for a "not found" response, you can only do that in a middleware, but you could catch the CultureNotFound exception and rewrite it to 404 response elsewhere.

adamjez: Yes, silently eating the error is pretty nasty, but IMHO 500 Internal Server Error is worse. Error 500 should indicate that something is bad with your code. We can add `culture` constraint but that doesn't solve the problem. Returning 404 would be nice, but IMHO we shouldn't require for user to implement this behavior.

exyi: Maybe I could do a redirect to a default culture. What would you think about that?

adamjez:
That sounds good.

Od: Stanislav Lukeš <notifications@github.com>
Odesláno: Wednesday, December 27, 2017 2:41:58 PM
Komu: riganti/dotvvm
Kopie: Adam Jež; Comment
Předmět: Re: [riganti/dotvvm] Localizable presenter (#513)


@exyi commented on this pull request.

In src/DotVVM.Framework/Hosting/LocalizablePresenter.cs<https://github.com/riganti/dotvvm/pull/513#discussion_r158814865>:

> +            this.getCulture = getCulture;
+            this.nextPresenter = nextPresenter;
+        }
+
+        public Task ProcessRequest(IDotvvmRequestContext context)
+        {
+            var culture = this.getCulture(context);
+            if (culture != null)
+                CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = culture;
+            return this.nextPresenter(context);
+        }
+
+        public static Func<LocalizablePresenter> BasedOnParameter(string name)
+        {
+            var presenter = new LocalizablePresenter(
+                context => context.Parameters.TryGetValue(name, out var value) && !string.IsNullOrEmpty(value as string) ? new CultureInfo((string)value) : null,


Maybe I could do a redirect to a default culture. What would you think about that?

src/DotVVM.Framework/Hosting/LocalizablePresenter.cs
adamjez: That sounds good.

src/DotVVM.Framework/Hosting/LocalizablePresenter.cs
```diff
@@ -10,8 +10,10 @@
 
 namespace DotVVM.Framework.Hosting
 {
+    /// <summary>
     /// A DotVVM Presenter that reads culture by <see cref="getCulture" />,
     /// sets the thread culture and invokes the default dotvvm presenter (obtained from IServiceProvider)
+    /// </summary>
     public class LocalizablePresenter : IDotvvmPresenter
```
exyi: To be honest, I don't like that at all. I don't want to mess the codebase with VisualStudio hacks. The `summary` element is simply totally ugly and it's not required.

* By specification - the specification actually does not force any rules here, the comment just has to be valid XML. There are just recommended tags. (that also means that VS behavior is correct, but in the same way notepad is valid C# IDE)
* By all sane tools I know - Omnisharp and JetBrains rider work.
image
image
And if it does not work with any other sane IDE, it can't be a problem to write a plugin or fork and compile :P

tomasherceg:
Is there any way of telling DotVVM to use this presenter on all routes by default? 
If not, we should add this - maybe something like `config.Routes.DefaultPresenterFactory = ...`.

exyi:
You can use IServiceCollection to register your presenter
implementation, adding a property to configuration seems like
duplicating the DI's work. The problem is, that this presenter is going
to invoke the default one when the culture is set, which will simply
cause stack-overflow.

I think that it does not make much sense to register it globally - this
presenter is intended for per-route usage, if you want a global rule
you can use the default Asp.Net middleware.

tomasherceg:
Is there any type you can register for the default presenter? I don't think `services.AddTransient<DefaultDotvvmPresenter, LocalizablePresenter>()` will work. 
And registering it for just `IDotvvmPresenter` is not wise because of custom presenters.

exyi:
> Is there any type you can register for the default presenter? I don't think `services.AddTransient<DefaultDotvvmPresenter, LocalizablePresenter>()` will work. 
`IDotvvmPresenter`

> And registering it for just `IDotvvmPresenter` is not wise because of custom presenters.
I don't understand what is the problem with it. This way you can register your custom presenter as a default one, or you can use it just locally, I don't see any problem.

tomasherceg:
Oh, I understand - custom presenters can be registered as specific types and the default one will be registered as `IDotvvmPresenter`. OK, we can use it like this, but we should document this behavior.

Issue #463, opened 2017-09-30
RouteTable.Add with DI injected presenter


exyi:
The `RouteTable.Add` method has a `presenterFactory` parameter, that can be used for creating a custom presenter, but it can't easily come from DI as there is no `IServiceProvider` parameter. I think that we should add an overload that would allow that. The presenter injection could then look like:
```C#
routes.Add(..., presenterFactory: ServiceProviderServiceExtensions.GetRequiredService<SomePresenter>)
```

Do you think that it's enough, or one more overload with something like `presenterType: Type` would be useful?

djanosik:
The solution with `presenterFactory` overload feels good enough. We can add `presenterType` later if needed.

tomasherceg:
I think that having the overload with just presenterType is a nice way - also, the custom presenter doesn't need the view path parameter so we can get rid of this parameter.

exyi:
I'd only implement the overload with `presenterFactory`, it's more explicit, not that much verbose and this feature is probably not going to be used much in real projects, so readability is more important than writability :).

Leeemons:
https://github.com/riganti/dotvvm/commit/644cd589932996dc9f05ef6e0800a5e275544da3

exyi:
This does not solve the issue at all, we wanted an overload with the `presenterFactory` factory of type `Func<IServiceProvider, IDotvvmPresenter>` and we were uncertain about the generic form (I don't like it very much...)

Leeemons:
I created PR #514

quigamdev:
Solved by #514

PR #514, opened 2017-12-09
RouteTable.Add overload with generic param of IDotvvmPresenter type

Leeemons:


src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -104,6 +104,21 @@ public void Add(string routeName, string url, string virtualPath, object default
...
+        /// Adds the specified route name.
+        /// </summary>
+        /// <typeparam name="T">Type of presenter</typeparam>
+        /// <param name="routeName">Name of the route.</param>
+        /// <param name="url">The URL.</param>
+        /// <param name="virtualPath">The virtual path of the Dothtml file.</param>
+        /// <param name="defaultValues">The default values.</param>
+        public void Add<T>(string routeName, string url, string virtualPath, object defaultValues = null) where T : IDotvvmPresenter
+        {
+            Func<IDotvvmPresenter> presenterFactory = presenterFactory = () => configuration.ServiceProvider.GetRequiredService<T>();
```
exyi: As @djanosik noticed, this does not work with request-scoped services.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -104,6 +104,21 @@ public void Add(string routeName, string url, string virtualPath, object default
...
 
+        /// <summary>
+        /// Adds the specified route name.
+        /// </summary>
+        /// <typeparam name="T">Type of presenter</typeparam>
+        /// <param name="routeName">Name of the route.</param>
+        /// <param name="url">The URL.</param>
+        /// <param name="virtualPath">The virtual path of the Dothtml file.</param>
+        /// <param name="defaultValues">The default values.</param>
+        public void Add<T>(string routeName, string url, string virtualPath, object defaultValues = null) where T : IDotvvmPresenter
```
exyi: The `virtualPath` parameter is completely redundant when the presenter is specified.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -104,6 +104,21 @@ public void Add(string routeName, string url, string virtualPath, object default
...
         }
 
+        /// <summary>
+        /// Adds the specified route name.
+        /// </summary>
+        /// <typeparam name="T">Type of presenter</typeparam>
+        /// <param name="routeName">Name of the route.</param>
+        /// <param name="url">The URL.</param>
+        /// <param name="virtualPath">The virtual path of the Dothtml file.</param>
+        /// <param name="defaultValues">The default values.</param>
```
exyi: When I look at the parameter name and the comment, I don't see any single bit of information that the comment gives me...

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
exyi: As I told you in #463, we wanted a method with `Func<IServiceProvider, IDotvvmFactory>` parameters, this overload should be just a sugar for that (-> I'd make that an extension method). And I actually think that we don't need this overload, the presenterFactory can be quite easily used for this, see the discussion in #463

Leeemons: Why dont you like generic type?

exyi:
> Why dont you like generic type?
It's not about generic types, the point that we need the overload with
the factory function, so this one becomes redundant.

quigamdev:
I changed signature for DotvvmRouteTable.Add. There is breaking change, that requires change of docs.
@Leeemons  Please add tests for scoped lifecycle.

src/DotVVM.Framework/Routing/RouteBase.cs
```diff
@@ -184,7 +184,7 @@ public static void AddOrUpdateParameterCollection(IDictionary<string, object> ta
         /// <summary>
         /// Processes the request.
         /// </summary>
-        public abstract IDotvvmPresenter GetPresenter();
+        public abstract IDotvvmPresenter GetPresenter(IServiceProvider provider);
```
exyi: Btw: The comment is completely misleading.

quigamdev: @Leeemons  Please change the comment.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -93,15 +93,37 @@ public void AddGroup(string groupName, string urlPrefix, string virtualPathPrefi
...
+        /// <summary>
+        /// Adds the specified route name.
+        /// </summary>
+        /// <param name="routeName">Name of the route.</param>
+        /// <param name="url">The URL.</param>
+        /// <param name="presenterType">The presenter factory.</param>
+        /// <param name="defaultValues">The default values.</param>
+        public void Add(string routeName, string url, Type presenterType, object defaultValues = null)
+        {
+            Func<IServiceProvider, IDotvvmPresenter> presenterFactory = provider => (IDotvvmPresenter)provider.GetService(presenterType);
```
tomasherceg: I would add the following check and provide the user with a meaningful error message:

`if (!typeof(IDotvvmPresenter).IsAssignableFrom(presenterType)) throw ...`

quigamdev: @tomasherceg  Please check the message text.

tomasherceg: I cannot see any message in the code, however I would use something like this:

"The presenter type {presenterType.ToString()} used in the route '{routeName}' does not implement IDotvvmPresenter!"

PR #519, opened 2017-12-13
Added minified versions of DotVVM.js and globalize.js

adamjez:
Added minified versions of `DotVVM.js` and `globalize.js`. 
Added watch and `build` npm tasks. 

I'm creating this PR mainly to discuss whether we want to use the **non-minified** versions of the scripts when `debug` is set to `true`. Currently we have a mechanism that is usable only in the dotvvm repository, because it uses file paths, called ` SetEmbeddedResourceDebugFile`.


djanosik:
Using `SetEmbeddedResourceDebugFile` is good enough. Users can clone DotVVM repository and set the correct path in their applications if they need to debug our scripts. We don't need to embed non-minified resources in the assembly.

src/DotVVM.Framework/package.json
```diff
@@ -4,10 +4,17 @@
...
+    "onchange": "^3.3.0"
   },
   "scripts": {
-    "bundle": "npm run bundle:js",
-    "bundle:js": "uglifyjs -o Resources/Scripts/Polyfills/polyfill.bundle.js -c -m -- Resources/Scripts/Polyfills/promise.js Resources/Scripts/Polyfills/fetch.js"
+    "build": "npm run compile:ts && npm run minify:dotvvm:js && npm run bundle:polyfill:js",
+    "compile:ts": "tsc",
+    "minify:dotvvm:js": "uglifyjs -o Resources/Scripts/DotVVM.min.js -c -m -- Resources/Scripts/DotVVM.js",
+    "bundle:polyfill:js": "uglifyjs -o Resources/Scripts/Polyfills/polyfill.bundle.js -c -m -- Resources/Scripts/Polyfills/promise.js Resources/Scripts/Polyfills/fetch.js",
+    "watch": "concurrently --kill-others \"npm run watch:ts\" \"npm run watch:js\"",
```
djanosik: What is the advantage of using `concurrently`?

adamjez: `concurrently` runs multiple tasks at once which is needed because `watch:ts` task never ends and therefore it's not possible to use `&&`

djanosik: Have you tried `npm run watch:ts | npm run watch:js`?

adamjez: It works too, but the first task `stdout` is sent to the second task `stdin`. It might lead to unpredictable behavior in a future.

PR #521, opened 2017-12-15
Fixed RoleView.cs for anonymous users

ganzikk02:
When HideForAnonymousUsers was set to false and user was anonymous (not authenticated) the User.IsInRole does not work as you may expect and throws exception "System.SystemException: The trust relationship between this workstation and the primary domain failed.". Solution I found to solve this issue was taken from SO: https://stackoverflow.com/a/22767760

### DIFF
--- a/src/DotVVM.Framework/Controls/ClaimView.cs
+++ b/src/DotVVM.Framework/Controls/ClaimView.cs
@@ -88,11 +88,11 @@ namespace DotVVM.Framework.Controls
         protected internal override void OnInit(IDotvvmRequestContext context)
         {
             var user = context.HttpContext.User;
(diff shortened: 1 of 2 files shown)


adamjez:
I don't really think that It's bug in our code. The exception message `The trust relationship between this workstation and the primary domain failed.` suggest some kind of problem in windows configuration. 

But that additional check doesn't break anything. UI tests passed successfully. What do you think @djanosik @exyi ? Should we merge it?

djanosik:
I think we can merge it.

exyi:
Ok, I'm merging it, it is harmless and if it triggers less errors in 3rd-party libraries, why not...

PR #523, opened 2017-12-20
Fixed generation of url with optional parameter and without default value

adamjez:
**Fixed bug in generating url and added UI tests**.

Currently there is a **bug** with generating url with optional parameter and without specifying default values
(client side rendering)

e.g. Url `ControlSamples/Repeater/RouteLink/{Id?}` generates `/ControlSamples/Repeater/RouteLink/%7BId?}`

There is still different behavior when rendering at client/server side:
* client: `ControlSamples/Repeater/RouteLink/`
* server: `ControlSamples/Repeater/RouteLink`


adamjez:
Fixed the url generation at client side when a route **contains optional parameter**.

**Current implementation** of dotvvm routes and url generation a little bit messy. Specification of routes format doesn't exists (except source code). There exists two implementations (client and server side) which generates urls.

To correct it I would suggest these steps:
1. Create a **route format specification** - valid route formats with lot of examples
2. Create mechanism to share implementations of url generation. It may consists of translating `UrlBuilders` to javascript dynamically or sharing same `regex` which generates url.

I would like to hear your opinions: @exyi @djanosik @quigamdev

exyi:
I think that you are right with the specification, it should contain *precise* grammar of the route strings + the examples explaining the grammar. Actually, I'm afraid that we will need to write specification of almost everything - technically, there are no bugs in dotvvm, because the source code is the specification :)

About the translation fo `UrlBuilders` - I think that it is a good idea, but the JS compiler simply isn't mature enough - naively translating them now would generate ugly expression like `(function (p) { return (function(_) { return "/rr" })(p) + ...)({Id: SomeExpression...})` and it would require adding support for basic `Dictionary` operations.

djanosik:
It sounds good to me. But we should file another issue to solve this.

exyi:
Merging this PR as a bug fix. I have created separate issues for the translation of UrlBuilders #526 and specification https://github.com/riganti/dotvvm-docs/issues/49

PR #529, opened 2017-12-30
 Redesigned resource link controls

exyi:
You can now explicitly say "I want to have the resources rendered here" using <dot:HeadResourceLinks /> or <dot:BodyResourceLinks />. The resources are rendered on the first place where this control occurs.

HtmlGenericControl renders this control implicitly if the TagName is body or head (I have moved the logic from DefaultOutputRenderer, I think that it should improve performance)

ResourceManager now knows if the resources have been already rendered and it will throw an exception when a new one is added. This is potentially a runtime breaking change, but it should only throw on code that is actually incorrect.

### DIFF
--- a/src/DotVVM.Framework/Controls/HtmlGenericControl.cs
+++ b/src/DotVVM.Framework/Controls/HtmlGenericControl.cs
@@ -124,7 +124,7 @@ namespace DotVVM.Framework.Controls
         }
 
         /// <summary>
-        /// Verifies that the control hasn't any HTML attributes or Visible or DataContext bindings set.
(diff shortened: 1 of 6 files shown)


exyi:
It seems to be complete performance win according to riganti/dotvvm-benchmarks (the strange triplets are "minimal / average / maximal" change in the value):
image
There is some noise in memory column, some tests had 0 allocated on the Base version, in reality there is basically no change.
The request tests have on average 70% better wall clock time, but it is a bit biased as most of the tests are pretty mininal pages. But when I tried to pick few biggest pages (acording to median time and Gen0 garbage collections), there is still always at least few percents improvement.

exyi:
This one is also already 3 weeks old, merging it now.

PR #534, opened 2018-01-23
Added a script to complete xml documentation with <summary> tags

exyi:
It will allow us to write comments without the `<summary>` tags - the script completes them when the package is published.

I simply don't like the verbosity of XML comments with the `summary` tags, it feels much nicer to write comments without the mess around. And it's not required by 

To be honest, I don't like that at all. I don't want to mess the codebase with VisualStudio hacks. The `summary` element is simply totally ugly and it's not required:
* By specification - the specification actually does not force any rules here, the comment just has to be valid XML. There are just recommended tags that have defined meaning.
* By all sane tools I know - Omnisharp and JetBrains Rider work.
image
image

Unfortunatelly VisualStudio is not one of the sane tools, so I have written this script to mitigate the issues.

I have not removed any `<summary>` tags in the PR, but I'm totally going to do that when it's merged into master (just want to avoid a ton merge conflicts)

cc @adamjez @djanosik @tomasherceg What do you think?

### DIFF
--- a/src/DotVVM.Core/ViewModel/AllowStaticCommandAttribute.cs
+++ b/src/DotVVM.Core/ViewModel/AllowStaticCommandAttribute.cs
@@ -2,6 +2,7 @@
 
 namespace DotVVM.Framework.ViewModel
 {
+    /// Allows remote invocation of this method from a staticCommand binding
     public class AllowStaticCommandAttribute: Attribute
     {
     }
--- a/src/Tools/add-summary-comments.fsx
+++ b/src/Tools/add-summary-comments.fsx
@@ -1,0 +1,40 @@
+#!/usr/bin/fsharpi
+
+#r "System.Xml.Linq"
+open System
+open System.Xml.Linq
+let rootElements = Set.ofList [ "example"; "code"; "exception"; "include";  "returns"; "param"; "permission"; "remark"; "seealso"; "value"; "typeparam" ]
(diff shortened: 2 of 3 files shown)


djanosik:
There is no hacking of VS IDE involved. We will simply not see contents of "summary" comments without `<summary>` tags in VS IDE while working on DotVVM. But all tags are automagically added during build on CI. So consumers of NuGet packages won't be affected in any way.

However. I am not sure whether to accept this PR. Most of us work in VS and it's sometimes very helpful to see summaries of our own methods.

@exyi Have you filed an issue anywhere? Will VS ever support comments without `<summary>` tags?

exyi:
The problem is that the C# spec does not specify any specific format of the comments, it should be just valid XML. Then there are few "recommended" tags to support. I don't think VS is going to support that anytime soon, the same way it does not support `<inheritdoc/>` and the same way notepad does not support C# intellisense...

The removal of documentation comments may only cause little problems when working directly with dotvvm source code, as far as I know we don't do that on any our project. And if you are working on dotvvm itself, you will need to go to definition anyway, the comments on the more internal parts are not very complete :P

Long long time ago I filed an issue to roslyn repo https://github.com/dotnet/roslyn/issues/16470 when I did not know this is not a C# problem but a Visual Studio problem. Noone even noted that this is just VS glitch :(

Issue #442, opened 2017-08-20
Easier way to write controls


exyi:
If you want to write a control that is more complex than a `div` that adds one attribute, it is usually quite messy and long code, that often has to utilize a badly documented API. You have to manually check if there is binding in each property and decide if you want to render a value, a binding or both. And there is quite a lot of cases, that influence the rendering, that may break your control and that you have to care about - control is in client-side template (so *some* DataContent in the hierarchy is null and *some* bindings may return unexpected null, I haven't seen control that handles this correctly...), server/client side rendering, value/resource binding/value binding in the property, binding in non-nullable property may throw NullReferenceException...

I think that most of the complexity could be handled by built-in dotvvm controls if you could use them easily, so I propose a simple sugar for making dotvvm controls that would allow you put the logic into one function. It would take the `DotvvmPropertie`s as parameters (to reduce boilerplate code) and return a tree of components:

```C#
protected override DotvvmControl GetContents(
     ValueOrBinding<string> myProperty,
     Dictionary<string, object> attributes
)
{
     return new HtmlGenericControl(div) {
         Attributes = attributes,
         Content = { new Literal() { Text = myProperty } }
     };
}
```

To help you with creating the content, I think we could a helper function `DotvvmControl CreateControls(ParametrizedString str)` that would take dothtml string with parameters and return the tree:

```C#
protected override DotvvmControl GetContents(
     ValueOrBinding<string> myProperty,
     Dictionary<string, object> attributes
)
{
     return CreateControls($@"
           <div {attributes}>{myProperty}</div>
     ");
}
```

What do you think?
cc @martindybal  @djanosik

djanosik:
@exyi Have you seen the API we've created when building Business Pack?

symbiogenesis:
This idea would definitely help for your ReactBridge. Trying to get complicated React controls like the React Data Grid to work would be very difficult without this.

exyi:
@symbiogenesis I'm not sure if this is going to help you. It basically only brings the advantages of dothtml markup to the code-only controls, and is actually fairly inspired by React, so it may look a bit familiar ;)

However, if you'd like to ReactBridge for a complicated control (that requires complex configuration object, special content, or something like that) I'd recommend you to create a specialized React control that will handle the configuration for you and use only that from dothtml markup. Something like this:

```JSX
// define the React part of your page
function OrderWithCustomerGridView(props) {
    const data = props.data; // just take data from the props
[...]

PR #535, opened 2018-01-23
Easier way to write controls

exyi:
Implemented `ValueOrBinding<T>` and `CompositeControl` types + various tweaks in the framework to make this possible. It's the first part of #442

`ValueOrBinding<T>` is a discriminated union of `IBinding` and `T`, with few helper methods. It's required when you want to carry DotvvmProperty in one variable - which is exactly what I wanted to do - put every dotvvm property into a simple function parameter.

`CompositeControl` is a control base class (formerly known as `EpicCoolControl`) It automatically creates and maps dotvvm properties from function parameters, without the need to declare them in the boilerplate way. The function simply gets the properties and returns dotvvm control tree which should be rendered - which implies that this way of rendering controls is not designed to replace the legacy one, as the returned control need to be rendered in some way ;). I think that each method has it's strong/weak spots and having both of them will help us cover more use-cases. Little example:

```C#
public class MyControl: CompositeControl {
     public static DotvvmControl GetContents(
          // this property is required and supports both bindings and hard-coded values
          ValueOrBinding<string> myProperty,
          // this one is optional and does not allow value binding (resource binding are allowed, it's simply evaluated before executing this method)
          string tagName = "div"
     )
     {
          return new HtmlGenericControl(tagName) {
             Children = { new Literal(myProperty) }
         };
   }
}
```

The `GetContents` method is found by name, it can be either static or instance and it has to return `DotvvmControl` or `IEnumerable<DotvvmControl>`. It's invoked `OnLoad` (only once, so evaluated bindings will not be updated, you should pass the evaluation to "primitive controls"). And it's still a DotvvmControl, so you can declare dotvvm properties in the old way, access attached or inherited properties, do something on lifecycle events and so on. 

It also supports all the attributes you'd expect from dotvvm property on the parameter - data context change, markup options and so on. And property groups are also supported, see [the sample](https://github.com/riganti/dotvvm/blob/easier-controls/src/DotVVM.Samples.Common/Controls/EpicCoolControlSample.cs)

### Control Capabilities

This PR also includes a feature called "control capabilities". It's essentially a group of properties that can be declared once and then reused. It's declared as a data class with the desired properties, see https://github.com/riganti/dotvvm/pull/535/files#diff-896c261fa17422d2cbad9aa5ee21140eR328 for an example. Then, when you want to support the capability, you just register a single property and pass the entire class to the "primitive control" or some helper function that will actually render it.

This feature is designed to replace the stupid antipattern of using inheritance to for only getting some common properties (see `ItemsControl`, ...)


exyi:
Note: It will have to be supported in the Visual Studio extension - we will either have to dump properties from DotVVM.Compiler or add support for these controls. cc @Mylan719

tomasherceg:
We've discussed an interesting thing with @quigamdev - what about the inheritance of composite controls? 

Naturally, the `GetContent` in the derived class will need to request all properties of the base class, and I think we should not shadow the base class properties - we should "de-duplicate" them, possibly by name.

```
public class Parent : CompositeControl {
  public static DotvvmControl GetContents(
    ValueOrBinding<string> prop1
  ) {
    ...
  }
}
public class Child : Parent {
  public static DotvvmControl GetContents(
    ValueOrBinding<string> prop1,
    ValueOrBinding<string> prop2
  ) {
 
    ...
       Parent.GetContents(prop1)   // or new Parent(prop1)
    ...

  }
}
```

I think that we should register prop1 only for the Parent class and not for the Child class (and throw an exception if there is a same-named property in the Child class that has a different type).

It would be very useful in Bootstrap for DotVVM. 
For example, we have plenty of controls that derive from [FormGroup](https://www.dotvvm.com/docs/controls/bootstrap4/FormGroup/2.0): `TextBoxFormGroup`, `ComboBoxFormGroup` etc. 
All of these controls have the `Label` property, and I think we have some other control in which we browse all form groups in the form, and need to access the labels universally.

quigamdev:
Merging this PR to `alpha-testing` for further testing and revision of the design.

Mylan719:
We seems to have en issue here with the newly added HtmlCapability property registered under the name "Html" on the HtmlGenericControl. The control HtmlLiteral is based on HtmlGenericControl and has own property named "Html".

src/DotVVM.Framework.Tests/ControlTests/CompositeControlTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                <cc:WrappedHtmlControl TagName=div> Something here </cc:WrappedHtmlControl>
+                "
+            );
+
+            check.CheckString(r.FormattedHtml, fileExtension: "html");
+        }
+
+
+        // [TestMethod]
+        // public async Task CommandBinding()
```
quigamdev, line 52: Please remove the commented code

src/DotVVM.Framework/Binding/ValueOrBinding.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public ValueOrBinding<T2> UpCast<T2>()
+            where T2 : T =>
+            this.binding != null ?
+            new ValueOrBinding<T2>(this.binding) :
+            new ValueOrBinding<T2>((T2)this.value!);
+
+        public ParametrizedCode GetParametrizedJsExpression(DotvvmBindableObject control, bool unwrapped = false) =>
+            ProcessValueBinding(control,
+                value => new ParametrizedCode(JsonConvert.SerializeObject(value), OperatorPrecedence.Max),
```
exyi, line 80: Serialization option that we use elsewhere.

src/DotVVM.Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
```diff
@@ -652,21 +652,21 @@ public void ProcessControlContent(IAbstractControl control, IEnumerable<DothtmlN
                     content.Add(node);
                 }
             }
-            if (control.Metadata.DefaultContentProperty != null)
+            if (control.Metadata.DefaultContentProperty is IPropertyDescriptor contentProperty)
```
exyi, line 655:
```suggestion
            if (control.Metadata.DefaultContentProperty is {} contentProperty)
```

src/DotVVM.Framework/Controls/CompositeControl.cs
```diff
@@ -0,0 +1,88 @@
...
+                var arguments = method.GetParameters().Select(initializeArgument);
+
+                if (!controlInfoCache.TryAdd(controlType, new ControlInfo { GetContentsMethod = method, Properties = arguments.ToImmutableArray() }))
+                    throw new Exception("no");
+            }
+        }
+
+        protected internal override void OnLoad(IDotvvmRequestContext context)
+        {
+            this.Children.Clear();
```
exyi, line 72: Throw exception when not empty.

src/DotVVM.Framework/Controls/DotvvmBindableObject.cs
```diff
@@ -85,16 +85,16 @@ public T GetValue<T>(DotvvmProperty property, bool inherit = true)
         internal object? EvalPropertyValue(DotvvmProperty property, object? value)
         {
             if (property.IsBindingProperty) return value;
-            if (value is IBinding)
+            while (value is IBinding binding) //TODO: MERGE CONFLICT
```
exyi, line 88: This is badly merged

src/DotVVM.Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -22,21 +22,166 @@ public static class DotvvmBindableObjectHelper
...
             return DotvvmProperty.ResolveProperty(property.DeclaringType!, property.Name) ?? throw new Exception($"Property '{property.DeclaringType!.Name}.{property.Name}' is not a registered DotvvmProperty.");
         }
+        public static DotvvmProperty GetDotvvmProperty<TControl>(this TControl control, string propName)
+            where TControl : DotvvmBindableObject
+        {
+            return DotvvmProperty.ResolveProperty(typeof(TControl), propName) ?? throw new Exception($"Property '{typeof(TControl)}.{propName}' is not a registered DotvvmProperty.");
+        }
 
-        public static TControl SetBinding<TControl, TProperty>(this TControl control, Expression<Func<TControl, TProperty>> prop, IBinding? binding)
+        public static TControl SetProperty<TControl, TProperty>(this TControl control, Expression<Func<TControl, TProperty>> prop, IBinding? binding)
```
exyi, line 31: Leave the old SetBinding and SetValue function and make them obsolete.

src/DotVVM.Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -22,21 +22,166 @@ public static class DotvvmBindableObjectHelper
...
+            return control;
+        }
+
+        public static TControl SetAttribute<TControl, TProperty>(this TControl control, string attribute, ValueOrBinding<TProperty> value)
+            where TControl : IControlWithHtmlAttributes
+        {
+            return SetAttribute(control, attribute, value.BindingOrDefault ?? value.BoxedValue);
+        }
+
+        public static TControl SetAttribute<TControl, TProperty>(this TControl control, string attribute, ValueOrBinding<TProperty>? value)
```
exyi, line 153:
```suggestion
        public static TControl SetAttribute<TControl, TValue>(this TControl control, string attribute, ValueOrBinding<TValue>? value)
```

src/DotVVM.Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -22,21 +22,166 @@ public static class DotvvmBindableObjectHelper
...
+            {
+                var c = DotvvmCapabilityProperty.Find(typeof(TControl), typeof(TCapability), prefix);
+                if (c is null)
+                    throw new Exception($"Capability {prefix}{typeof(TCapability)} is not defined on {typeof(TControl)}");
+                c.SetValue(control, capability);
+            }
+            return control;
+        }
+
+        public static T AppendChildren<T>(this T control, IEnumerable<DotvvmControl>? children) where T : DotvvmControl
```
exyi, line 172: Params overload.

exyi: Allow nulls in the collection

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -99,6 +126,29 @@ public bool Visible
...
             DotvvmProperty.Register<bool, HtmlGenericControl>(t => t.Visible, true);
 
+
+        public HtmlCapability HtmlCapability
+        {
+            get => (HtmlCapability)this.GetValue(HtmlCapabilityProperty)!;
+            set => this.SetValue(HtmlCapabilityProperty, value);
+        }
+        public static readonly DotvvmCapabilityProperty HtmlCapabilityProperty =
+            DotvvmCapabilityProperty.RegisterCapability("HtmlCapability", typeof(HtmlGenericControl), typeof(HtmlCapability),
```
exyi, line 136: Add overload with generic argument like we have for DotvvmProperty

src/DotVVM.Framework/Controls/TextOrContentCapability.cs
```diff
@@ -0,0 +1,41 @@
...
+            if (text is object && !"".Equals(text))
+                return new TextOrContentCapability { Text = ValueOrBinding<string>.FromBoxedValue(text) };
+            else
+                return new TextOrContentCapability { Content = control.Children.ToList() };
+        }
+
+        public void WriteToChildren(DotvvmControl control, DotvvmProperty textProperty)
+        {
+            control.SetProperty(textProperty, Text);
+            control.Children.Clear();
```
exyi, line 28: Throw is not empty.

src/DotVVM.Samples.Common/Controls/CompositeControlSample.cs
```diff
@@ -0,0 +1,48 @@
...
+            [CollectionElementDataContextChange(1)]
+            IValueBinding<string> titleBinding,
+
+            [ControlPropertyBindingDataContextChange("DataSource")]
+            [CollectionElementDataContextChange(1)]
+            IValueBinding<int?> numberBinding,
+
+            HtmlCapability html,
+
+            [DotvvmControlCapability(prefix: "inner-li:")]
```
exyi, line 27: Make the : separator default

PR #555, opened 2018-02-05
Add JsonIgnore atribute to make ResolvedTypeDescriptor deserializible

Mrnda:
This change will make DotvvmConfiguration deserializable when any ExtensionParameter with ResolvedTypeDescriptor is present there


exyi:
FYI: We have decided with @Mrnda that it will be better solved by a custom JsonConverter as the information in `Type` is actually needed in VS Extension.

PR #556, opened 2018-02-07
Added default values to ParametrizedCode's parameters.

exyi:
It allows data-context shifting for command/staticCommand bindings.
It should also improve performance of value binding rendering a bit.

### DIFF
--- a/src/DotVVM.Framework/Binding/Expressions/CommandBindingExpression.cs
+++ b/src/DotVVM.Framework/Binding/Expressions/CommandBindingExpression.cs
@@ -49,8 +49,11 @@ namespace DotVVM.Framework.Binding.Expressions
 
             public class Methods
             {
-                public CommandJavascriptBindingProperty CreateJs(IdBindingProperty id) =>
-                    new CommandJavascriptBindingProperty(CreateJsPostbackInvocation(id.Id));
+                public CommandJavascriptBindingProperty CreateJs(IdBindingProperty id, ExpectedTypeBindingProperty expectedType = null) =>
+                    new CommandJavascriptBindingProperty(CreateJsPostbackInvocation(
+                        id.Id,
+                        needsCommandArgs: expectedType?.Type?.GetDelegateArguments()?.Length != 0
+                    ));
 
                 public ExpectedTypeBindingProperty GetExpectedType(AssignedPropertyBindingProperty property = null)
                 {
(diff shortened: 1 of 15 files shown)


src/DotVVM.Framework/Binding/Expressions/CommandBindingExpression.cs
```diff
@@ -49,8 +49,11 @@ public class OptionsAttribute : BindingCompilationOptionsAttribute
 
             public class Methods
             {
-                public CommandJavascriptBindingProperty CreateJs(IdBindingProperty id) =>
-                    new CommandJavascriptBindingProperty(CreateJsPostbackInvocation(id.Id));
+                public CommandJavascriptBindingProperty CreateJs(IdBindingProperty id, ExpectedTypeBindingProperty expectedType = null) =>
+                    new CommandJavascriptBindingProperty(CreateJsPostbackInvocation(
+                        id.Id,
+                        needsCommandArgs: expectedType?.Type?.GetDelegateArguments()?.Length != 0
```
djanosik: It won't work when `GetDelegateArguments()?.Length` returns `null`.

src/DotVVM.Framework/Compilation/Javascript/ParametrizedCode.cs
```diff
@@ -58,18 +75,38 @@ public string ToString(Func<object, CodeParameterAssignment> parameterAssignment
...
+        {
+            if (this.evaluatedDefault != null)
+                return this.evaluatedDefault;
+            return ToString(_ => default);
         }
 
         /// <summary>
         /// Assigns parameters and return new ParametrizedCode. If parameter is not assigned, it is copied to the resulting parameter. Assigner can also replace parameter by script that contains another parameters.
         /// </summary>
         public ParametrizedCode AssignParameters(Func<object, CodeParameterAssignment> parameterAssignement)
```
djanosik: I think you should change argument type to `CodeSymbolicParamerer` (btw: typo in name).

djanosik: BTW: Is there a nice way to find out what parameters were actually assigned? I can do it in `parameterAssignement` (typo in name, should be `parameterAssignment` without `e`) but it isn't really nice.

src/DotVVM.Framework/Binding/Expressions/CommandBindingExpression.cs
exyi: It will - when it's null, it's not equal to `0`, so command args are required.

djanosik: Ok. Then I don't get what `needsCommandArgs` really means.

exyi: The point is to only include the commandArgs when they are required. But I have changed implementation so it also supports "strict mode" when it will throw an exception unless the commandArgs are specified.

src/DotVVM.Framework/Compilation/Javascript/ParametrizedCode.cs
exyi: No, there is no way to get assigned parameters, they are wiped away when assigned (only the resulting string is left). If you need more powerful API you will need the Javascript AST

djanosik: Ok. I've realized I can look what params are defined in `ParametrizedCode` before assigning them.

exyi:
According to our benchmarks, this changeset seems to be a noticable improvement. 43.8% is the average improvement on GET request, the "real-world-scenario" test is improved by 14.4%.

Issue #554, opened 2018-02-02
Finish REST API bindings and Swagger generation


tomasherceg:
I have found several issues with REST API bindings that should be resolved before the 2.0 release. 

### Swagger generation and casing issues

Most of the REST APIs return objects with camel case notation of property names. 
When NSwag generates C# API client, the properties are converted to Pascal case. I think it seems natural for C# users and we should preserve this behavior. 
The TypeScript client and the client side uses camel case however.  
NSwag generates `JsonProperty` attributes in the DTOs, which haven't been supported until my commit https://github.com/riganti/dotvvm/commit/fb991f80523b3dcd911e584fa42eeddcb581ab95. 
This works well so far.

But when we use `GridViewDataSet` or some other DotVVM types (`SortingOptions`, `PagingOptions`) in the API, this behavior is not wanted. 
I have added extensions for `Swashbuckle`, the most popular package for generating Swagger files for .NET REST APIs. This extension adds an extra property `x-dotvvm-returnsDataSet` to indicate that the API method returns `GridViewDataSet`. 
However, we could make this behavior more generic. I was thinking about adding something like `x-returns-knownType="FullTypeName"`. 
The Swashbuckle extensions could be configured to detect known types and use this Swagger metadata to provide the hint for the code generator to reuse existing type.
In this case, we also need to keep the casing as is for such types. 

In the NSwag generator, these known types should not emit DTO classes. Instead, they should bind to existing types.

The same logic should be used for method arguments - we should add something like `x-knownType="fullTypeName"`.

Additinally, the DotVVM Swashbuckle extension adds a custom attribute called `AsObject` which allows to use complex types in GET requests. The code generator doesn't explode the complex type and keeps it as it was used on the server side. This helps a lot when the user is passing `PagingOptions` and `SortingOptions` as GET method parameters and allows the method signature to be easy. 
This behavior should be preserved and we may allow to turn on the `AsObject` attribute for these types by default. Again, this could be configured when the user is adding the Swashbuckle extension. 

### HTTP methods in C# API client

We need to extend NSwag generator to add HTTP methods in the C# client. Without that, we are not able to detect whether the method reads or changes the data. 
We should introduce a custom attribute that would be used in the C# client on each method to tell DotVVM, what HTTP method is used. Then, we should use this information to emit correct API calling code - see the TODO comment here https://github.com/riganti/dotvvm/commit/7e07e20b23165fdf93a4ddc7074cdf6f0fbb62c0

PR #561, opened 2018-02-12
Clients REST API generation

adamjez:
**_Do not merge: Work In Progress_**

This PR is introducing two new parameters to Open API specification generated with Swashbuckle: `x-dotvvm-known-type` and `x-dotvvm-name`. These parameters provide required information for NSwag when generating **TypeScript** and **C#** clients. 

* Specifically `x-dotvvm-known-type` is present for types that exist in both **REST API project** and **DotVVM project** so NSwag doesn't generate DTOs for these types. Generic types get special treatment.
* `x-dotvvm-name` specifies *names* for known types.

Additionally, there is new attribute `HttpMethod` to annotate client C# methods with information about used http method.

Currently, It's all implemented only for AspNetCore. I plan to port it when the following issues are resolved.

Unresolved issues:
* Known types are provided in `DotvvmApiKnownTypesOptions` as `Type[]`. Filters get this options object via `IOptions` so users can add their own types. DotVVM types are _hardcoded_ into the options object, because in method `EnableDotvvmIntegration` I don't have access to service provider.
* `DotVVM.Framework.Api.Swashbuckle.AspNetCore` project is only referencing `DotVVM.Core` so logic for determining property name when serializing in `DotvvmViewModel` can't be shared and therefore is copied inside API project. 

Resolves https://github.com/riganti/dotvvm/issues/554


src/DotVVM.Framework/Compilation/BindingRequiredResourceVisitor.cs
```diff
@@ -23,61 +23,57 @@ public BindingRequiredResourceVisitor(ControlResolverMetadata requiredResourceCo
...
 
         public override void VisitPropertyBinding(ResolvedPropertyBinding propertyBinding)
         {
-            var r = propertyBinding.Binding.Binding.GetProperty<RequiredRuntimeResourcesBindingProperty>(ErrorHandlingMode.ReturnNull);
-            if (r != null)
-                requiredResources = requiredResources.Union(r.Resources);
+            var resourceProperty = propertyBinding.Binding.Binding.GetProperty<RequiredRuntimeResourcesBindingProperty>(ErrorHandlingMode.ReturnNull);
+            if (resourceProperty != null)
+            {
+                requiredResources = requiredResources.Union(resourceProperty.Resources);
```
exyi: I'm sorry, but this change does not make it more readable, I actually struggle to visually recognize. I'd argue that single-letter name like `r` is completely reasonable in this context.

```
resourceProperty and requiredResources
```

adamjez: I admit that in this situation, more descriptive naming of variable doesn't add too much, but in the future you can add more statements to the function and from single-letter name like `r` you will not be able to determine what the variable represents. If you struggle with recognizing `resourceProperty` and `requiredResources` I can change it to `property` if you find to more recognizable.

exyi: That would be better for visual recognision but even less descriptive than `r`. In this context it's totally unimportant that it's a binding property, the important fact is that it contains the  resource requirements of this binding.

adamjez: ok, so in this situation I'd choose `requiredResourceProperty`. @djanosik what do you think?

adamjez:
I ported client generation to Owin projects. Unfortunatly I can't run DI tests on this PR because branch is located in my own repo. I ran all tests locally and one test is failing: `JavascriptCompilation_Api_DateParameter` but I'm not aware about any changes in this area.

About the issues mentioned in the main post. 
* I moved the logic for determining property name in ViewModel into separate class which is in **DotVVM.Core** project.
* I let DotVVM known types _hardcoded_ in the `DotvvmApiOptions` because I didn't come with better solution.
* In Swashbuckle integration for Owin I cannot use `IOptions` wrapper because swashbuckle filters are not resolved via DI. User can add its own known types via `Action<DotvvmApiOptions>` delegate which is parameter in `EnableDotvvmIntegration` configuration method. ThirdParty libs can add theirs known types with extensions method on `DotvvmApiOptions`.

PR #565, opened 2018-02-21
IDotvvmStartup signature changed 

quigamdev:
These changes are needed for Dotvvm.Compiler operability. We cannot evaluate Configure method in  Startup.cs everytime compilatior is started. This could cause demages in DB, or just execute something unsafe. So we decided to change structure of IDotvvmStartup to be able mark (register) services that are necessary for dotvvm compilation. 

IDotvvmStartup - added ConfigureServices needed for dotvvm compiler - BREAKING CHANGE
Moved generic parameter from UseDotVVM() to AddDotVVM on aspnet core - BREAKING CHANGE

### DIFF
--- a/src/DotVVM.Compiler/OwinInitializer.cs
+++ b/src/DotVVM.Compiler/OwinInitializer.cs
@@ -42,7 +42,7 @@ namespace DotVVM.Compiler
             var startupClass = webSiteAssembly.CustomAttributes?.Where(s => s.AttributeType == typeof(Microsoft.Owin.OwinStartupAttribute)).FirstOrDefault()?.ConstructorArguments[0].Value as Type;
             //if (startupClass != null)
             //{
-                
+
             //    var configureMethod = startupClass.GetRuntimeMethods().FirstOrDefault(s =>
             //        s.Name == "Configuration" && s.GetParameters().Length == 1 &&
             //        s.GetParameters()[0].ParameterType == typeof(IAppBuilder));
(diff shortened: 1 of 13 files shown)


exyi:
You don't need to call Configure in DotVVM.Compiler, ConfigureServices is enough.

exyi:
You will actually have to make sure that all your services that are needed for dotvvm.compiler will be registered in this method, which is pretty much unsystematic and IMO completely breaks the principles ASP.NET Core DI idea - it's not single registration interface for everything anymore ;(

djanosik:
@exyi Well, DotVVM still supports OWIN and there is no ConfigureServices method in that case.

exyi:
> @exyi Well, DotVVM still supports OWIN and there is no
> ConfigureServices method in that case.

Yes, but I think that it would be better to have recommended naming
convention on OWIN than to cripple Asp.Net Core

quigamdev:
I don't like this change too. The reason is to have some check for compiler but naming convention should be enough.

quigamdev:
We will use naming convention to configure services. In case that method is missing the VS extension will add error/warning with link to documentation.

exyi:
> In case that method is missing the VS extension will add
> error/warning with link to documentation.

I think that you can add that error to DotVVM.Compiler itself

tomasherceg:
In OWIN, we are registering services into `options` in `UseDotVVM`:

```
            var dotvvmConfiguration = app.UseDotVVM<DotvvmStartup>(applicationPhysicalPath, options: options =>
            {
                options.Services.AddSingleton<IViewModelLoader>(serviceProvider => new WindsorViewModelLoader(container));                
            });
```

We should make `Services` in the `options` obsolete in order to make the users move these registrations in `DotvvmStartup`.

PR #572, opened 2018-02-26
Implemented CSS style bindable

### LINKED ISSUE #569: Make inline CSS styles bindable
We can allow this by adding `CssStyles` property group. It will work the same way as `CssClasses` group but it will generate KO [style](http://knockoutjs.com/documentation/style-binding.html) binding.

```html
<div style-width="{value: MyWidth}">
```

Don't forget to test whether style names containing hyphen work properly (eg. `font-weight`).

adamjez:
Added `Style-*` property group that generates the `style` knockout binding. I'm not sure what type should I use for the property group, currently there is `object`.

Usage example:
```Html
<div Style-color="{value: Color}" Style-display="{value: Condition ? 'none' : 'inline-block'}"/>
<div style="background-color: green;" Style-width="{value: Width + 'px'}"/>
```

Resolves #569


src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -194,6 +200,28 @@ private void AddCssClassesToRender(IHtmlWriter writer)
...
+            KnockoutBindingGroup stylePropertiesBindingGroup = null;
+            foreach (var styleProperty in StyleProperties.Properties)
+            {
+                if (stylePropertiesBindingGroup == null) stylePropertiesBindingGroup = new KnockoutBindingGroup();
+                stylePropertiesBindingGroup.Add(styleProperty.GroupMemberName, this, styleProperty);
+
+                try
+                {
+                    var value = GetValue(styleProperty).ToString();
+                    writer.AddStyleAttribute(styleProperty.GroupMemberName, value);
```
exyi: When the value is null **or empty**, the style property is not set by knockout. The null case is actually consistent with behavior of this, but I think that we could do that without the NullReferenceException ;)

Also note that when the value contains invalid stuff (like a malitious input), the assignment is ignored on client side but can do some harm on the server. I'm not sure if we can do anything about it, but it's worth noting.

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -51,6 +51,11 @@ public HtmlGenericControl(string tagName, bool allowImplicitLifecycleRequirement
         public static DotvvmPropertyGroup CssClassesGroupDescriptor =
             DotvvmPropertyGroup.Register<bool, HtmlGenericControl>("Class-", "CssClasses");
 
+        public VirtualPropertyGroupDictionary<object> StyleProperties => new VirtualPropertyGroupDictionary<object>(this, StylePropertiesGroupDescriptor);
```
djanosik: I would rename it to `CssStyles` - just to make it similar to `CssClasses`.

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -194,6 +200,31 @@ private void AddCssClassesToRender(IHtmlWriter writer)
...
+            {
+                if (stylePropertiesBindingGroup == null) stylePropertiesBindingGroup = new KnockoutBindingGroup();
+                stylePropertiesBindingGroup.Add(styleProperty.GroupMemberName, this, styleProperty);
+
+                try
+                {
+                    var value = GetValue(styleProperty)?.ToString();
+                    if (!string.IsNullOrEmpty(value))
+                    {
+                        writer.AddStyleAttribute(styleProperty.GroupMemberName, value.ToString());
```
djanosik: You don't need to call `ToString` here.

adamjez: You're right, I forgot that after the last changes

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
adamjez: ok, let's be consistent...

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -194,6 +200,31 @@ private void AddCssClassesToRender(IHtmlWriter writer)
             if (cssClassBindingGroup != null) writer.AddKnockoutDataBind("css", cssClassBindingGroup);
         }
 
+        private void AddStylePropertiesToRender(IHtmlWriter writer)
+        {
+            KnockoutBindingGroup stylePropertiesBindingGroup = null;
+            foreach (var styleProperty in StyleProperties.Properties)
+            {
+                if (stylePropertiesBindingGroup == null) stylePropertiesBindingGroup = new KnockoutBindingGroup();
+                stylePropertiesBindingGroup.Add(styleProperty.GroupMemberName, this, styleProperty);
```
djanosik: This will add hard-coded values to binding group. I think it's not necessary because hard-coded values are added to `style` attribute and can't be changed.

src/DotVVM.Framework/Controls/HtmlGenericControl.cs
```diff
@@ -194,6 +200,31 @@ private void AddCssClassesToRender(IHtmlWriter writer)
             if (cssClassBindingGroup != null) writer.AddKnockoutDataBind("css", cssClassBindingGroup);
         }
 
+        private void AddStylePropertiesToRender(IHtmlWriter writer)
+        {
+            KnockoutBindingGroup cssStylesBindingGroup = null;
+            foreach (var styleProperty in CssStyles.Properties)
+            {
+                if (cssStylesBindingGroup == null) cssStylesBindingGroup = new KnockoutBindingGroup();
+                cssStylesBindingGroup.Add(styleProperty.GroupMemberName, this, styleProperty);
```
djanosik: This will add hard-coded values to binding group. I think it's not necessary because hard-coded values are added to `style` attribute and can't be changed.

adamjez: Good point. I changed it accordingly. thank you

djanosik: Ok, but it should be good enough to set the `nullBindingAction`. You don't need to call `HasValueBinding`.

djanosik: Or do you want to avoid rendering of empty binding group?

adamjez: I see. But yes, I wanted to avoid rendering `style` data-bind when there are only _hard-coded_ values. I didn't know about `nullBindingAction` and it's somehow strange that you have to allocate delegate to do nothing.

djanosik: Ok. Can you also change `CssClasses` to not render empty group in this case?

adamjez: I did. Is it okey?

adamjez:
thanks for review, I'm merging it.

PR #573, opened 2018-03-03
Debug mode initialization

quigamdev:
Old project template tells to you to configure Debug property after DotvvmConfiguration is created. 
With DotVVM 2.0 you cannot configure Debug property after configuration is created, because some services copies the value.
That is is why Debug property is read-only now and the only way how to set it up from a different assembly is to use EnvironmentDotvvmConfigurationInitializer that logs whether the configuration was  configured or not.

- DotvvmConfiguration.Debug - read-only
- Added debug parameter to aspnet core ApplicationBuilderExtensions


src/DotVVM.Framework.Hosting.AspNetCore/ApplicationBuilderExtensions.cs
```diff
@@ -46,12 +47,14 @@ public static DotvvmConfiguration UseDotVVM<TStartup>(this IApplicationBuilder a
...
         /// </param>
-        public static DotvvmConfiguration UseDotVVM(this IApplicationBuilder app, string applicationRootPath, bool? useErrorPages)
+        /// <param name="debug">A value indicating whether the application should run in debug mode.</param>
+        public static DotvvmConfiguration UseDotVVM(this IApplicationBuilder app, string applicationRootPath, bool? useErrorPages, bool? debug)
         {
             var env = app.ApplicationServices.GetRequiredService<IHostingEnvironment>();
             var config = app.ApplicationServices.GetRequiredService<DotvvmConfiguration>();
 
-            config.Debug = env.IsDevelopment();
+            new DotvvmConfigurationEnvironmentInitializer().Initialize(config, debug ?? env.IsDevelopment());
```
exyi: Why the hell are you creating the instance?

exyi:
> With DotVVM 2.0 you cannot configure Debug property after configuration is created, because some services copies the value.

Which services by the way?

adamjez:
As @exyi pointed out, if we can invoke `DotvvmStartup` before any middleware is created that would be best solution and you can set the `Debug` flag there if you want to override `IHostingEnvironment`  (on AspNetCore).

exyi:
On Sat, 2018-03-03 at 20:32 +0000, Adam Jež wrote:
> Comment should be changed. I would provide hint how to set `Debug
> Mode` properly.

Yes, I agree, but it applies to entire DotvvmConfiguration. We should
make really clear that any modificaitions after the application has
started may be ignored.

quigamdev:
DefaultResourceHashService and LocalResourceUrlManager needed Debug to be set in app.UseDotVVM() method. Now I changed alternateDirectories to be used when DotvvmConfiguration.Debug is set to true.

src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -13,6 +13,7 @@ public static class ServiceCollectionExtensions
         /// Adds DotVVM services with authorization and data protection to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        // ReSharper disable once InconsistentNaming
         public static IServiceCollection AddDotVVM<TServiceConfigurator>(this IServiceCollection services) where TServiceConfigurator : IDotvvmServiceConfigurator, new()
         {
             AddDotVVMServices(services);
```
exyi: Resharper is right in this case, we usualy use names like `AddDotvvmServices`

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
```diff
@@ -29,12 +29,11 @@ public static class DotvvmServiceCollectionExtensions
...
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
-        /// <param name="allowDebugServices">If the vs-diagnostics services should be registered</param>
-        public static IServiceCollection RegisterDotVVMServices(IServiceCollection services, bool allowDebugServices = true)
+        public static IServiceCollection RegisterDotVVMServices(IServiceCollection services)
         {
             services.AddOptions();
 
-            if (allowDebugServices) services.AddDiagnosticServices();
+            services.AddDiagnosticServices();
```
exyi: Do we really want to register these services on production?

src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
```diff
@@ -9,24 +9,41 @@ namespace DotVVM.Framework.Diagnostics
...
+
+        private DiagnosticsServerConfiguration configuration;
+        private DateTime configurationLastWriteTimeUtc;
+
+        public string DiagnosticsServerHostname
         {
-            LoadConfiguration();
+            get
+            {
+                EnsureConfigurationValid();
```
exyi: What is the point of this change? Why not load the configuration in constructor?

src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
```diff
@@ -9,24 +9,41 @@ namespace DotVVM.Framework.Diagnostics
...
+        {
+            get
+            {
+                EnsureConfigurationValid();
+                return configuration?.Port;
+            }
+        }
 
-        private void LoadConfiguration()
+        private void EnsureConfigurationValid()
```
exyi: I think that this should be called `LoadConfiguration` as it really loads it instead of just validating it.

quigamdev: What about `EnsureConfigurationLoaded`?

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
quigamdev: They already was registered on production. `bool allowDebugServices = true` was never false on production. I checked that even when they are registered they are not used when Debug is true.

src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
quigamdev: In case that Visual Studio changes the address while app is running.

src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
quigamdev: The method name is same as long as we added support for dotnet core.  I just added overload. I don't think it is good idea to change the name. What about `UseDotVVM`? That's the same case.

src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
exyi: Ok, it's at least not misleading.

src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
exyi: no we can't touch the public api and it's not against the naming convention if `DotVVM` is not followed by any other text. I was looking at `AddDotVVMServices`, just misplaced the comment, sorry.

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
exyi: That seems like a bug.

Don't tell me that DiagnosticsRenderer is not used when it's registered like that: https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs#L99

src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs
exyi: No, even `EnsureConfigurationLoaded` is totally misleading - it feel to me that it loads the configuration only once while this method realoads every time. Maybe it should be rather called `ReloadConfiguration`

src/DotVVM.Framework.Hosting.AspNetCore/ApplicationBuilderExtensions.cs
```diff
@@ -48,25 +46,32 @@ public static DotvvmConfiguration UseDotVVM<TStartup>(this IApplicationBuilder a
...
+        {
+
             var env = app.ApplicationServices.GetRequiredService<IHostingEnvironment>();
             var config = app.ApplicationServices.GetRequiredService<DotvvmConfiguration>();
-
             config.Debug = env.IsDevelopment();
             config.ApplicationPhysicalPath = applicationRootPath ?? env.ContentRootPath;
+            startup.Configure(config, applicationRootPath);
 
             if (useErrorPages ?? env.IsDevelopment())
```
adamjez: Maybe we should use `config.Debug` as for other dotvvm options

src/DotVVM.Framework/Configuration/DotvvmConfiguration.cs
```diff
@@ -87,9 +89,32 @@ public class DotvvmConfiguration
...
+        {
+            get => debug;
+            set
+            {
+                VerifyIsFrozen();
+                debug = value;
+            }
+        }
+
+        private void VerifyIsFrozen()
```
djanosik: EnsureNotFrozen would a better name here, wouldn't be?

djanosik: Is it intended to call this method only in the `Debug` property?

src/DotVVM.Samples.BasicSamples.AspNetCore/Properties/launchSettings.json
```diff
@@ -12,16 +12,16 @@
       "commandName": "IISExpress",
       "launchBrowser": true,
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
+        //"ASPNETCORE_ENVIRONMENT": "Development"
```
djanosik: Did you really mean to comment out the variable?

exyi:
On Wed, 2018-03-14 at 16:08 +0000, Dusan Janosik wrote:
> EnsureNotFrozen would a better name here, wouldn't be?

I'd rather name it ThrowIfFrozen, the same way as it's called in JsNode
- https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Co
mpilation/Javascript/Ast/JsNode.cs#L16

exyi:
On Wed, 2018-03-14 at 16:11 +0000, Dusan Janosik wrote:
> Is it intended to call this method only in the `Debug` property?

Basically, yes. I also don't like this inconsistency, but the Debug
property is the main source of potential errors and allow freezing for
the rest of the configuration (including the Lists and Dictionaries)
would be a real pain.

src/DotVVM.Samples.BasicSamples.AspNetCore/Properties/launchSettings.json
quigamdev: Nop.

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
quigamdev: Inside of these services is condition that if config.Debug is false then diagnostic sevices are not used.

PR #577, opened 2018-03-09
Fixed textbox formatting alongside Type attribute

adamjez:
Input of type `number` and `date` expects values in specific format. Specifying `FormatString` is not appropriate and it is very likely that it breaks the input.

I'm suggesting throwing exception when user tries to set `FormatString` alongside `date` or `number` input type.

@quigamdev @djanosik @exyi


src/DotVVM.Framework/Controls/TextBox.cs
```diff
@@ -125,11 +125,21 @@ public static bool NeedsFormatting(IValueBinding binding) => binding != null &&
 
         protected internal override void OnPreRender(IDotvvmRequestContext context)
         {
-            isFormattingRequired = !string.IsNullOrEmpty(FormatString) ||
-                #pragma warning disable
+            if (!string.IsNullOrEmpty(FormatString) && (Type == TextBoxType.Number || Type == TextBoxType.Date))
```
exyi: What about `TextBoxType.Time`? It also requires special format. Note that there is a few more allowed `type` value that `TextBoxType` does not support - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input, what about adding support for them?

exyi:
Just note that this is a little breaking change, if you encounter any problem with this it's likely to fall into these categories:
* You have `FormatString="yyyy-MM-dd"`, in that case you can remove the property
* You use some Javascript library that expects different format. In that case shot it's author and workaround it by removing `Type="Date"` and using `html:type="date"`
* Your date input was broken, you are welcome to fix it by removing the `FormatString` property.

src/DotVVM.Framework/Controls/TextBox.cs
adamjez: Yeah, we should add support for any input `types` where it make sense. 
But `TextBoxType.Time` type is little complicated. What C# type should user use? `TimeSpan` (currently not supported) or `DateTime` (what will happend if user bind the same property in `Date` and `Time` input type?)
We should make a **list of input types** where it make sense to threat them specially.

exyi:
> But `TextBoxType.Time` type is little complicated. What C# type
> should user use? `TimeSpan` (currently not supported) or `DateTime`
> (what will happend if user bind the same property in `Date` and
> `Time` input type?)

DateTime should definitely work. <input type="date"> is not very well
suited for TimeSpan, it's supposed to represent time in day, not a
timespan. 

> We should make a **list of input types** where it make sense to
> threat them specially.

I propose this one:

Types not supported by TextBox (handled by different control):
* button
* checkbox
* file
* hidden
* image - like a button, currently not supported, we could add a
feature to Button
* reset - button that clears the form. No point in wrapping this one as
dotvvm control.
* radio
* submit


Type not handled specially (should be bound to string):
* color
* email
* password
* search
* tel (representing telephone number as integer is not great idea)
* text
* url (maybe we could add support for Uri, but it's not necesarry)

Type handled specially (when not bound to string):
* date
* datetime-local (we should add support) - date with time
* month (we should add support) - date without a day field
* week (we should add support) - date in format "YEAR-WEEK". This is a
bit tricky as I'm not aware of any way the .NET format string could get
correct week number.
* number
* range (we should add support) - like a number, with limits and
usually represented as a slider

Also note that some of them support `min` and `max` (number, date, ...)
attributes that need to be formatted in specific ways, so we should
also add support for that.

djanosik:
@exyi Do you think it's ok for `TextBox` to render a slider when `Type` is set to range?

exyi:
On Fri, 2018-03-09 at 17:13 +0000, Dusan Janosik wrote:
> @exyi Do you think it's ok for `TextBox` to render a slider when
> `Type` is set to range?

Better than to fail :) We could add another control for this one, it
seems to be supported in all browsers and work pretty well.

adamjez:
@exyi I agree with your list but I'm not sure about handling `min`, `max`, `step`. Maybe new controls would be fine for these cases.

Currently there is strange behavior when `DateTime` property is binded to both `time` and `date`. Both input contains option to clear the input which will clear **both time and date** part of the property

src/DotVVM.Framework/Controls/TextBox.cs
```diff
@@ -125,11 +125,21 @@ public static bool NeedsFormatting(IValueBinding binding) => binding != null &&
...
+            {
+                throw new NotSupportedException($"Property StringFormat cannot be used with Type set to '{ Type }'." +
```
djanosik: StringFormat => FormatString. And BTW: This is not about localization.

src/DotVVM.Framework/Controls/TextBox.cs
```diff
@@ -125,11 +125,21 @@ public static bool NeedsFormatting(IValueBinding binding) => binding != null &&
...
+                    $" In this case browsers localize '{ Type }' themselves.");
+            }
+
+            // Don't format <input type="Number" ... browsers expect
+            // value in specific format and localization breaks it
+            isFormattingRequired =  Type != TextBoxType.Number &&
```
djanosik: I think you should handle `TextBoxType.Date` here as well.

djanosik:
I think we should merge this PR. Adding support for other types is out of scope of this PR. @adamjez Can you file an issue for that?

src/DotVVM.Framework/Controls/TextBox.cs
adamjez: Thanks.

src/DotVVM.Framework/Controls/TextBox.cs
adamjez: In case of `TextBoxType.Date`  formatting is **required**. But I will try to make the solution more generic.

adamjez:
@djanosik I want to make some minor change in this PR and after the merge I will file the issue.

PR #578, opened 2018-03-11
Added AssertConfigurationIsValid on DotvvmConfiguration

quigamdev:
Added AssertConfigurationIsValid that check whether DotvvmConfiguration all registered routes and control depended on file are correct.


src/DotVVM.Framework/DotVVM.Framework.csproj
```diff
@@ -92,6 +92,11 @@
     <Reference Include="System.ComponentModel.DataAnnotations" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Build.Tasks.v4.0">
+      <HintPath>..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.7.1\Microsoft.Build.Tasks.v4.0.dll</HintPath>
+    </Reference>
```
exyi: What the hell is this?

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
...
+            var invalidControls = new List<DotvvmConfigurationAssertResult<DotvvmControlConfiguration>>();
+            var loader = new AggregateMarkupFileLoader();
+            foreach (var route in config.RouteTable.Where(s => !string.IsNullOrWhiteSpace(s.VirtualPath)))
+            {
+                if (string.IsNullOrWhiteSpace(route.RouteName))
+                {
+                    invalidRoutes.Add(new DotvvmConfigurationAssertResult<RouteBase>(route, DotvvmConfigurationAssertReason.MissingRouteName));
+                }
+
+                var content = loader.GetMarkup(config, route.VirtualPath);
```
exyi: This check is invalid - you can register you own loader that will understand the path and load it correctly.

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
...
+                {
+                    invalidRoutes.Add(new DotvvmConfigurationAssertResult<RouteBase>(route, DotvvmConfigurationAssertReason.MissingFile));
+                }
+            }
+            foreach (var control in config.Markup.Controls.Where(s => !string.IsNullOrWhiteSpace(s.Src)))
+            {
+                if (string.IsNullOrWhiteSpace(control.TagPrefix))
+                {
+                    invalidControls.Add(new DotvvmConfigurationAssertResult<DotvvmControlConfiguration>(control, DotvvmConfigurationAssertReason.MissingControlTagPrefix));
+                    break;
```
exyi: mmm, what about `else`?

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
...
+                }
+
+                if (string.IsNullOrWhiteSpace(control.TagName) && !string.IsNullOrWhiteSpace(control.Src))
+                {
+                    invalidControls.Add(new DotvvmConfigurationAssertResult<DotvvmControlConfiguration>(control, DotvvmConfigurationAssertReason.MissingControlTagName));
+                    break;
+                }
+
+                var content = loader.GetMarkup(config, control.Src);
```
exyi: This check is also invalid, see above

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
                 table.Add(route.RouteName, route);
             }
         }
+
+        public static void AssertConfigurationIsValid(this DotvvmConfiguration config)
```
exyi: And where is it invoked?

src/DotVVM.Framework/DotVVM.Framework.csproj
quigamdev: This should not be there. I didn't added it intentionally. I will remove it.

src/DotVVM.Framework/Routing/RouteHelper.cs
quigamdev: You can invoke this method after registration of dotvvm by UseDotVVM that returns DotvvmConfiguration. Then you can call config.AssertConfigurationIsValid() to verify that you configuration is valid. The usage of this method is same as AutoMapper.AssertConfigurationIsValid(). You can use it if you need it but it is optional.

src/DotVVM.Framework/Routing/RouteHelper.cs
quigamdev: We can provide optional parameter IMarkupFileLoader.

exyi: I don't see any, where is the parameter? You are initializing the `loader` here: https://github.com/riganti/dotvvm/pull/578/files/a4ffcbd72298c4a8d8d71626ff6627f4cc97c280#diff-c9adfdb58f0b70c3f469077f400c8d38R44

And why make things complicated, you can simply get the loader from service collection.

src/DotVVM.Framework/Routing/RouteHelper.cs
exyi: In that case it's fairly useless, I guess.

src/DotVVM.Framework/Routing/RouteHelper.cs
adamjez: @exyi `else`? what about other controls without `TagPrefix`

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
...
+            }
+            foreach (var control in config.Markup.Controls.Where(s => !string.IsNullOrWhiteSpace(s.Src)))
+            {
+                if (string.IsNullOrWhiteSpace(control.TagPrefix))
+                {
+                    invalidControls.Add(new DotvvmConfigurationAssertResult<DotvvmControlConfiguration>(control, DotvvmConfigurationAssertReason.MissingControlTagPrefix));
+                    break;
+                }
+
+                if (string.IsNullOrWhiteSpace(control.TagName) && !string.IsNullOrWhiteSpace(control.Src))
```
adamjez: Is there the `!string.IsNullOrWhiteSpace(control.Src)` check really needed? It is filtered with `Where` in `foreach` statement

src/DotVVM.Framework/Routing/RouteHelper.cs
```diff
@@ -35,5 +36,54 @@ public static void AutoDiscoverRoutes(this DotvvmRouteTable table, IRoutingStrat
...
+                    invalidControls.Add(new DotvvmConfigurationAssertResult<DotvvmControlConfiguration>(control, DotvvmConfigurationAssertReason.MissingFile));
+                }
+            }
+
+            if (invalidControls.Any() || invalidRoutes.Any())
+            {
+                throw new DotvvmConfigurationException(invalidRoutes, invalidControls);
+            }
+
+
```
adamjez: Remove empty lines :)

src/DotVVM.Framework/Routing/RouteHelper.cs
exyi: Sorry, I thought this was a `continue` statement... Anyway, to be explicit - the `break` statement jumps after the `foreach` statement which means that only one error is added to the collection `invalidControls`. Why is it a collection then? I'd change it to a series of `if ... else` statements and add exactly one or zero errors **per registered control**

src/DotVVM.Framework/Routing/RouteHelper.cs
quigamdev: The Where condition should not be there. Because empty tagname and existing src means that you missed to set tagname.

src/DotVVM.Framework/Routing/RouteHelper.cs
quigamdev: @adamjez  The purpose of this validation method is to let developer know that something in configuration is incorrectly registered. As developer, you want to know whether all routes in your app or src of markup controls are correct. This behavior is same as AutoMapper do this. So I think that **502.5 Process Failure** is correct behavior.

Of course when you want to use this somewhere else you can create a postback and there call `Context.Configuration.AssertConfigurationIsValid()`. Then it throws exception in postback and you can catch it and provide a more info at output.

src/DotVVM.Framework/Configuration/DotvvmConfigurationException.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        private static string BuildMessage(List<DotvvmConfigurationAssertResult<RouteBase>> routes, List<DotvvmConfigurationAssertResult<DotvvmControlConfiguration>> controls)
+        {
+
+            var sb = new StringBuilder();
+            BuildRoutesMessage(routes, sb);
+            BuildControlsMessage(controls, sb);
+            return sb.ToString();
+
+
```
djanosik: Why so many empty lines?

src/DotVVM.Framework/Routing/RouteHelper.cs
djanosik: I don't think we need an optional parameter. The loader is retrieved from `ServiceProvider` - it's good enough. But I agree, this check is invalid.

djanosik: Or maybe not. The loader does not work with the route itself. And the `VirtualPath` is not validated here.

exyi: Originally the loader was not retrieved from the `ServiceProvider`, my comment was about that. But you are right, that the virtual path is actually interpreted by the `IDotvvmPresenter`, so the check is invalid anyway...

src/DotVVM.Framework/Routing/RouteHelper.cs
djanosik: Hmm, this is something you don't want to call multiple times. I believe it should not be used in production at all. Or are there any reasons to do so?

src/DotVVM.Samples.Common/DotvvmStartup.cs
```diff
@@ -113,7 +113,6 @@ private static void AddControls(DotvvmConfiguration config)
             config.Markup.AddCodeControls("cc", typeof(Controls.PromptButton));
             config.Markup.AddMarkupControl("IdGeneration", "Control", "Views/FeatureSamples/IdGeneration/IdGeneration_control.dotcontrol");
             config.Markup.AddMarkupControl("FileUploadInRepeater", "FileUploadWrapper", "Views/ComplexSamples/FileUploadInRepeater/FileUploadWrapper.dotcontrol");
-            config.Markup.AddMarkupControl("sample", "PasswordStrengthControl", "Views/FeatureSamples/ClientExtenders/PasswordStrengthControl.dotcontrol");
```
exyi: You are right that this route is wrong, but it would be nicer to replace it with the correct one instead of removing it :) The correct file is at `Views/FeatureSamples/ClientExtenders/PasswordStrength.dothtml`

quigamdev: This route is no longer needed. The `Views/FeatureSamples/ClientExtenders/PasswordStrength.dothtml` is automatically resolved by route strategy.

PR #580, opened 2018-03-12
Improve README

Sejsel:
This is what I would personally like to change in the README. The part this changes was written by @exyi, so it is his call if he wants to keep something the way he wrote it. Feel free to take just some of the changes.

### The ";)"
I am not really fond of having these in such a prominent text, I wouldn't put them in a documentation and I believe the README is a part of a documentation. They don't feel very professional to me. 
### Start of the second paragraph
The static command `staticCommand`s looked weird to me, especially since it was a link (including the s).
### Other changes
These are minor, you might want to use a diff that can compare changes on single lines. They generally just "felt better" to me.

## Potential other changes (not done)
- Perhaps split the lines so the diffs are more readable

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ How to Start
 
 How Does it Work
 ----------------
-DotVVM is no magic, so let's have a look at how it works. Or at least how would a simplified core work ;)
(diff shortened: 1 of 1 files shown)


exyi:
Thanks, it also generally feels better to me.

PR #581, opened 2018-03-13
Refactored DataPager - added extensibility points

MichalTichy:
Added ways how to modify individual parts (buttons) of datapager

### DIFF
--- a/src/DotVVM.Framework/Controls/DataPager.cs
+++ b/src/DotVVM.Framework/Controls/DataPager.cs
@@ -138,12 +138,12 @@ namespace DotVVM.Framework.Controls
         public static readonly DotvvmProperty EnabledProperty =
             DotvvmPropertyWithFallback.Register<bool, DataPager>(nameof(Enabled), FormControls.EnabledProperty);
 
-        private HtmlGenericControl content;
-        private HtmlGenericControl firstLi;
-        private HtmlGenericControl previousLi;
-        private PlaceHolder numbersPlaceHolder;
-        private HtmlGenericControl nextLi;
-        private HtmlGenericControl lastLi;
+        protected HtmlGenericControl content;
+        protected HtmlGenericControl firstLi;
+        protected HtmlGenericControl previousLi;
+        protected PlaceHolder numbersPlaceHolder;
+        protected HtmlGenericControl nextLi;
+        protected HtmlGenericControl lastLi;
 
         protected internal override void OnLoad(IDotvvmRequestContext context)
         {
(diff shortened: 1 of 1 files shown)


src/DotVVM.Framework/Controls/DataPager.cs
```diff
@@ -138,12 +138,12 @@ public bool Enabled
...
-        private HtmlGenericControl previousLi;
-        private PlaceHolder numbersPlaceHolder;
-        private HtmlGenericControl nextLi;
-        private HtmlGenericControl lastLi;
+        protected HtmlGenericControl content;
+        protected HtmlGenericControl firstLi;
+        protected HtmlGenericControl previousLi;
+        protected PlaceHolder numbersPlaceHolder;
+        protected HtmlGenericControl nextLi;
+        protected HtmlGenericControl lastLi;
```
djanosik: I think I don't like this. Can we instead have protected virtual methods like `CreateFirstLi()`?

adamjez: I'm not sure if methods would be proper here. Maybe protected property with more generic name would be sufficient?

exyi:
DataPager is a complete mess and I'd really like to completely redesign it's implementation, so I don't really like to introduce new public API here. Why do you need that? Why do you need to inherit from DataPager?

By the way, I'd not consider marking methods and fields as protected or virtual as refactoring :P

src/DotVVM.Framework/Controls/DataPager.cs
quigamdev: At least properties, but not fields.

exyi: @quigamdev Is there any difference between properties and fields? I know that you could potentially reimplement the property setter and getter to maintain binary compatibility, but if we'd remove these fields it would not make any sense to emulate them somehow, as these controls would not exist anymore and thus manipulating them whould be useless.

MichalTichy:
The reason for this changes is that we need to modify buttons in datapager in order to update bs:DataPager to Bootstrap 4

Currently only option how to do this is to override AddItemCssClass which does not allow us to change inner elements.

So I added few methods
CreatePageNumberButton
CreateNavigationButton
CreateWrapperList

src/DotVVM.Framework/Controls/DataPager.cs
```diff
@@ -138,12 +138,12 @@ public bool Enabled
...
-        private HtmlGenericControl previousLi;
-        private PlaceHolder numbersPlaceHolder;
-        private HtmlGenericControl nextLi;
-        private HtmlGenericControl lastLi;
+        protected HtmlGenericControl ContentWrapper { get; set; }
+        protected HtmlGenericControl GoToFirstPageButton { get; set; }
+        protected HtmlGenericControl GoToPreviousPageButton { get; set; }
+        protected PlaceHolder NumberButtonsPlaceHolder { get; set; }
+        protected HtmlGenericControl GoToNextPageButton { get; set; }
+        protected HtmlGenericControl GoToLastPageButton { get; set; }
```
djanosik: Do they need to have protected setter?

PR #584, opened 2018-03-20
Added support for title attribute binding in option element 

### LINKED ISSUE #583: Add support for title attribute in ComboBox
Currently, there is no way to render `title` attribute in option with `ComboBox`:
```html
<select>
  <option value="1" title="Desired Title">Text 1</option>
  <option value="2" title="Desired Title">Text 2</option>
</select>
```
We are using knockout.js [**options** binding](http://knockoutjs.com/documentation/options-binding.html) which lacks support of this scenario. 

Any idea how could we resolve it? @exyi @djanosik

adamjez:
Change needed in https://github.com/riganti/knockout/commit/8212a7a493eaf0d5c46d9ece3d9c8728a1aa352a
Resolves #583


src/DotVVM.Samples.Common/Utilities/JavaScriptUtils.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DotVVM.Framework.Utils;
+
+namespace DotVVM.Samples.Common.Utilities
+{
+    public static class JavaScriptUtils
+    {
+        public static string LimitLength(string text, int length) => StringUtils.LimitLength(text, length);
```
exyi: Why not just use the StringUtils.LimitLength in the value binding?

exyi: And what about to add support for this function to framework?

src/DotVVM.Samples.Common/CommonConfiguration.cs
```diff
@@ -16,6 +20,12 @@ public static void ConfigureServices(IServiceCollection services)
             });
 
             services.Configure<DotvvmResourceRepository>(RegisterResources);
+
+            services.Configure<JavascriptTranslatorConfiguration>(c => {
+                c.MethodCollection.AddMethodTranslator(typeof(JavaScriptUtils),
+                   nameof(JavaScriptUtils.LimitLength),
+                   new GenericMethodCompiler((a) => new JsIdentifierExpression("limitLength").Invoke(a)));
```
exyi: If we'd like to add this function to framework, it does not seem like a good idea to use the limitLength as a global. And even here, it scares me a bit :P

adamjez: I created it only for testing purpose obviously. I can put it in a js object if you want to.

src/DotVVM.Samples.Common/Utilities/JavaScriptUtils.cs
adamjez: > Why not just use the StringUtils.LimitLength in the value binding?

I wanted to have a proxy class for these javascript utilities functions

> And what about to add support for this function to framework?

I'm not sure about adding support in the framework, instead I would like to make the **javascript method registration** a lot easier.

src/DotVVM.Framework/Controls/SelectHtmlControlHelpers.cs
```diff
@@ -42,6 +42,11 @@ public static void RenderOptionsProperties(IHtmlWriter writer, SelectorBase sele
                 writer.AddKnockoutDataBind("optionsValue", "function (i) { return ko.unwrap(i)[" + KnockoutHelper.MakeStringLiteral(selector.ValueMember) + "]; }");
             }
             #pragma warning restore
+
+            if (selector.ItemTitleBinding != null)
+            {
+                writer.AddKnockoutDataBind("optionsTitle", selector.ItemTitleBinding.GetProperty<SelectorItemBindingProperty>().Expression, selector);
```
djanosik: @exyi What is the `SelectorItemBindingProperty` good for?

exyi: @djanosik https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Compilation/Binding/GeneralBindingPropertyResolvers.cs#L328 - it's a lambda of the binding.

PR #586, opened 2018-03-23
Interface IPresenterActionFilter

quigamdev:
- added interface IPresenterActionFilter that contains OnPresenterExecuting, OnPresenterExecuted, OnPresenterException
- IPresenterActionFilter implemented on ActionFilterAttribute
- DotvvmRoutingMiddleware.Handle - changed resolving of attributes.

Reason: 
When action filter is registered as global filter then IPageActionFilter.OnPageLoadingAsync and IPageActionFilter.OnPageLoadedAsync are invoked two times.


exyi:
The problem is, that if someone would have implemented a custom Authorize attribute (or something simmilar) it will just stop working with this patch, without any error and only in strange situations (so it probably wont be spotted for a pretty long time by tests nor users). And with possible security implications.

> Please make sure it won't break anything.
When you stop calling a method it will definitely be a breaking change.

I think that just removing the invocation from DotvvmPresenter would have much lower impact - I can't imagine a scenario when someone would depend on the double invocation. But technically the breaking change is the same. So I think that leaving it "as it is" is the most reasonable option in this case.

Other possibility would be to rename/remove the methods just to cause compile-time breaking change (which would annoy a bunch of people as a false alarm)

tomasherceg:
I agree with @exyi about the renaming/removing the methods. 
`OnPageLoadingAsync` and `OnPageLoadedAsync` are not good names - we should make these methods obsolete and introduce new ones - maybe `OnPageInitializedAsync` and `OnPageRenderedAsync`. And describe the change in the description of Obsolete attribute, something like this:

"The behavior of the OnPageLoadingAsync method was changed. Use `OnPresenterExecutingAsync` to add custom logic before any DotVVM page or custom presenter is invoked, or `OnPageInitializedAsync` to add custom logic in DotVVM pages after the viewmodel is created."

"The behavior of the OnPageLoadedAsync method was changed. Use `OnPresenterExecutedAsync` to add custom logic after any DotVVM page or custom presenter is invoked, or `OnPageRenderedAsync` to add custom logic in DotVVM pages after the response is rendered."

Also, we should look at `DotvvmPresenter` and review the events we are calling on the filters. Maybe some of them are duplicate and some might be missing.

tomasherceg:
Any comments on my last suggestion in this thread?

exyi:
I still think that warning is totally not enough for change that could have pretty much undiscoverable security implications.

adamjez:
I've renamed `OnPageLoadingAsync` and `OnPageLoadedAsync` methods to both pick more accurate name and  trigger compilation error (This breaking change must be added into **documentation** and **How to Migrate** document)

PR #587, opened 2018-03-23
Wrapped observables are writeable when possible

djanosik:
Binding expressions like `NestedModel.Property` are read-only when wrapped in computed observables. It makes it impossible to use nested view models with BP controls. I've changed `wrapKnockoutExpression` and `EnsureObservableWrapped` to make them writeable.

It also fixes issues caused when `NestedModel` is initially set to `null`.


src/DotVVM.Framework/Compilation/Javascript/Ast/JsAstHelpers.cs
```diff
@@ -79,17 +79,37 @@ public static IEnumerable<JsExpression> GetLeafResultNodes(this JsExpression exp
...
-            !expression.DescendantNodes().Any(n => (n.HasAnnotation<ResultIsObservableAnnotation>() && !n.HasAnnotation<ShouldBeObservableAnnotation>()) || n.HasAnnotation<ObservableUnwrapInvocationAnnotation>()) ?
-
-                expression.WithAnnotation(ShouldBeObservableAnnotation.Instance) :
+            if (!expression.DescendantNodes().Any(n => (n.HasAnnotation<ResultIsObservableAnnotation>() && !n.HasAnnotation<ShouldBeObservableAnnotation>()) || n.HasAnnotation<ObservableUnwrapInvocationAnnotation>()))
+            {
+                return expression.WithAnnotation(ShouldBeObservableAnnotation.Instance);
+            }
+            else
+            {
+                var arguments = new List<JsExpression>(3);
```
exyi: The `3` looks a bit funny here :P

src/DotVVM.Framework/Compilation/Javascript/Ast/JsAstHelpers.cs
```diff
@@ -79,17 +79,37 @@ public static IEnumerable<JsExpression> GetLeafResultNodes(this JsExpression exp
...
+                    if (expression.SatisfyResultCondition(n => n.HasAnnotation<ResultIsObservableArrayAnnotation>()))
+                    {
+                        arguments.Add(new JsLiteral(true));
+                    }
+                }
+
+                return new JsIdentifierExpression("dotvvm").Member("evaluator").Member("wrapKnockoutExpression").Invoke(arguments)
                     .WithAnnotation(ResultIsObservableAnnotation.Instance)
                     .WithAnnotation(ShouldBeObservableAnnotation.Instance);
+            }
```
exyi: I just think that it would be nicer to produce raw `ko.pureObservable` in case the result is not observable. In that case you'd have to remove the `ShouldBeObservableAnnotation`.

djanosik: Yes. I think it's a good idea.

PR #588, opened 2018-03-29
Added support for SPA navigation using History API

### LINKED ISSUE #449: Use History API to do SPA instead of hashbang

adamjez:
### Added support for SPA navigation using [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API)
The current SPA navigation using *Anchor* has **not been changed**, so users which expect urls which contains `hashbang + url` can still use the old navigation style. 

To choose between navigation styles there is the `UseHistoryApiSpaNavigation` property in the `DotvvmConfiguration`. Default value is `true`.

#### Advantages of history API:
 * Navigation between pages works even without JavaScript (correct url is rendered into `href` attribute)
 * User see prettier urls: `PageB` instead of `/PageB#!/PageB` 
 * Navigating to SPA url will trigger one http request instead of two
 * User can refresh currently loaded page clicking at a link to current page

#### Notes:
  * Property `PrefixRouteName` doesn't make sense with history API navigation
  * `UseHistoryApiSpaNavigation` configuration is being sent to client
  * Old urls will work correctly. Navigating to `/PageB#!/PageB` will work. Url in a address bar will be replaced with `PageB`
  * Testing old and new mechanism for navigation with same web project is pain. I added ActionFilter to change `UseHistoryApiSpaNavigation` during http request, but obviously it's pretty bad solution. If UI tests run sequentially It can work.
  * To make the `DefaultRouteName` property work (in the `SpaContentPlaceHolder` control) I added `data-dotvvm-spacontentplaceholder-content` attribute. This attribute is set when the `SpaContentPlaceHolder` contains some elements otherwise navigation to `DefaultRouteName` will be triggered. @exyi tell me please if you find out a better solution?

Resolves #449

I would like to hear some feedback.


src/DotVVM.Framework/Controls/RouteLink.cs
```diff
@@ -119,5 +119,17 @@ protected override void RenderContents(IHtmlWriter writer, IDotvvmRequestContext
...
+        protected virtual void WriteOnClickAttribute(IHtmlWriter writer, IDotvvmRequestContext context)
+        {
+            if ((bool)GetValue(Internal.IsSpaPageProperty) && context.Configuration.UseHistoryApiSpaNavigation)
+            {
+                writer.AddAttribute("onclick", "return !this.hasAttribute('disabled') && dotvvm.handleSpaNavigation(this.getAttribute('href'));");
+            }
+            else
+            {
+                writer.AddAttribute("onclick", "return !this.hasAttribute('disabled');");
+            }
```
exyi: I think that it should be enough to rely on the `popstate` event.

exyi:
I'm not sure if the `DefaultRouteName` property makes any sense - as far as I understand it, it's just a hack to keep the urls with hashbang shorter.

src/DotVVM.Framework/Controls/RouteLink.cs
adamjez: It is necessary to distinguish links inside and outside of the SPA

adamjez:
I added another `UseHistoryApi` property to `SpaContentPlaceHolder`. This way user can override global settings in `DotvvmConfiguration` and we can easily create UI tests for both SPA navigation methods.

@exyi Should it throw exception when `UseHistoryApi` and `PrefixRouteName` or `DefaultRouteName` are set? or it should do nothing so current behavior would be ok?

src/DotVVM.Framework/Controls/RouteLink.cs
exyi: 1. If you are handling popstate event correctly, it should invoke the SPA navigation anyway.
2. You are rendering this handler for any route, regardless of it's target page.

exyi:
Btw, In the UI tests you could bind the `UseHistoryApi` to a query string parameter, so we could test both modes with just one test ;)

Although throwing an exception would be apropriate in that case (as it's "invalid" combination), I think that just ignoring it will be nicer to our users. The properties do not provide any critical features, so it should be safe to ignore.

src/DotVVM.Framework/Controls/RouteLink.cs
adamjez: >If you are handling popstate event correctly, it should invoke the SPA navigation anyway.

As much as I know, `popstate` event is not invoked on navigation via link

>You are rendering this handler for any route, regardless of it's target page.

You're right

adamjez:
I would love to use these UI tests for both navigation modes but their behavior differs. For example old mode does redirect to `PageA#!/PageB` when you navigate to `PageA` and when `DefaultRouteName` is `PageB`. These UI tests count with this behavior. 

Another UI test check `PrefixRouteName` behavior.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -233,9 +235,20 @@ class DotVVM {
...
+            if (this.useHistoryApiSpaNavigation) {
+                hashChangeHandler = (initialLoad: boolean) => this.handleHashChangeWithHistory(viewModelName, spaPlaceHolder, initialLoad);
+            }
+
+            var spaChangedHandler = () => hashChangeHandler(false)
+            this.domUtils.attachEvent(window, "hashchange", spaChangedHandler);
+            hashChangeHandler(true);         
         }
+
+        window.addEventListener('popstate', (event) => this.handlePopState(viewModelName, event, spaPlaceHolder != null));
```
djanosik: Do we need to handle popstate when history API navigation is disabled?

adamjez: It is required when you navigate Back from normal page to SPA page, you need that navigation handle yourself.

adamjez:
It looks like everyone who wanted to review this PR already did. I'm merging it.

PR #589, opened 2018-04-04
Added double quotes around file name in Content-Disposition header.

kowgli:
This fixes an issue with file names containing spaces in Firefox.
For example if file name is something like "aaa bbb ccc.txt", Firefox (without the double quotes) would try saving it as just "aaa".

### DIFF
--- a/src/DotVVM.Framework/Hosting/Middlewares/DotvvmReturnedFileMiddleware.cs
+++ b/src/DotVVM.Framework/Hosting/Middlewares/DotvvmReturnedFileMiddleware.cs
@@ -32,7 +32,7 @@ namespace DotVVM.Framework.Hosting.Middlewares
             var id = Guid.Parse(context.Request.Query["id"]);
             using (var stream = returnedFileStorage.GetFile(id, out metadata))
             {
-                context.Response.Headers["Content-Disposition"] = "attachment; filename=" + metadata.FileName;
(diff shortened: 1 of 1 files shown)


kowgli:
Would have used string interpolation instead of '+' concatenation, but didn't want to break the existing convention.

djanosik:
I think it may cause issues in other browsers. Wouldn't it be better to encode the file name?

kowgli:
It doesn't cause issues. Tested in Chrome (latest stable), IE9 and Edge (latest stable).
Encoding the filename doesn't work because it doesn't get "unencoded", resulting in + or %20 in the file name instead of spaces.

adamjez:
Encoding filename in `Content-Disposition` is [a little bit complicated](https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http).
We would have to handle it differently for different browsers, but we should try it.

exyi:
I'm afraid that we will have to whitelist allowed characters (and throw exceptions) to prevent possible security issues. I'm not sure how much harm can be done by injecting malitions stuff into this HTTP header...

djanosik:
It might be worth to check [ContentDispositionHeaderValue](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs) implementation. But I would merge this PR to fix the issue in Firefox.

PR #591, opened 2018-04-10
Added ContentDisposition header encoding using *.Net.Http packages

adamjez:
Added encoding of filename for Content-Disposition header.

Encoding is handled in `ContentDispositionHeaderValue` type which is in `System.Net.Http` assembly for .net framework or `Microsoft.Net.Http.Headers` package for .net standard. 
It is okay to add these dependencies? @exyi 

Relates to #589


adamjez:
Now I've tested it in Internet Explorer, Edge and it fails. Result file name is eg. : `=_utf-8_B_MyDFmcWhxJvFmSDEm8WhxZkgxJsucG5n_=`

In Chrome it works as expected. 

I will have a look at it.

djanosik:
`Microsoft.Net.Http` package is outdated. Is it possible to use `System.Net.Http` package?

djanosik:
I see, it's just a naming issue. `Microsoft.Net.Http.Headers` comes from ASP.NET Core. It's ok to use it.

PR #596, opened 2018-05-03
Fixes RequiredResources not working in Repeater when DataSource is empty

### LINKED ISSUE #595: Can't add StylesheetResource to Head, it is already rendered
Some of our controls (eg. `Repeater`) are adding controls to `Children` in the `Render` method. When any of the added controls require a new resource it will fail with an exception like *"Can't add StylesheetResource to Head, it is already rendered"*.

We need to always build the control tree before the control is rendered (eg. in `OnLoad` method).

djanosik:
Fixes #595.

### DIFF
--- a/src/DotVVM.Framework/Controls/Repeater.cs
+++ b/src/DotVVM.Framework/Controls/Repeater.cs
@@ -13,6 +13,7 @@ namespace DotVVM.Framework.Controls
     public class Repeater : ItemsControl
     {
         private EmptyData emptyDataContainer;
+        private DotvvmControl clientSeparator;
 
         public Repeater(bool allowImplicitLifecycleRequirements = true)
         {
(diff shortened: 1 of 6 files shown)


src/DotVVM.Framework/Controls/Repeater.cs
```diff
@@ -97,7 +98,11 @@ public string WrapperTagName
         /// </summary>
         protected internal override void OnLoad(IDotvvmRequestContext context)
         {
-            SetChildren(context);
+            if (context.IsPostBack)
+            {
+                SetChildren(context, true);
```
djanosik: @exyi Do we need to populate `Children` in `OnLoad` when the request is not a postback?

exyi: Yes we need, we need to build the tree in OnPreRender, it's required for `Postback.Update` to work and we need to render it in OnLoad, it's needed for invocation of a command in the children.

djanosik: Exactly. But it is enabled for postbacks. It can't break command invocation, can it?.

djanosik: @tomasherceg What do you think about this?

PR #610, opened 2018-06-05
Return 400 Bad request when csrf token is invalid

adamjez:
**4xx** status code should be returned when CSRF token is invalid. 
Currently we are returning 500 internal error.

I chose status code 400 as ASP.NET team https://github.com/aspnet/Antiforgery


adamjez:
@exyi I've created separate issue for the html page. I'm merging it.

PR #613, opened 2018-06-06
use DisplayAttribute for PropertyName in ViewModelValidationRuleTranslator

ppech:
This pull request adds ability to use DisplayAttribute to localize property name in client side validation.

`[Display(Name = nameof(Resources.Fields.City), ResourceType = typeof(Resources.Fields))]
        [Required(ErrorMessageResourceName = nameof(Resources.ValidationMessages.Required), ErrorMessageResourceType = typeof(Resources.ValidationMessages))]
        public string City { get; set; }`

### DIFF
--- a/src/DotVVM.Framework/ViewModel/Validation/ViewModelPropertyValidationRule.cs
+++ b/src/DotVVM.Framework/ViewModel/Validation/ViewModelPropertyValidationRule.cs
@@ -19,13 +19,19 @@ namespace DotVVM.Framework.ViewModel.Validation
         public ValidationAttribute SourceValidationAttribute { get; set; }
 
         [JsonIgnore]
-        public string PropertyName { get; set; }
+        public string PropertyName => PropertyNameResolver?.Invoke() ?? StaticPropertyName;
 
-        public ViewModelPropertyValidationRule(ValidationAttribute sourceValidationAttribute, string propertyName,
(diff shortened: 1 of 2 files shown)


adamjez:
Thanks for the PR. 

There is a problem that these `ViewModelPropertyValidationRule` are created once and then cached inside `ViewModelPropertyMap`. So the property name is *baked*  and it doesn't matter what culture is requested.

You can look at `ErrorMessage` property inside `ViewModelPropertyValidationRule`. The `ErrorMessage` is resolved with each request. Something similar is necessary with `PropertyName`.

adamjez:
If you want some help let me know.

adamjez:
In the ` ViewModelPropertyValidationRule` class the `PropertyName` property has to be evaluated each request so in the class there should probably be `Func<string> PropertyNameResolver` property/field that will resolve localized name (eg. `() => displayAttribute.GetName()`).


The `PropertyName` can have only getter: `PropertyName => PropertyNameResolver?.Invoke() ?? staticPropertyName;`. If the `DisplayAttribute` exists the `PropertyNameResolver` property would be called otherwise static value inside `staticPropertyName` is used.

ppech:
Second version based on your idea is commited.

adamjez:
Looks good. Thanks!

Issue #528, opened 2017-12-29
Always add defer attribute to script resources


symbiogenesis:
As suggested by exyi in this [comment](https://github.com/riganti/dotvvm/pull/464#issuecomment-354357901)

This may offer some performance wins at almost no cost, and at minimum may improve Google PageSpeed rankings for DotVVM sites.

This should be the default behavior, but we could also offer a way to toggle this off.

exyi:
I totally agree, maybe we could also try using `<link rel="preload" ...` for faster resource loads.

exyi:
This is basically resolved by #614

PR #614, opened 2018-06-11
Added ability to preload link resource

adamjez:
Added ability to preload link resource
 * I added preloading only to scripts, because stylesheets are loaded immediately before preload links
 * I'm not convinced that I have chosen the right approach to add preload links

#528

### DIFF
--- a/src/DotVVM.Framework/Controls/Infrastructure/HeadResourceLinks.cs
+++ b/src/DotVVM.Framework/Controls/Infrastructure/HeadResourceLinks.cs
@@ -24,11 +24,12 @@ namespace DotVVM.Framework.Controls
             foreach (var resource in resourceManager.GetNamedResourcesInOrder())
             {
                 if (resource.Resource.RenderPosition == ResourceRenderPosition.Head)
-                    resource.RenderResourceCached(writer, context);
-
-                if (resource.Resource is IPreloadResource preloadResource)
                 {
-                    preloadResource.RenderPreload(resource.Name, writer, context);
+                    resource.RenderResourceCached(writer, context);
+                }
+                else if (resource.Resource is IPreloadResource preloadResource)
+                {
+                    preloadResource.RenderPreloadLink(writer, context, resource.Name);
                 }
             }
         }
--- a/src/DotVVM.Framework/Controls/Infrastructure/HeadResourceLinks.cs
+++ b/src/DotVVM.Framework/Controls/Infrastructure/HeadResourceLinks.cs
@@ -15,14 +15,21 @@ namespace DotVVM.Framework.Controls
     {
         protected override void RenderControl(IHtmlWriter writer, IDotvvmRequestContext context)
         {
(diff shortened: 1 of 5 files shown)


src/DotVVM.Framework/ResourceManagement/IPreloadResource.cs
```diff
@@ -0,0 +1,13 @@
+﻿using DotVVM.Framework.Hosting;
+
+namespace DotVVM.Framework.ResourceManagement
+{
+    /// <summary>
+    /// Represents a resource thas is preloaded.
```
exyi: Maybe a more descriptive documentation would be nice, there is a few things to note:

* prerendered means a `<link rel="preload" ...>` is rendered into header.
* prerendered means that it is also rendered in the specified destination
* it does not have to be prerendered (for example when the resource is added after head is rendered)

src/DotVVM.Framework/Controls/Infrastructure/HeadResourceLinks.cs
```diff
@@ -15,15 +15,22 @@ public class HeadResourceLinks : DotvvmControl
...
+            // set the flag before the resources are rendered, so they can't add more resources to the list during the render
+            resourceManager.HeadRendered = true;
+
+            // render resource links and preloads
             foreach (var resource in resourceManager.GetNamedResourcesInOrder())
             {
                 if (resource.Resource.RenderPosition == ResourceRenderPosition.Head)
                     resource.RenderResourceCached(writer, context);
+
+                if (resource.Resource is IPreloadResource preloadResource)
```
exyi: I think there should be a `else if`, we don't want to "prefetch" resources that are rendered above in the header.

src/DotVVM.Framework/ResourceManagement/ResourcesRenderer.cs
```diff
@@ -18,12 +18,22 @@ public static void RenderResourceCached(this NamedResource resource, IHtmlWriter
...
         }
 
-        public static string GetRenderedTextCached(this NamedResource resource, IDotvvmRequestContext context) => 
+        public static string GetRenderedTextCached(this NamedResource resource, IDotvvmRequestContext context) =>
             // dont use cache when debug, so the resource can be refreshed when file is changed
             context.Configuration.Debug ?
             RenderToString(resource, context) :
             renderedCache.GetValue(resource.Resource, _ => RenderToString(resource, context));
 
+        public static void RenderPreload(this IPreloadResource resource, string resourceName, IHtmlWriter writer, IDotvvmRequestContext context)
```
exyi: I'd rather let the implementation of IPreloadResource decide what does it want to render into the header. IMO, it will be more powerful and more systematic (almost same signature as Render).

adamjez:
Thanks for the feedback. What do you think now?

exyi:
Nice, thanks, I'm happy with that :)

PR #619, opened 2018-06-22
Added JsExpression PostBackHandler option type

adamjez:
* Added `JsExpression` `PostBackHandler` option type
* Added `JsExpressionInString` to easily insert client postback expression into `PostBackHandler` options.


src/DotVVM.Framework/Controls/KnockoutHelper.cs
```diff
@@ -215,23 +215,39 @@ private static string GetPostBackHandlersScript(DotvvmBindableObject control, st
...
+            switch (optionValue)
+            {
+                case IValueBinding binding:
+                    return new JsIdentifierExpression(
+                        JavascriptTranslator.FormatKnockoutScript(binding.GetParametrizedKnockoutExpression(handler, unwrapped: true),
+                            new ParametrizedCode("c"), new ParametrizedCode("d")));
+                case IStaticValueBinding staticValueBinding:
+                    return new JsLiteral(staticValueBinding.Evaluate(handler));
+                case JsExpression expression:
+                    return expression;
```
exyi: We should definitelly `Clone` the expression - the tree is linked two-way, so one node can't be added multiple times into it.

src/DotVVM.Framework/Controls/KnockoutHelper.cs
```diff
@@ -215,23 +215,39 @@ private static string GetPostBackHandlersScript(DotvvmBindableObject control, st
...
+                default:
```
exyi: To prevent misconceptions with using command binding in the postback handers, it would be nice to throw an exception that general `IBinding` is not supported.

src/DotVVM.Framework/Compilation/Javascript/Ast/JsExpressionInString.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace DotVVM.Framework.Compilation.Javascript.Ast
+{
+    public class JsExpressionInString : JsExpression
+    {
+        public string Expression { get; set; }
+
+        public JsExpressionInString(string expressions)
```
exyi: It would be nicer to take ParametrizedString here.

I also think that we should handle operator precedence here.

src/DotVVM.Framework/Controls/KnockoutHelper.cs
adamjez: I didn't know that.. thank you

src/DotVVM.Framework/Controls/KnockoutHelper.cs
adamjez: okay, I'll add the exception

exyi:
I have removed the `JsExpressionInString`, it's basically a subset of JsSymbolicParameter, I have also added helper method to create it with ad-hoc parameter and default value. 


Although it may seem a bit unintuitive to use SymbolicParameter as a `ParametrizedCode` placeholder, it's behavior in the tree is actually very similar, so it may reduce the complexity of processing the tree. I also quite like the possibility of replacing the hard-coded template.

PR #620, opened 2018-06-22
Fixed empty url registration for Routes in RouteTableGroups

Rumbeer:
Fixed Route registration in RouteTableGroups where empty Route url could not be registered like this: 

```
config.RouteTable.AddGroup("Dashboard", "dashboard", "Views/Dashboard/", opt =>
{
    opt.Add("default", "", "Default.dothtml");
    opt.Add("Files", "files", "FileOption.dothtml");
    opt.Add("Projects", "projects", "ProjectsList.dothtml");
});
```

### DIFF
--- a/src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
+++ b/src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
@@ -65,14 +65,17 @@ namespace DotVVM.Framework.Routing
         /// <param name="content">Contains routes to be added</param>
         public void AddGroup(string groupName, string urlPrefix, string virtualPathPrefix, Action<DotvvmRouteTable> content)
         {
-            if (groupName == null || groupName == "")
+            if (string.IsNullOrEmpty(groupName))
+            {
                 throw new ArgumentNullException("Name of the group cannot be null or empty!");
+            }
             if (routeTableGroups.ContainsKey(groupName))
             {
                 throw new InvalidOperationException($"The group with name '{groupName}' has already been registered!");
             }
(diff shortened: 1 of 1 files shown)


src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -65,14 +65,17 @@ public DotvvmRouteTable GetGroup(string groupName)
...
+            {
                 throw new ArgumentNullException("Name of the group cannot be null or empty!");
+            }
             if (routeTableGroups.ContainsKey(groupName))
             {
                 throw new InvalidOperationException($"The group with name '{groupName}' has already been registered!");
             }
-            urlPrefix = group?.UrlPrefix + ((urlPrefix == null || urlPrefix == "") ? "" : urlPrefix + "/");
-            virtualPathPrefix = group?.VirtualPathPrefix + ((virtualPathPrefix == null || virtualPathPrefix == "") ? "" : virtualPathPrefix + "/");
+            urlPrefix = CombinePath(group?.UrlPrefix, urlPrefix);
```
exyi: Note that this is technically a runtime breaking change - it does not append the slash to UrlPrefix anymore on initialization.

adamjez: Url which ends with a slash is forbidden and exception will be thrown (if I understand your note)

exyi: I mean the format of data in `UrlPrefix` has changed.

adamjez: Okay, I get it

PR #621, opened 2018-06-27
Make DotvvmCompilationException serializable so it can be deserialized from the Dotvvm.Compiler output

Mrnda:


exyi:
Makes sense, thanks.

PR #622, opened 2018-06-27
Nested RouteTableGroup

Rumbeer:
I think that when creating a new `RouteTableGroup` its `DotvvmRouteTable` should be added to all of its parent `DotvvmRouteTables` and not just the first one.


exyi:
I don't think so, the groups should form the hierarchy.

Rumbeer:
They still form a hierarchy, but you can retrieve all child groups from the parent (like routes) and there cannot be conflicting group names between different levels.

exyi:
Kind of, they would form a stupid hierarchy, where you'd not know if it's your child or grandchild or what... All hierarchies you can find in software have only parent-child relations recorded, you can walk the tree if you want. Have a look at syntax trees, dotvvm control trees, filesystem trees, ... there is no flattening anywhere.

Rumbeer:
Ok, that sounds right.

PR #629, opened 2018-07-22
Add C# to TypeScript compiler

mrnustik:
**Before merging wait for new version of [Buildalyzer](https://github.com/daveaglick/Buildalyzer) package to be released, without the new version of this package the Compiler wont be able to load project information.** 

Add command line tool for compiling C# code to Javascript, so it can be used in clients browser.

Requirements:
- Typescript compiler (tsc) in `PATH` environment variable

### Currently supported C# features:
- Standard imperative language features (if/else, while, do-while, for, for-each)
- Local variables declaration, access and assignments
- All unary, binary and ternary operations (+,-,*,~,....)
- Properties assignment
- Object creation and manipulation
- Invocation of other client side methods
- Access to array member via Indexer

### Currently supported .NET library methods:
- Console output
- List operations (Add, Remove, Clear, Count)
- string Contains operation

### Example usage:
This feature introduces two new attributes to DotVVM.Core project:
- `[ClientSideMethod]` - to be used for marking methods that should be translated
- `[ClientSideConstructor]` - to be used for marking constructors that should be translated

Sample client side code:

ViewModel:
```
public class ObjectOperationsViewModel : MasterpageViewModel
{
    public List<string> NamesList { get; set; } 
					= new List<string>();

    public string Name { get; set; }
    public int Age { get; set; }
    public List<PersonDto> Persons { get; set; } = new List<PersonDto>();

    [ClientSideMethod]
    public void AddPerson(string name, int age)
    {
       Persons.Add(new PersonDto(Name, Age));
    }

    [ClientSideMethod]
    public void Iterate()
    {
        for (int i = 0; i < NamesList.Count; i++)
        {
            var name = NamesList[i];
            if (!name.Contains("iterated"))
                NamesList[i] = name + " iterated";
        }
    }
}
```

View:
```
<dot:GridView DataSource="{value: Persons}">
	<dot:GridViewTextColumn ValueBinding="{value:  Name}" />
	<dot:GridViewTextColumn ValueBinding="{value: Age}" />
</dot:GridView>
Name: <dot:TextBox Text="{value: Name}" />
Age: <dot:TextBox Text="{value: Age}" />
<dot:Button Text="Add Person“ Click="{staticCommand: AddPerson(Name, Age)}" />
<dot:Button Text="Iterate“ Click="{staticCommand: Iterate()}" />
```


src/DotVVM.TypeScript.Compiler/Translators/Operations/OperationTranslatingVisitor.cs
```diff
@@ -0,0 +1,703 @@
...
+                return _factory.CreateConditionalExpression(expression, trueStatement as IExpressionSyntax, falseStatement as IExpressionSyntax, parent);
+        }
+
+        public override ISyntaxNode VisitLiteral(ILiteralOperation operation, ISyntaxNode parent)
+        {
+            var value = "";
+            if (operation.ConstantValue.HasValue) value = operation.ConstantValue.ToString();
+            if (operation.Type.IsEquivalentTo(typeof(string)))
+            {
+                value = $"'{value}'";
```
exyi: Hey, the string may contain an apostrophe. You'd rather just JSON serialize the value.

exyi:
I was discussing it with @tomasherceg a bit and we have few more comments.

There is a pretty fundamental conflict between the runtime oriented nature of everything in DotVVM and the static compiler that "only" compiles the stuff and does not care about any context. One of the actual problems that arise from that is that you have separate registration of translatable methods for the static TypeScript compiler and for the runtime binding compiler. Another problem is that you can rename a viewmodel property for client-side usage, but the static compiler will simply ignore this fact and produce incorrect code.

One possible way to overcome this problem could be integration with DotVVM compiler, which is basically a compromise between run-time and compile-time - it initializes the DotvvmConfiguration and it can use it to compile views, binding and similar things - and one of them could be the TypeScript compilation of ViewModels. Do you think it would work this way?

exyi:
I'm sorry, but I think we should close this PR. There is a number of flaws in the implementation and most importantly it basically does not integrate with the current set of options users of DotVVM have. While this thing may be usable for a demo with default options everywhere, it will fall apart when you figure out you can't reconfigure the serializer, use REST API bindings, invoke server from the translated code, register custom translations of some method, and probably many more things I have not thought of. The implementation is very independent of DotVVM and translating C# to JS is a separate huge project I certainly don't want to maintain (not to say there is a lot of more mature implementations). If we really want this (or similar) feature we should meet try to come up with a reasonable solutions to these problems. I'm quite sure that integration of similar quality could be coded to the next midnight on top of Fable and does not bring much value.

The main reason I'm closing this PR (at least for now) is that is creates a false impression that this feature is not only coming, but already implemented only waiting for package release. Truth is unfortunately very distant from that as there is not even a little interest in having this merged (well not even a comment from you @mrnustik for half a year). Of course, feel free to comment or reopen if you have any news on that.

tomasherceg:
We should meet and discuss how to implement this in a way with using DotVVM Javascript AST. I definitely want this feature in DotVVM.

S pozdravem
Tomáš Herceg

RIGANTI s.r.o. | DOTNETCOLLEGE s.r.o. | DotVVM
www.riganti.cz | www.dotnetcollege.cz | www.dotvvm.com

PR #630, opened 2018-07-24
Fixed RouteLink placed inside an element with Events.Click attribute

adamjez:
This PR fixes `RouteLink` placed inside an element with `Events.Click` attribute. Currently, the `RouteLink` redirection is canceled and command is invoked.

Example:
```HTML
<dot:GridView DataSource="{value: ...}">
        <RowDecorators>
            <dot:Decorator Events.Click="{command: ...}" />
        </RowDecorators>
        <Columns>
            <dot:GridViewTemplateColumn>
                <dot:RouteLink Text="Test Click Propagation" RouteName="Default"/>
            </dot:GridViewTemplateColumn>
        </Columns>
</dot:GridView>
```

What do you think about letting people control when the `event.stopPropagation()' is rendered and when not.


exyi:
I'm afraid this is basically a feature that was added by the `event.stopPropagation()` :(

exyi:
I don't really like this solution as it will only work for `RouteLink`s and not other clickable components (inputs, normal links, ...). I'm afraid that the only correct solution is not to use `Events.Click` on a container that contains other clickable stuff.

exyi:
Although I don't really like this unsystematic solution it's still better than nothing... @adamjez Thank you.

PR #641, opened 2018-09-05
Redesigned how _index works

exyi:
Removed JavascriptTranslator.CurrentIndex symbolic parameter, it's usage
is replaced with `$context.$index` (as we'd do with any other
non-special property of the knockout context). This fixes a bunch of
issues with data context shifting of the context.

Repeater and GridView in server rendering mode use dotvvm-SSR-foreach
and dotvvm-SSR-item binding handlers that:
1. Automatically assign `$index` in each template
2. Remove the DataSource expression from every item of the Repeated

Method BindingHelper.SetDataContextForItem was removed since it can't be
implemented without the CurrentIndexParameter in the ParametrizedCode.

This commit should fix issues with $index in server rendering, $index in
staticCommand and $index in somewhat deeply nested data contexts.

### DIFF
--- a/src/DotVVM.Framework/tsconfig.json
+++ b/src/DotVVM.Framework/tsconfig.json
@@ -8,7 +8,7 @@
     "outFile": "Resources/Scripts/DotVVM.js",
     "declaration": true,
     "strictNullChecks": true,
-    "lib": ["dom", "es2015.promise", "es5", "es2015.symbol"]
+    "lib": ["dom", "es2015.promise", "es5"]
   },
   "files": [
     "Resources/Scripts/DotVVM.Polyfills.ts",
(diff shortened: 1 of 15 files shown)


src/DotVVM.Framework/tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "outFile": "Resources/Scripts/DotVVM.js",
     "declaration": true,
     "strictNullChecks": true,
-    "lib": ["dom", "es2015.promise", "es5"]
+    "lib": ["dom", "es2015.promise", "es5", "es2015.symbol"]
```
adamjez: What's the purpose of including `es2015.symbol`? Symbols are not supported by IE.

adamjez:
Merging it since more people hit the _index bug.

PR #648, opened 2018-10-13
DefaultOutputRenderer adds Content-Length header

exyi:
This should making the HTTP request/response a bit cheaper, as it would not have the overhead of chunked encoding while having the possibility to reuse the TCP/TLS connection.

Note that I have changed protected API of `DefaultOutputRenderer`, but I guess nobody would use that. The change only applies to GET requests, postbacks are rendered using the `IViewModelSerialized` and I did not want to break this API for a little benefit.


src/DotVVM.Framework/Runtime/DefaultOutputRenderer.cs
```diff
@@ -32,8 +33,9 @@ public virtual async Task WriteHtmlResponse(IDotvvmRequestContext context, Dotvv
             context.HttpContext.Response.ContentType = "text/html; charset=utf-8";
             SetCacheHeaders(context.HttpContext);
             var html = RenderPage(context, view);
+            context.HttpContext.Response.Headers["Content-Length"] = html.Length.ToString();
             CheckRenderedResources(context);
-            await context.HttpContext.Response.WriteAsync(html);
+            await html.CopyToAsync(context.HttpContext.Response.Body);
```
adamjez: We should dispose the `html` stream, even if it's `MemoryStream` to avoid unnecessary use of finalizer.

exyi: You are right that not disposing something disposable is not a best practise, but `MemoryStream` nor `Stream` don't have any finalizer. https://source.dot.net/#System.Private.CoreLib/shared/System/IO/MemoryStream.cs,1a4dcb744a23ba6f

I have added a `using` block anyway, the MemoryStream is not sealed and in theory someone could plug it's own implementation that could rely on the Dispose method.

PR #650, opened 2018-10-18
Number binding formatting

martinbojnansky:
**dotvvm.globalize.formatString** returns null when combined with **dotvvm.globalize.bindingNumberToString**, because value is changed to null when date parsing is unsuccessful.


exyi:
> could you please review that and close this PR?

Ok, sorry for asking for that hard thing. Closing it myself.

PR #652, opened 2018-10-27
Various fixes of Observable handling in the JS Translator

### LINKED ISSUE #604: _api.RefreshOnChange doesn't work with multi-property expressions
Using `_api.RefreshOnChange(..., Prop1 + Prop2)` renders Prop1() + Prop2()` which cannot be observed.
The reason is that `JsViewModelPropertyAdjuster` is executed after the translation, so the viewmodel properties are missing some annotations.

I have tried to fix this in https://github.com/riganti/dotvvm/commit/ba572bf6667aa43b21fd8fb90b8480a16c37c743, but it requires more time and testing to do that properly.

exyi:
Resolves #604, the expression `_api.RefreshOnChange(..., Prop1 + Prop2)`
translated to something that wraps the Prop1+Prop2 to `ko.pureComputed`.
This required a new JsNode annotation that instructs the
KnockoutObservableHandling visitor to execute a certain function during
the execution of the visitor.

Also added a bunch of tests of assignments into assignments and fixed
the behavior.


exyi:
We have walked through this with @tomasherceg personally, he has added some tests, so I'm merging it.

PR #664, opened 2019-01-28
Added caching interface to framework

quigamdev:


src/DotVVM.Framework/Runtime/Caching/IDotvvmCache.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DotVVM.Framework.Runtime.Caching
+{
+    public interface IDotvvmCache
+    {
+        T GetOrAdd<T>(object key, Func<object, DotvvmCachedItem<T>> factoryFunc);
```
exyi: If it would have signature `T GetOrAdd<T, TKey>(TKey key, Func<TKey, DotvvmCachedItem<T>> factoryFunc);` there would be no need for casts in the factory.

src/DotVVM.Framework/Runtime/Caching/IDotvvmCache.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public DotvvmCachedItem(T value)
+        {
+            this.Value = value;
+        }
+
+        public T Value;
+        public TimeSpan? Expiration;
+        public DotvvmCacheItemPriority Priority;
+        public TimeSpan? SlidingExpiration;
```
exyi: What about making all of these immutable and assignable through constuctor?

src/DotVVM.Framework/Runtime/Caching/IDotvvmCache.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public class DotvvmCachedItem<T>
+    {
+        public DotvvmCachedItem(T value, DotvvmCacheItemPriority priority)
+        {
+            this.Value = value;
+            Priority = priority;
+        }
+
+        public DotvvmCachedItem(T value)
```
exyi: Why not one constructor with default parameter? Then it immediately obvious what are the default and which parameters can be used.

src/DotVVM.Framework/Runtime/Caching/IDotvvmCache.cs
quigamdev: Implemented, I have just changed the order of params to be `<Tkey, TValue>`

exyi:
Ok, thanks. CacheAdapter also makes sense to me.

PR #666, opened 2019-02-06
Enforced client side validation by default.

quigamdev:
Issue description: 
When a string value is in the textbox which was bound to nullable numeric type then validation pass. 
This is wrong behavior. Validation has to fail and a user should be notified that value that is in the textbox has an incorrect format.

Solution: 
I have enforced DotvvmEnforceClientFormat on all numeric nullable types. Due to compatibility, I have just added this attribute to these properties and added Enforce property to be able to disable it.


src/DotVVM.Core/ComponentModel/DataAnnotations/ClientFormatAttribute.cs
```diff
@@ -9,15 +9,24 @@ namespace System.ComponentModel.DataAnnotations
     /// Checks whether the value matches the format specified by DotVVM control (e.g. TextBox).
     /// The value can be empty string or null. If it is not, it must match the FormatString that the control specifies.
     /// </summary>
+    [Obsolete("The client format is enforced by default. If you need to disable it use ClientFormatAttribute", true)]
```
exyi: I'd also include a note that `ClientFormatAttribute` also allows you to configure the behavior.

src/DotVVM.Framework/ViewModel/Validation/ViewModelValidationRuleTranslator.cs
```diff
@@ -34,7 +36,9 @@ public virtual IEnumerable<ViewModelPropertyValidationRule> TranslateValidationR
                         validationRule.ClientRuleName = "range";
                         validationRule.Parameters = new[] { rangeAttr.Minimum, rangeAttr.Maximum };
                         break;
-                    case DotvvmEnforceClientFormatAttribute enforceClientFormatAttr:
+                    case DotvvmClientFormatAttribute enforceClientFormatAttr:
+                        addEnforceClientFormat = false;
+                        if (enforceClientFormatAttr.Disable) yield break;
```
exyi: Hey, this `yield break;` interrupts the execution and will suppress any following validators.

src/DotVVM.Framework/ViewModel/Validation/ViewModelValidationRuleTranslator.cs
```diff
@@ -46,9 +50,21 @@ public virtual IEnumerable<ViewModelPropertyValidationRule> TranslateValidationR
                         validationRule.ClientRuleName = string.Empty;
                         break;
                 }
-      
+
                 yield return validationRule;
             }
+            // enforce client format by default
+            if (addEnforceClientFormat && property.PropertyType.IsNullable() && Nullable.GetUnderlyingType(property.PropertyType).IsNumericType())
```
exyi: You should also add it to nullable `DateTime`, not only numerics

src/DotVVM.Samples.Common/ViewModels/FeatureSamples/Validation/EnforceClientSideValidationDisabledViewModel.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Linq;
+using System.Text;
+using DotVVM.Framework.ViewModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace DotVVM.Samples.BasicSamples.ViewModels.FeatureSamples.Validation
+{
+	public class EnforceClientSideValidationDisabledViewModel
+    {
+        [DotvvmClientFormat(Disable = true, ErrorMessage = "This message will not be display. Because this client format enforcing is disabled.")]
```
exyi: display**ed**

exyi:
Note that this PR includes quite nasty breaking changes, specifically when viewmodel includes a nullable numeric property it's validation is enforced, which was not the case previously. We should make sure that we inform everyone to verify that nothing breaks in their applications.

src/DotVVM.Framework/ViewModel/Validation/ViewModelValidationRuleTranslator.cs
quigamdev: Oh, that's a silly mistake. Sorry about that.

quigamdev:
I have made some additional changes. @exyi Please, could you review the changes?

src/DotVVM.Framework/ViewModel/Validation/ViewModelValidationRuleTranslator.cs
```diff
@@ -34,10 +36,16 @@ public virtual IEnumerable<ViewModelPropertyValidationRule> TranslateValidationR
                         validationRule.ClientRuleName = "range";
                         validationRule.Parameters = new[] { rangeAttr.Minimum, rangeAttr.Maximum };
                         break;
-                    case DotvvmEnforceClientFormatAttribute enforceClientFormatAttr:
+                    case DotvvmClientFormatAttribute enforceClientFormatAttr:
+                        addEnforceClientFormat = false;
+                        if (enforceClientFormatAttr.Disable)
+                            break;
```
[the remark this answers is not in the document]
quigamdev: When ECFAttribute.Disable is false, it means that the whole attribute should be skipped.

PR #668, opened 2019-02-09
Fixed unique ID generation for some cases

exyi:
We are now using unique id counter instead of relying on List.Count,
which solves problems with controls added in a strange order (like
`x.Add(controlA); x.Add(controlB); x.Remove(controlA); x.Add(controlC)`,
where B and C remain in `x` with the same id).

When a control is added to the collection, it's id is assigned only if
it's parent also has the id. And when the id is assigned, it's assigned
recursively to the entire subtree that does not have any ids. This solves
problems with entire hiearchies being added into the control tree (that
are heavily used in DotVVM BP)

@quigamdev Could you please check that it really solves these problems in BusinessPack and that it does not create new ones?


quigamdev:
I have tested these changes in DotVVM BusinessPack and I have committed change of order of IDs assignation and execution of lifecycles. Everything else seems to be ok. Tests are passing, so I will merge this PR.

PR #674, opened 2019-03-10
Infrastructure for knockout named templates

exyi:
I have taken @adamjez's work from almost a year ago and modified it to be a bit "more conservative" in order to work with BP controls.

This solves the problem when a repeater with separator templare is placed inside a separator template. Similar approach should be taken by some BP controls that use inline `<script type="text/html" ...` templates.

There is a new API for that:
* `string ResourceManager.AddTemplateResource(string)` registers a new resource that will be rendered as the knockout `script` based template. It will be assigned a unique Id that is based on it's content (a cryptografic hash) and the Id will be returned for you to use.
* `string ResourceManager.AddTemplateResource(IDotvvmRequestContext, DotvvmControl)` - does basically the same as previous method, it just renders the provided control. Thus


quigamdev:
@exyi These changes seem to work nice and smooth. I added a sample view. I'm gonna add more samples and UI tests.

exyi:
Ok, thank you.

PR #679, opened 2019-04-04
Added RedirectToLocalUrl

tomasherceg:
The implementation of `IsLocalUrl` is copied from ASP.NET Core MVC.

### DIFF
--- a/src/DotVVM.Framework.Tests.Common/Routing/UrlHelperTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Routing/UrlHelperTests.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DotVVM.Framework.Routing;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DotVVM.Framework.Tests.Common.Routing
+{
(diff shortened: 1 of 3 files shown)


exyi:
I have merged this, as it seemed to be ok. Then it I decided I will at least try to find a bug in the heuristic `IsLocalUrl` ... and it was not that hard ... So I reverted the merge and it's impossible to reopen the PR :( Sorry for messing it up.

For some reason, newlines are removed from the URL when it's redirected on postback (in both FF and Chrome). This means that this redirects to a different domain, although seems to be a local url to the `IsLocal` function:
```
/
/google.com
```

I have no idea why does it happen and if it happens with in any other case :(

tomasherceg:
I have gone through the specification of URI () and here are the characters that can appear in path, query and fragment. We should throw an exception if the user tries to add there any other character:

```
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz
0123456789
-._~
%
!$&'()*+,;=
:@/?
```

exyi:
Ok, I have reverted that revert of the merge and merged your fixes :D. So it should be included in the master branch. Thank you for the implementation, hope this API will help people get they apps more XSS-proof

Issue #671, opened 2019-02-20
Route constraints & default parameter values


tomasherceg:
We are missing multiple features:
* Multiple route constraints
* Default values in route URLs (`{param:contraint1:constraint2=defaultValue}`)

exyi:
>    * Default values in route URLs (`{param:contraint1:constraint2=defaultValue}`)

We have a separate parameter for default values:
```C#
routes.Add("R1, ""something/{param123}", "R1.dothtml", defaultValues: new { param123 = "default" });
```

ridercz:
> We have a separate parameter for default values:

Yes, but inline specification is more concise and is present in MVC/Razor for several versions now and people are used to it. The separate parameter is old-school :)

exyi:
> inline specification is more concise and is present in MVC/Razor for several versions now and people are used to it. The separate parameter is old-school :)

Well, I'm not closing this issue - just point out you can already do that, so the feature is not missing. However, having duplicate API for such a simple thing is also not something we are looking for and we certainly can't get rid of the old-school one.


@tomasherceg Btw, do you have a use case for multiple constraints? There is also a slight problem with that - the constraint is actually a type specifier and provides a [parse function](https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Routing/IRouteParameterConstraint.cs#L12) that transforms the string into an object (`{x:int}` will be `int` not `string`). When we'd have more of these parameters, the result type could be ambiguous.

ridercz:
The use case for multiple constraints is when you want to specify type and then some other restricting condition, ie. `"product/{id:int:min(1)}"` - the product ID has to be integer greater than zero

tomasherceg:
I have implemented the support for multiple route constraints and default values in routes, but I need to make sure it behaves the same way as in MVC. And I really don't like how the built-in route constraints in DotVVM are implemented, so I plan to do some refactoring.

PR #680, opened 2019-04-06
Support for multiple route constraints and default values in routes

tomasherceg:
See https://github.com/riganti/dotvvm/issues/671 for details of the feature.

* I have refactored the way how route constraints were implemented. 
* I have also added some tests - the route constraints were not covered properly.
* And finally, I have found and fixed a bug in parameter matching - when the route URL contained nothing but a parameter which was optional, it was not matched when the parameter was not specified.

There is a **breaking change** in API, but I believe it affects only the users who wrote their own route constraints and we can introduce this in `2.3`. First, I tried to avoid this breaking change, but the implementation of the route constraints would be more difficult and confusing. I wanted to keep it as simple as possible.

* Removed `GenericRouteParameterType` class - implemented each constraint as a separate class (with some base classes). I believe that nobody except us has used this class so it was safe to remove.
* Changed `ParseString` that accepted `string` values to `ParseValue` which accepts `object`. This will affect people who implemented their own constraints, but the fix is very simple and there are not many constraints that need to worry about other than string types. The first constraint of a parameter always gets `string`, and only constraints that parse values (`int`, `bool`, `Guid` etc.) convert these values and pass them converted to the following constraints.


tomasherceg:
We have decided to remove the conversion functionalities from the route constraints and to keep them string-only.

However, I have done a quick research and the number of changes is huge and it involves:
* The route constraints themselves - they need to be changed to work with strings.
* The default parameter values which can be objects.
* The property group of RouteLink and other controls.

quigamdev:
This would break many people. So we decided to close this PR without merging.

PR #683, opened 2019-04-24
Returned files support changing the Content-Disposition header type

tomasherceg:
One of the customers asked us to support different `Content-Disposition` headers - he needs `inline` instead of `attachment` which was hard-coded.

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
@@ -174,13 +174,13 @@ public static class DotvvmRequestContextExtensions
     /// <summary>
     /// Redirects the client to the specified file.
     /// </summary>
-    public static void ReturnFile(this IDotvvmRequestContext context, byte[] bytes, string fileName, string mimeType, IEnumerable<KeyValuePair<string, string>> additionalHeaders = null) =>
-        context.ReturnFile(new MemoryStream(bytes), fileName, mimeType, additionalHeaders);
+    public static void ReturnFile(this IDotvvmRequestContext context, byte[] bytes, string fileName, string mimeType, IEnumerable<KeyValuePair<string, string>> additionalHeaders = null, string attachmentDispositionType = null) =>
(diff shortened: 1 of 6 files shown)


src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -174,13 +174,13 @@ public static string TranslateVirtualPath(string virtualUrl, IHttpContext httpCo
     /// <summary>
     /// Redirects the client to the specified file.
     /// </summary>
-    public static void ReturnFile(this IDotvvmRequestContext context, byte[] bytes, string fileName, string mimeType, IEnumerable<KeyValuePair<string, string>> additionalHeaders = null) =>
-        context.ReturnFile(new MemoryStream(bytes), fileName, mimeType, additionalHeaders);
+    public static void ReturnFile(this IDotvvmRequestContext context, byte[] bytes, string fileName, string mimeType, IEnumerable<KeyValuePair<string, string>> additionalHeaders = null, string attachmentDispositionType = null) =>
```
exyi: This is chaging the binary API, so already compiled libraries will break on MethodNotFoundException. We had quite a lot of these errors recently, I guess it's the right time to stop doing so on minor releases. Could you please add the original overload so we don't have to worry about BP/BS/Contrib/Whatever compatibility. We can then remove it with the next not-so-minor release (maybe we could use a marker comment?).

I guess it will be better to make `"attachment"` the default, not `null`. It will be clearer.

tomasherceg: For this case, we can add another overload, but I think that we won't be able to avoid binary changes between minor versions in general. 
That's why we try to upgrade and publish all NuGet packages with every minor release, and I'd like to keep the convention that `DotVVM 2.2.something` should work with `DotVVM.BusinessPack 2.2.something`.

exyi: Of course it's not possible to avoid it in general. But adding that overload in this case will save us quite a lot of trouble with synchronizing the release of everything, so we can release basically immediately. We can remove it when we will make some unavoidable breaking change.

PR #686, opened 2019-04-25
Fixed persisted viewmodel in Chrome

### LINKED ISSUE #685: Chrome doesn't remember viewmodel after history change
When you leave the page by going back and then going forward, the viewmodel is reset to its initial state.

<details>
    <summary>Example (GIF)</summary>
    
</details>

This is because Chrome doesn't remember the value of the `<input type="hidden">` across history states. If we change the following in [BodyResourceLinks.cs](https://github.com/riganti/dotvvm/blob/dccddcf2bc8f2688af454e3831f3324801371e8d/src/DotVVM.Framework/Controls/Infrastructure/BodyResourceLinks.cs#L31-L36):

```diff
             // render the serialized viewmodel
             var serializedViewModel = ((DotvvmRequestContext) context).GetSerializedViewModel();
-            writer.AddAttribute("type", "hidden");
+            writer.AddAttribute("type", "text");
+            writer.AddAttribute("style", "display:none!important;");
             writer.AddAttribute("id", "__dot_viewmodel_root");
             writer.AddAttribute("value", serializedViewModel);
             writer.RenderSelfClosingTag("input");
```

then Chrome will remember the viewmodel:

<details>
    <summary>Example (GIF)</summary>
    
</details>

When I searched for this issue, I found an issue from 2010 with the same problem ([here](https://stackoverflow.com/questions/4025843/chrome-doesnt-cache-hidden-form-field-values-for-use-in-browser-history)) in WebForms. So I don't think this behavior won't change in Chrome if it didn't change in 9 years.

What would be the best solution here? Should we change the viewmodel input to an text-field and hide it (with maybe an setting in `DotvvmConfiguration` for compability)?

---

**Browser:** Google Chrome
**Version:** 73.0.3683.103 (Official Build) (64-bit)
**Dotvvm build commit:** https://github.com/riganti/dotvvm/commit/dccddcf2bc8f2688af454e3831f3324801371e8d
**Hosting:** ASP.Net Core

GerardSmit:
See #685 for more information. Maybe we should add an setting in `DotvvmConfiguration` to configure this behaviour incase someone uses an element selector that searches for `input[type=hidden]` in his code.

---

fixes #685

### DIFF
--- a/src/DotVVM.Framework/Controls/Infrastructure/BodyResourceLinks.cs
+++ b/src/DotVVM.Framework/Controls/Infrastructure/BodyResourceLinks.cs
@@ -30,7 +30,8 @@ namespace DotVVM.Framework.Controls
 
             // render the serialized viewmodel
             var serializedViewModel = ((DotvvmRequestContext) context).GetSerializedViewModel();
-            writer.AddAttribute("type", "hidden");
+            writer.AddAttribute("type", "text");
+            writer.AddAttribute("style", "display:none!important;");
             writer.AddAttribute("id", "__dot_viewmodel_root");
             writer.AddAttribute("value", serializedViewModel);
             writer.RenderSelfClosingTag("input");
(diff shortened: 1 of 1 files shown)


GerardSmit:
Actually, I just realized that `input[type=input]` can have issues with autocomplete. Maybe we should add the attribute `autocomplete=off` on the input just to be sure.

GerardSmit:
This solution will be re-written with the History API, see https://github.com/riganti/dotvvm/issues/685#issuecomment-486970843 for more information.

PR #693, opened 2019-05-16
Checkbox - render the same html structure when text is empty or not

quigamdev:
This is not a nice breaking change. The current state is that checkbox is not rendering label when the text is empty. This PR is just removing the case when the text is empty. Unfortunately, the structure of rendered HTML is not correct as well. I think it would be better to render the input element and then the label element. The label element should contain an ID of the checkbox in the 'for' attribute.

### DIFF
--- a/src/DotVVM.Framework/Controls/CheckableControlBase.cs
+++ b/src/DotVVM.Framework/Controls/CheckableControlBase.cs
@@ -10,7 +10,6 @@ namespace DotVVM.Framework.Controls
     /// </summary>
     public abstract class CheckableControlBase : HtmlGenericControl
     {
-        private bool isLabelRequired;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


exyi:
I get the point that it's messy when some comboboxes are wrapped in `<label>` and some don't. But this way, there no way to create just the `<select>` element without the `<label>`, which is quite unfortunate in case you want to handle the label yourself. I think we could render the `<label>` in case the property `Text` is not set or it's equal to `null`.

> Unfortunately, the structure of rendered HTML is not correct as well.

What's wrong about it, having the input element inside the label should be perfectly fine (https://www.w3.org/TR/html52/sec-forms.html#the-label-element)

tomasherceg:
We've decided to keep the current CheckBox behavior.

PR #697, opened 2019-05-28
Fixed update of validator binding

### LINKED ISSUE #691: Validation doesn't get updated when value in ViewModel changes
I'm facing an issue with validation not responding to changes in the ViewModel.
The scenario I have is that I have a page with a ComboBox and a form beneath it. Based on the value selected in the ComboBox I either need to fill in the **Address** in the form or not.

1. When I select value **_A_** in the ComboBox, I set **Address** property to a new object and show the form fields associated with it (by using _IncludeInPage_). Then the form validation works fine, the address gets validated and if there are any missing values, the validation errors get displayed in the form.
2. When I select value **_B_** in the ComboBox I set the **Address** property to _null_ and hide the associated form fields. Then the validation works fine,  address doesn't interfere with the validation and the validation gets displayed for the other fields within the form (non-address fields).
3. When I select value **_A_** in the ComboBox (and the property gets set to null), however then I decide to select value **_B_** (and the property gets set to a new object) - the validation errors for the address fields don't show up.

The ComboBox calls a command on the ViewModel when the Changed event gets called and it sets the value for **Address** in the C# code.
I use server-side validation in this scenario, calling an API from the backend and getting the validation errors. I've checked that I get the errors correctly in all the cases. The errors and their structure are the same in cases 1 and 3. The object is initialized in case 3 and the data binding works (I'm sending it to the API server and it gets the values correctly). But the validation errors fail to show up in case 3.

exyi:
Fixed a bug in `dotvvmValidation` binding - it ignored updates of the observable so it have not displayed anything when the view model was not loaded with the page.

Btw, I think we might have very similar in the other binding handlers too, as they also ignore the `update` event.

resolves #691 

EDIT: sorry, I have pressed Ctrl+Enter too early


tomasherceg:
I have fixed an unrelated bug reported by a customer - when there were multiple errors, the `InvalidCssClass` was added multiple times and it was not removed properly. 
We were concatenating the value of `class` attribute manually (and it was wrong), so I have changed the implementation to use `element.classList` collection (it is supported in all browsers included IE11).

PR #699, opened 2019-06-04
Validation of coupled properties in RadioButton, CheckBox and ComboBox

exyi:
It now validates that properties `DataSource`/`SelectedValue`,
`CheckedItems`/`CheckedValue` and `CheckedItem`/`CheckedValue` are of compatible
types. Where compatible means that they are basically of the same, with
tolerance for the writable property being nullable. For example, for `RadioButton` with properties `CheckedValue` and `CheckedItem` set

```
CheckedValue T  & CheckedItem T   -> ok, it's the same type
CheckedValue T? & CheckedItem T?  -> ok, it's the same type
CheckedValue T  & CheckedItem T?  -> ok, you can assign non-nullable into nullable
CheckedValue T? & CheckedItem T   -> error, you can't assign nullable into non-nullable 
```

Note that this is a breaking change, it add the validation to `ComboBox` and forbids some questionable usage. It should always fail during page compilation, so it's at least not anything nasty that occurs at runtime in strange conditions. Anyway, if we don't want that breaking change, we can merge only the `CheckBox` and `RadioButton` changes, as they are more lax than before.


exyi:
Also, we should make sure that BusinessPack controls that have these coupled properties do something similar.

tomasherceg:
We should finish the TODOs in tests. Then we can merge.

src/DotVVM.Framework/Compilation/Validation/DefaultControlUsageValidator.cs
```diff
@@ -59,12 +59,15 @@ public IEnumerable<ControlUsageError> Validate(IAbstractControl control)
...
         {
+            if (type == typeof(object)) return new MethodInfo[0];
             var methods = type.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static)
                 .Where(m => m.IsDefined(typeof(ControlUsageValidatorAttribute)))
                 .ToArray();
-            if (methods.Length > 0) return methods;
-            else if (type == typeof(object)) return new MethodInfo[0];
-            else return FindMethods(type.GetTypeInfo().BaseType);
+
+            var overrideValidation = methods.Select(s => s.GetCustomAttribute(typeof(ControlUsageValidatorAttribute))).Any(s => ((ControlUsageValidatorAttribute)s).Override);
```
exyi: It's probably ok that this is a breaking change, but we should make sure that all declarations are consistent in the `Override` property. It's not very intuitive, what this should do (if not error):

```
[ControlUsageValidatorAttribute(Override = true)
Method1()...

[ControlUsageValidatorAttribute(Override = false)
Method2()...
```

src/DotVVM.Framework/Controls/SelectorBase.cs
```diff
@@ -104,5 +109,29 @@ public IValueBinding ItemTitleBinding
...
             DotvvmProperty.Register<IValueBinding, SelectorBase>(nameof(ItemTitleBinding));
+
+        [ControlUsageValidator]
+        public static IEnumerable<ControlUsageError> ValidateUsage(ResolvedControl control)
+        {
+            if (control.Properties.ContainsKey(SelectorBase.ItemValueBindingProperty) &&
+                control.Properties.GetValue(SelectorBase.ItemValueBindingProperty) is ResolvedPropertyBinding valueBinding)
+            {
+                var t = valueBinding.Binding.ResultType;
+                if (!t.IsPrimitiveTypeDescriptor())
```
exyi: You also actually don't need the `IsPrimitiveTypeDescriptor` method implemented on `ITypeDescriptor` since `valueBinding.Binding.ResultType` will always be `ResolvedTypeDescriptor`.

src/DotVVM.Framework/Compilation/ControlTree/TypeDescriptorUtils.cs
```diff
@@ -1,11 +1,49 @@
...
+            return type.IsAssignableTo(EnumTypeDescriptors);
+        }
+        public static bool IsStringTypeDescriptor(this ITypeDescriptor type)
+        {
+            return StringTypeDescriptors.IsEqualTo(type);
+        }
+
+        public static bool IsPrimitiveTypeDescriptor(this ITypeDescriptor type)
+        {
+            return type.IsStringTypeDescriptor()|| type.IsNumericTypeDescriptor() || type.IsEnumTypeDescriptor();
```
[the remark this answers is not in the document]
quigamdev: I added another method IsPrimitiveOrEnumTypeDescriptor.

src/DotVVM.Framework/Compilation/Validation/DefaultControlUsageValidator.cs
quigamdev: I have added the consistency check.

tomasherceg:
I have fixed some bugs in current validation rules and added more tests.
I had to consolidate and move few helper methods (IsPrimitiveType, IsCollection, IsComplexType) because they were spawned on multiple places in the project and had slightly different behavior - for example IsPrimitiveType on TypeDescriptor behaved differently (and incorrectly) than the one on Type etc. This change polluted the diff of this PR, but I believe that it fixed many tiny bugs.

I have also reviewed the entire PR so we should be able to merge it (if UI tests pass). 
Since this functionality is not covered by many UI tests, we have to test this extensively on customer projects to be sure we haven't made bugs in the type validation.

quigamdev:
All tests have passed. @exyi Can you check the new changes or do you think it is ready for merge?

tomasherceg:
Merging now because we need to use it on a customer project asap.

exyi:
> Merging now because we need to use it on a customer project asap.

Then you should have used the version from git instead of messing this PR up.

PR #703, opened 2019-06-14
NuGet package with TypeScript definitions

tomasherceg:
The `DotVVM.Typescript.Definitions` package adds DotVVM.d.ts (and its dependencies) in the project. 

For .NET Framework, the regular `content` mechanism in NuGet packages is used. 

For .NET Core, the `contentFiles` mechanism is replaced by a custom MSBuild target which copies the files in the project before build. The `contentFiles` are just a MSBuild thing and they don't physically copy the files in the project so `tsc` cannot see them. The path where the files are placed can be customized using the `DotVVMTypescriptDefinitionsDestinationPath` MSBuild variable (only on .NET Core).

We should publish internal NuGet packages from this feature branch and test the entire workflow before we merge in master. I'd like to see a real experience on a real project before we push the package out as I am not sure whether the .NET Core "hacky" way of doing the stuff is OK.


tomasherceg:
We won't go further with this solution as it is quite hacky. 
We'll publish npm package instead of this.

PR #705, opened 2019-06-16
Lazy csrf tokens

exyi:
The CSRF token can be loaded after the page is displayed.

* When the config option `CsrfTokenLazyLoading` is enabled, CSRF token is not rendered by default and will be requested before the first postback. This removes one source of non-determinism and may allow easier caching of entire DotVVM pages.
* In any case, when the CSRF token or Session ID is invalid (for any reason), the server returns that as a JSON error message, the client automatically asks for a new token and retries the postback.


tomasherceg:
I have merged new commits from master and changed the way how this feature is switched on using the new feature flags API - it allows to enable or disable it for particular routes. 
I think we can now merge this feature in master and create an internal release to test it on real projects.

tomasherceg:
I'd like to merge this feature in order to test it on real projects. Or do you think it is not ready for that yet?

exyi:
Yea, it should be probably usable.

I'd still like to combine the error response with "view model not cached" and maybe add some smarter retry logic to the fetch of the CSRF token.

tomasherceg:
Ready for merge after resolving the conflicts.

PR #706, opened 2019-06-27
Implement InlineStylesheetResource for small piece of CSS

martindybal:


exyi:
It makes sense to me to have this resource, but I don't really like how the API is so different from InlineScript even though one would expect them to be the same. Do you think we could unify them, possibly by extending both of them to support creation from both string and resource location?

If we are loading from resource location, it should not cache the result in debug mode. It's annoying to restart the server just to invalidate a resource cache.

Last but not least, we should make sure that the content of the code can not end the `style` element.

tomasherceg:
We have to solve an issue with possible XSS - if the inline style contains user-provided content, it could end the style definition and inject a script in the page. We need to detect all variants of ending the `<style>` section.

exyi:
I have included a check that the style resource does not contain a `</style` substring. The specification claims that `style` and `script` elements are "raw text" (https://html.spec.whatwg.org/multipage/syntax.html#raw-text-elements), and that means that it is ended by a `</name-of-element` substring (https://html.spec.whatwg.org/multipage/syntax.html#cdata-rcdata-restrictions).

@tomasherceg please have a look, and we can merge it.

src/DotVVM.Framework/ResourceManagement/InlineScriptResource.cs
```diff
@@ -22,23 +23,33 @@ public InlineScriptResource(string code, ResourceRenderPosition renderPosition =
...
             }
         }
 
         internal static void InlineScriptContentGuard(string code)
         {
+            // We have to make sure, that the element is not ended in the middle.
+            // <style> and <script> tags have "raw text" content - https://html.spec.whatwg.org/multipage/syntax.html#raw-text-elements
+            // and those element must not contain "</name-of-the-element" substring - https://html.spec.whatwg.org/multipage/syntax.html#cdata-rcdata-restrictions
             if (code?.IndexOf("</script", StringComparison.OrdinalIgnoreCase) >= 0)
                 throw new Exception($"Inline script can't contain `</script>`.");
```
tomasherceg: The error message is technically not true because the guard will prevent the user to use for example `</scriptContent>` or `</scripts>` which can be used in a string legitimately. I know it is difficult to restrict only `script` end tag because HTML allows attributes on closing tags, so never mind - we can just restrict that. But I would remove the `>` characted from the error message so it is more accurate.

exyi:
On 11/5/19 5:21 PM, Tomáš Herceg wrote:
> The error message is technically not true because the guard will prevent
> the user to use for example |</scriptContent>| or |</scripts>| which can
> be used in a string legitimately. I know it is difficult to restrict
> only |script| end tag because HTML allows attributes on closing tags, so
> never mind - we can just restrict that. But I would remove the |>|
> characted from the error message so it is more accurate.

That's right, I have fixed that.

PR #711, opened 2019-07-08
Render disabled attribute of RouteLink on server

ENgateman:
Render the HTML disabled attribute, when the value of the `EnabledProperty` is false on the server.

Should some visual flashing issues on page load.

### DIFF
--- a/src/DotVVM.Framework/Controls/RouteLink.cs
+++ b/src/DotVVM.Framework/Controls/RouteLink.cs
@@ -11,6 +11,10 @@ namespace DotVVM.Framework.Controls
     /// </summary>
     public class RouteLink : HtmlGenericControl
     {
+        private const string EnableKnockoutBinding = "dotvvmEnable";
+
+        private bool shouldRenderText = false;
+
         /// <summary>
         /// Gets or sets the name of the route in the route table.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/DotVVM.Framework/Controls/RouteLink.cs
```diff
@@ -67,56 +71,47 @@ public RouteLink() : base("a")
...
-                WriteEnabledBinding(writer, (bool) enabledBinding);
+                writer.AddKnockoutDataBind(EnableKnockoutBinding, isEnabled.ToString().ToLower());
             }
-            else if (enabledBinding is IValueBinding)
+            else if (enabledValue is IValueBinding enabledBinding)
             {
-                WriteEnabledBinding(writer, (IValueBinding) enabledBinding);
+                writer.AddKnockoutDataBind(EnableKnockoutBinding, enabledBinding.GetKnockoutBindingExpression(this));
+            }
+            if (!GetValue<bool>(EnabledProperty))
```
quigamdev: When GetValue returns null this code will throw an exception.

ENgateman:
@exyi Reverted. Sorry about that. I knew this wasn't gonna fly but I had to try anyway.

(This also applies to #714 and #721.)

PR #714, opened 2019-07-11
Detect circular dependencies among resources

### LINKED ISSUE #710: Resources - Circular dependency - StackOverflow exception
Stack overflow exception is currently thrown when circular dependency exists in resources.
It would be better to throw some normal exception during resource registration.

ENgateman:
Fixes #710.

~The first commit merely satisfies my OCD. It's the second one that needs attention.~


src/DotVVM.Framework.Tests.Common/ResourceManagement/ResourceRepositoryTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DotVVM.Framework.Tests.Common.ResourceManagement
+{
+    [TestClass]
+    public class ResourceRepositoryTests
+    {
+        [TestMethod]
+        public void ResourceRepository_CyclicDependency_Throws()
```
adamjez: IMHO this unit test doesn't work as you intended.
If it throws or not the unit test will succeed.
If you except to throw exception you should use something like this: https://docs.microsoft.com/cs-cz/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.throwsexception?view=mstest-net-1.2.0

ENgateman: Good point. Fixed. Thank you. :)

ENgateman:
@exyi History unpolluted.

There's something I don't understand though. How exactly can it break in the render phase if the user is not able to register resources with circular dependencies? Wouldn't the check slow down the render phase?

exyi:
> How exactly can it break in the render phase if the user is not able to register resources with
> circular dependencies? Wouldn't the check slow down the render phase?

The problem is that the `Dependencies` property is mutable, so I can
register a resource without any dependency and create the cycle later.
Unfortunatelly we certainly can't remove that setter as it would break
most usages.

tomasherceg:
I think that there would be many cases like that because many things in the configuration is mutable. 
I think that we can live with that - if someone changes the configuration after startup, they deserve the exception.

ENgateman:
I added a check to the point where resources are ordered and to the `ResourceManager.AddRequiredResource` method since both caused `StackOverflowException`s when the user tampered with the dependencies after registration.

PR #717, opened 2019-07-12
Tasks sequences in commands

tomasherceg:
I have improved binding translation to support `async` methods (we don't support the `await` keyword in bindings, but it is "emulated" automatically on all async methods), and to behave correctly when Tasks are chained in command and static command bindings.

## Assignments and async methods

The `{command: Property = GetValueAsync()}` in commands is now translated to something like this:  
`GetValueAsync.ContinueWith(task => Property = task.Result)`

Therefore, DotVVM will be aware of the async operation and will await the resulting task.

## Chaining async tasks

We had an issue when multiple statements were used and one of them was async. For example, the following binding was executed incorrectly because only the first clause is a Task, but the resulting delegate is not. The command is executed but the async operation is not awaited and sometimes an exception is thrown because the async operation is not finished by the end of the HTTP request:

`{command: AsyncOperation(); somethingElse = 1}`

The sequence of async calls in commands is now handled - internally, the `ContinueWith` and `Unwrap` is used to chain all clauses. The resulting delegate is one Task that can be awaited. 

## Static commands

There is a slight problem with static commands as the binding expression is not evaluated on the server at once, but each clause is evaluated separately - either locally on the client, or a separate request is posted to the server. The rewritten expressions need to be reverted to the original form before the translation to JavaScript is performed. I have added an extra visitor that will remove the changes done in the earlier phase. In order to be sure that we match the rewrites correctly, all rewrites utilize methods defined in `CommandTaskSequenceHelpers`.

__I am not sure if I implemented the logic on the best places - maybe I overlooked something and there might be more clever solution than rewriting the expression tree twice. I'd be happy for any advice.__


tomasherceg:
We have decided to remove the unwrapping in assignments - we'll do this as a separate issue because we are considering introducing of the await keyword, or auto-unwrapping of all tasks in the expression.

PR #725, opened 2019-07-19
LinkButton and FileUpload empty href

MichalTichy:
LinkButton and File now are using javascript:; instead of # inside the href attribute when no link is set.

This change solves the problem of # being added to the URL


src/DotVVM.Framework/Controls/FileUpload.cs
```diff
@@ -210,7 +210,7 @@ private void RenderUploadButton(IHtmlWriter writer)
             writer.AddAttribute("class", "dotvvm-upload-button");
             writer.AddKnockoutDataBind("visible", "!IsBusy()");
             writer.RenderBeginTag("span");
-            writer.AddAttribute("href", "#");
+            writer.AddAttribute("href", "c");
```
tomasherceg: Is the "c" intentional, or is it a typo?

MichalTichy: typo
It should be fixed now

PR #727, opened 2019-07-22
Allow brackets in attributes name

martindybal:
I need to use [AMP Bind component](https://amp.dev/documentation/components/amp-bind/). So I have to allow brackets in attribute name

`<button [text]="'This is a attribute with special characters ' + (show ? 'true': 'false')">Show</button>`


tomasherceg:
The `ReadIdentifier` method is used to parse also the tag and directive names - this change will allow these characters in directives and tag names. I can live with that, but we should check if there aren't any other side effects of this change.

tomasherceg:
I have modified the code so we'll avoid the breaking change in tag and directive names. 
Interestingly, the HTML specs doesn't allow brackets or any other non-alphanumeric characters in HTML tag names (which is OK for us as we translate tags with these characters to plain HTML tags), but the attribute names allow almost everything (which means that our list is more restrictive than the specs).

But if the tests pass, I think we can merge this.

Issue #702, opened 2019-06-10
FallbackLocation renders script that rewrites the whole html of current page


quigamdev:
FallbackLocation renders script with a condition like this `condition || document.write("load html")`
In case that resource is requested via postback and not available then the whole HTML is rewritten by the script element and all content disappears. 

The rendering of this condition has to be more sophisticated. There are 2 scenarios. 
1) The resource is requested when the page is rendered by GET request.
2) The resource is requested via postback and this resource is rendered by script in dotvvm--internal

ENgateman:
Couldn't reproduce. Issue #715 needs to get fixed before this one.

PR #728, opened 2019-07-24
Support LocationFallback on ILinkResources

### LINKED ISSUE #715: LocationFallback of a resource doesn't get rendered
While trying to fix #702, I couldn't get a testing `LocationFallback` to load at all.

A sample can be found in the `DotVVM.Common.Samples` project in the `fix/feedback-location` branch. It's called "FeatureSamples_Resources_LocationFallback".

After the primary location of my testing resource failed, a loading script generated a new `<script>` tag that referenced - wait for it - the primary location once again. Or rather they both simply referenced the name of the resource not indicating what location should be used.
firefox_npoxby4kIR

ENgateman:
Fixes #715.

This PR adds support for `LinkResourceBase.LocationFallback` during URL building and resolution of `ILinkResources` by adding a `location` query parameter to the `dotvvmResource` route.

---

To illustrate the problem this PR solves, let's have the following resource registration:
```csharp
resources.Register("FeatureSamples_Resources_LocationFallback", new ScriptResource {
    Location = new FileResourceLocation("~/Scripts/testResource_failsToLoad.js"),
    LocationFallback = new ResourceLocationFallback(
        "window.isTestResourceLoaded",
        new FileResourceLocation("~/Scripts/testResource_locationFallback.js"))
});
```

Without the query parameter, the generated URL of the fallback location is the same as of the primary location and thus it has no effect.

---

I realize the main purpose of the `LocationFallback` property is having a backup plan for when a CDN is unavailable, and that a `FileResourceLocation` as fallback for another `FileResourceLocation` seems a little nonsensical but it is a valid configuration if only for testing the `LocationFallback` feature.


exyi:
Is there any practical need for the location distinction? Surprisingly having multiple `FileResourceLocation`s does not make much sense, since this fallback feature is designed to handle a failure to load a 3rd party resource. If we want to support failures to load local resource (due to missing file) I'd prefer to handle that server-side and not annoy the client-side about such issue.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -13,6 +14,10 @@ namespace DotVVM.Framework.ResourceManagement
 {
     public class LocalResourceUrlManager : ILocalResourceUrlManager
     {
+        private const string HashParameterName = "hash";
+        private const string NameParameterName = "name";
+        private const string LocationQueryName = "location";
```
exyi: Do you really think having constant for these trivial things helps anything? Constants are fine when they help you name the "magic literal", but in this case you just repeat the constant in the name...

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -52,16 +79,28 @@ protected virtual string DecodeResourceName(string name)
...
+            {
+                return null;
+            }
+
+            var name = DecodeResourceName((string)parameters[NameParameterName]);
+            var hash = (string)parameters[HashParameterName];
+            int? locationIndex = null;
+            if (context.Query.TryGetValue(LocationQueryName, out var locationString))
+            {
+                locationIndex = (int)ReflectionUtils.ConvertValue(locationString, typeof(int));
```
exyi: Why not `int.Parse`? This `ConvertValue` value function is not designed to take input from the network.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -77,20 +116,36 @@ private ILocalResourceLocation TryLoadAlternativeFile(string name, string hash,
...
+
+            mimeType = link.MimeType;
+            var locations = link.GetLocations().ToArray();
+            if (locationIndex == null)
+            {
+                return locations
+                    .OfType<ILocalResourceLocation>()
+                    .FirstOrDefault();
+            }
+            // unless the order of locations changes, the location has to be local
```
exyi: And when the order of locations changes then what? What does this comment says?

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -77,20 +116,36 @@ private ILocalResourceLocation TryLoadAlternativeFile(string name, string hash,
...
-                   ?.OfType<ILocalResourceLocation>()
-                   ?.FirstOrDefault();
+            if (!(resource is ILinkResource link))
+            {
+                mimeType = null;
+                return null;
+            }
+
+            mimeType = link.MimeType;
+            var locations = link.GetLocations().ToArray();
```
exyi: Why the `ToArray`?

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -52,16 +79,28 @@ protected virtual string DecodeResourceName(string name)
...
-            if (DotvvmRoutingMiddleware.FindMatchingRoute(new[] { resourceRoute }, context, out var parameters) == null) return null;
-            var name = DecodeResourceName((string)parameters["name"]);
-            var hash = (string)parameters["hash"];
+            if (DotvvmRoutingMiddleware.FindMatchingRoute(new[] { resourceRoute }, context, out var parameters) == null)
+            {
+                return null;
+            }
+
+            var name = DecodeResourceName((string)parameters[NameParameterName]);
+            var hash = (string)parameters[HashParameterName];
```
exyi: Why are the lines above just reformatted?

ENgateman:
The feature may have been designed to handle a failure in loading of a 3rd party resource but this intention is not enforced. The way it's designed (looking at the fallback condition) means it can be used to handle errors in a piece of javascript code. Having a fallback in that case could be useful.

Personally, I think we should support this scenario. But if we decide not to, I think, at the very least, we should notify the user that having multiple `FileResourceLocation`s is meaningless.

(There's also my need to test `LocationFallback` in order to fix #702, but that can be solved another way.)

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
ENgateman: I followed the logic handling query parameters until I arrived at `ParameterBindingAttribute`. `ConverValue` is used there and I wanted to be consistent.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
ENgateman: I tend to agree with you but only if the constant is used just once. If it's used multiple times, like in this case, they should imho be in constant fields.

Let's pretend there's a valid reason to change the constant and somebody changes only one or some of its occurrences. The changes would be seen in the PR's diff of course, but all the forgotten occurrences wouldn't.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
ENgateman: It meant to say, that the type of the location is guaranteed to be `ILocalResourceLocation`, which isn't true anyway. I added a condition instead.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
ENgateman: I'm assuming the order of locations provided by `GetLocations` is stable. To be frank, I'm not entirely certain whether this assumption holds water.

src/DotVVM.Framework/ResourceManagement/LocalResourceUrlManager.cs
ENgateman: Most cleanup I can refrain myself from doing. Overlong early returns on a single line in a file I'm already editing I cannot ignore. I like seeing where the method returns clearly.

ENgateman:
Merging after discussion with @exyi.

Issue #663, opened 2019-01-24
ValidationSummary performance


adamjez:
`dot:ValidationSummary` can take up to **seconds** to update knockout data binding.

The issue mainly shows up when complex page contains `ValidationSummary` with complex ViewModel target and option: `IncludeErrorsFromChildren` set to `true`.  Each ViewModel deserialization in that page triggers `dotvvm.validation.getValidationErrors` lot of times.

eg.:
![image](https://user-images.githubusercontent.com/656406/51681023-b074e980-1fe3-11e9-83f0-2108eea6695c.png)

ENgateman:
PR #729 Should fix this issue. Feel free to reopen it in case it doesn't.

PR #729, opened 2019-07-29
Improve performance of ValidationSummary

ENgateman:
Addresses #663 .

The method that `ValidationSummary` uses on client side (`getValidationErrors`) is spending way too much time in knockout-related functions. The reason for this, as far as I can see, is the fact that every error object is also a separate knockout observable so that `ValidationSummary` can simply render a `foreach` binding to iterate over them.

This PR makes a number of changes, some of which will definitely break things. However, it also improves the performance of `ValidationSummary` from entire seconds reported by @adamjez to a little over a hundred milliseconds (see the ValidationSummary Performace sample in this branch).

Major changes:
- Moves errors out of knockout observables
- Adds a custom knockout binding for `ValidationSummary` that hooks into the `validationErrorsChanged` event
- Changes the Validator binding to also hook into the `validationErrorsChanged` event


tomasherceg:
Merging, we'll have to test this on some larger project to make sure we haven't broken anything.

src/DotVVM.Framework/Resources/Scripts/DotVVM.Validation.ts
```diff
@@ -363,78 +399,102 @@ class DotvvmValidation {
...
-                }
+            return errors;
+        }
+        for (var propertyName in validationTarget) {
+            if (!validationTarget.hasOwnProperty(propertyName) || propertyName.indexOf("$") === 0) {
+                continue;
+            }
+
+            var property = validationTarget[propertyName];
+            if (!property || !ko.isObservable(property)) {
```
exyi: Again, redundant doublecheck.

src/DotVVM.Framework/Resources/Scripts/DotVVM.Validation.ts
```diff
@@ -445,12 +505,14 @@ class DotvvmValidation {
     public showValidationErrorsFromServer(args: DotvvmAfterPostBackEventArgs) {
         // resolve validation target
         var context = ko.contextFor(args.sender);
-        var validationTarget: KnockoutValidatedObservable<any>
-            = dotvvm.evaluator.evaluateOnViewModel(context, args.postbackOptions.additionalPostbackData.validationTargetPath);
-        if (!validationTarget) return;
+        var validationTarget = <KnockoutObservable<any>>dotvvm.evaluator.evaluateOnViewModel(
+            context,
+            args.postbackOptions.additionalPostbackData.validationTargetPath);
```
exyi: Could you please indent these argument more so it's clear that they are args of the `evaluateOnViewModel` method, this way it looks like a declaration of multiple variables.

PR #730, opened 2019-07-29
Don't allow service directives pointing to null to be part of DataContextStack

### LINKED ISSUE #713: NullReferenceException is thrown when a @service directive points to a non-existing type
It would be nice to inform user that the problem is missing file specified in directive.
screencapture-localhost-2109-2019-07-11-14_26_46

ENgateman:
Fixes #713.


tomasherceg:
Can you please add UI Test for the error page sample? After that, you can merge this in master - I think this is OK.

PR #732, opened 2019-08-01
Fix a SetValue exception when in multilevel mark-up control

ENgateman:
Fixes another bug reported by @JaskaRoman.

The problem is IMHO best described through the sample in this PR:

Let's have a `ComboBox` inside a markup control -`InnerWrapper` - that is wrapped in another markup control - `OuterWrapper`. The selected value of the `ComboBox` is bound using a value binding and two control bindings. If you try to set the `SelectedValue` of `InnerWrapper` using `SetValue` you get an exception because DotVVM doesn't currently allow bindings to propagate values in these cases.

That's what this PR is trying to fix by changing the resolver of `BindingUpdateDelegate`.


src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -85,6 +85,42 @@ static Expression GetDotvvmPropertyMember(Expression target, string name)
             );
         }
 
+        public static Expression SetMember(Expression node, Expression value)
```
exyi: This method does more than just setting a normal member and for example should not be used for setting a member when assignment operator is used.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -85,6 +85,42 @@ static Expression GetDotvvmPropertyMember(Expression target, string name)
...
+                && current is UnaryExpression unary)
+            {
+                current = unary.Operand;
+            }
+
+            if (current.NodeType == ExpressionType.Call
+                && current is MethodCallExpression call
+                && call.Method.DeclaringType == typeof(DotvvmBindableObject)
+                && call.Method.Name == nameof(DotvvmBindableObject.GetValue)
+                && call.Arguments.Count == 2
```
exyi: Could you please add an unit test for this case, so we at don't break it by adding an optional parameter to the GetValue method.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
ENgateman: Renamed it to `UpdateMember` and added a documentation comment.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
exyi: I thought of a unit test of this method, not a check that that overload still exists. You actually have a bug in here, the signature of `GetValue` is `GetValue(DotvvmProperty, bool)` not `GetValue(DotvvmProperty, object)`.

exyi: I have added few tests for that and fixed this problem.

ENgateman:
Merging as, since CI doesn't seem to work properly right now, I ran the tests locally and they passed.

PR #733, opened 2019-08-04
Added caching of resolved types in ReflectionUtils

quigamdev:
I have made performance profiling on one of our projects. The result was that ReflectionUtils.FindType was used many times. So I have added caching for this function.

### DIFF
--- a/src/DotVVM.Framework/Utils/ReflectionUtils.cs
+++ b/src/DotVVM.Framework/Utils/ReflectionUtils.cs
@@ -226,8 +226,18 @@ namespace DotVVM.Framework.Utils
             // convert
             return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
         }
-
+        private static ConcurrentDictionary<string, Type> cache_FindTypeHash = new ConcurrentDictionary<string, Type>();
+        private static ConcurrentDictionary<string, Type> cache_FindTypeHashIgnoreCase = new ConcurrentDictionary<string, Type>();
         public static Type FindType(string name, bool ignoreCase = false)
+        {
+            if (ignoreCase)
+            {
+                return cache_FindTypeHashIgnoreCase.GetOrAdd(name, a => FindTypeCore(a, true));
+            }
+            return cache_FindTypeHash.GetOrAdd(name, a => FindTypeCore(a, false));
+        }
+
+        private static Type FindTypeCore(string name, bool ignoreCase)
         {
             var stringComparison = ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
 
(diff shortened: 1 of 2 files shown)


src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -226,8 +226,18 @@ public static object ConvertValue(object value, Type type)
             // convert
             return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
         }
-
+        private static ConcurrentDictionary<string, Type> cache_FindTypeHash = new ConcurrentDictionary<string, Type>();
+        private static ConcurrentDictionary<string, Type> cache_FindTypeHashIgnoreCase = new ConcurrentDictionary<string, Type>();
```
tomasherceg: We can pass StringComparer.OrdinalIgnoreCase as a comparer to the dictionary - Roslyn uses this comparer for symbol equality, so it should be safe. http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/SymbolKey/SymbolKey.SymbolKeyComparer.cs,21

quigamdev: Fixed

tomasherceg:
I don't think it is an issue since there are thousands of other that will prevent unloading assemblies. All types cached in this dictionary are used on other places as well, e.g. in compiled views. 

We should consider refactoring the entire `ReflectionUtils` class - it doesn't contain only helpers, but there are quite large blocks of functionality that should probably be elsewhere. But I'd create a separate issue for this.

ENgateman:
Fair. If `ReflectionUtils` will get refactored at some point anyway, then I have no issue with the cache.

exyi:
On 6 August 2019 08:26:14 CEST, "Adam Štěpánek" <notifications@github.com> wrote:
>ENgateman commented on this pull request.
>
>I think caching the found types is a good idea, but I also think it
>shouldn't be implemented through a private static dictionary because
>a) I like static functions to be pure, but more importantly

Memoization is usually considered pure, it still does not have any observable side effects.

>b) If this function will be used with .NET Core 3's unloadable
>assemblies around, it will prohibit them from unloading since there
>will be a reference to them in a private static field that no-one can
>remove.

Would not a WeakReference<Type> solve that (for this function).


>IMHO it would be better to create a wrapper class that handles the
>caching, from which types can be removed.

In that case, it would certainly not be pure and it would not solve that problem. Only a very advanced user could (with enough motivation) some it for itself.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

exyi:
@quigamdev are you sure the problem is actually the `FindType` function rather than too frequent compilation of something in your project? From where is it called? Maybe caching some binding that is created over and over again would help much more.

ENgateman:
> Memoization is usually considered pure, it still does not have any observable side effects.

In this case, it doesn't have any observable side effects *yet*. It will once unloadable `AssemblyLoadContext`s will be a thing.

> Would not a WeakReference<Type> solve that (for this function).

Yes, I think it would. The unloadability design doc seems to agree with us:

_"Users can unload the AssemblyLoadContext when there are no outside hard references to the types and instance of types from the assemblies loaded into the context and the assemblies themselves."_

> In that case, it would certainly not be pure and it would not solve that problem. Only a very advanced user could (with enough motivation) some it for itself.

Fair, but there would be at least something the user could do about it.

src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -238,14 +249,16 @@ public static Type FindType(string name, bool ignoreCase = false)
             var split = name.Split(',');
             name = split[0];
 
-            var assemblies = ReflectionUtils.GetAllAssemblies();
+            var assemblies = ReflectionUtils.GetAllAssemblies().ToList();
```
exyi: Note that this `ToList` not very likely to improve performance, since `assemblies` are only iterated once.

PR #734, opened 2019-08-07
Support local variables inside ValidationError expressions

### LINKED ISSUE #724: ValidationErrorFactory.GetPathFromExpression path array access throws on dynamic index values
`var index = Items.IndexOf(item);`
`ValidationErrorFactory.GetPathFromExpression(ViewModel.Context.Configuration, (Expression<Func<FooViewModel, double?>>)(vm => vm.Items[index].Bar));`

Previous example throws this exception:
> **InvalidOperationException**:
> Can not translate field 'Int32 index' to Javascript'

Constant values work just fine.

`ValidationErrorFactory.GetPathFromExpression(
ViewModel.Context.Configuration, (Expression<Func<FooViewModel, double?>>)(vm => vm.Items[0].Bar));`

ENgateman:
Closes #724.

If merged, references to local variables inside expressions passed to the `ValidationErrorFactory` methods will be regarded as literals.


exyi:
Unfortunately, these literals don't work with the cache that is built into the `ValidationErrorFactory`. The problem is that C# compiler creates a display class that holds the local variables and then uses `Expression.Constant(instanceOfThatClass)`. That class does not have `Equals` and `GetHashCode` and is mutable, so we can't store the instance in the cache. (see [sharplab.io with a sample](https://sharplab.io/#v2:EYLgHgbALAPgAgJgIwFgBQcAMACOSB0AMgJYB2AjvgKJgAOATgKYDOzxA9qcwNzpwDMuBNgDC2AN7ps03IJoMWbTgB44AVmVkALgD4d2ALIAKbdmIBKCVJk24AdmxHLAXn3FeaGwF90XoA==))

I think we could propagate the constants in the expression before we look into the cache. That would also limit the scope of this change to the validation path translator instead of affecting the entire JS translator.

PR #737, opened 2019-08-14
StyleBuilder.SetControlProperty improvements

exyi:
* It should now work with property of all types (single control, control
  collection or a template)
* Added overloads to set html control or a literal
* Added support for appending two templates

It supports handling of
* `StyleOverrideOptions.Ignore` - does not set the property if it's already asigned
* `StyleOverrideOptions.Overwrite` - overwrites anything is assigned
* `StyleOverrideOptions.Append` - tries to append to a list (to the end, does not work with single control properties)


exyi:
I have fixed one more bug - it was impossible to set one property multiple times in one `Register` call.

cc @martindybal this PR should solve your issue.

PR #747, opened 2019-08-30
Freezable config

exyi:
This PR makes DotvvmConfiguration "frozen" after DotVVM is initialized. Any (well, almost) attempt to modify the configuration after the startup will fail with an exception.

This is a non-trivial breaking change, but most of the usages should be fine. If somebody is relying on runtime config modifications, they will get an exception, but their code would (very likely) be broken anyway.

We should also make sure, if the configuration deserializes properly - I have changed quite a bit of `{ get; private set; }` into ` => privateField` and I'm not sure how that affects the serializer.


src/DotVVM.Framework/Compilation/Styles/StyleRepository.cs
```diff
@@ -53,5 +65,17 @@ public StyleBuilder<HtmlGenericControl> Register(string tagName, Func<StyleMatch
                 return Register<HtmlGenericControl>(m => (string)m.Control.ConstructorParameters[0] == tagName, false);
             }
         }
+
+        private bool isFrozen = false;
+        private void ThrowIfFrozen()
+        {
+            if (isFrozen)
+                throw new InvalidOperationException("The StyleRepository is frozen and can be no longer modified.");
```
ENgateman: The exception message is (apart from the type name) the same for all freezable types, can we unify it as a single format string?

Also, the "The" at the beginning seems a little awkward to me, "This" would IMHO be better.

ENgateman: Note: `DotvvmConfiguration` itself has a different, older message.

src/DotVVM.Framework/Configuration/HtmlAttributeTransformConfiguration.cs
```diff
@@ -27,7 +37,10 @@ public HtmlAttributeTransformConfiguration()
 
         public IHtmlAttributeTransformer GetInstance()
         {
-            return instance.Value;
+            if (isFrozen)
+                return instance.Value;
+            else
+                throw new NotSupportedException("Can not use HtmlAttributeTransformConfiguration since it can still be modified");
```
ENgateman: "This HtmlAttributeTransformConfiguration must be frozen before the IHtmlAttributeTransformer instance can be returned." would IMHO be a better message.

src/DotVVM.Framework/Routing/RouteBase.cs
```diff
@@ -186,5 +192,21 @@ public static void AddOrUpdateParameterCollection(IDictionary<string, object> ta
...
+        }
+        public void Freeze()
+        {
+            this.isFrozen = true;
+            FreezableDictionary.Freeze(ref this._defaultValues);
+
+            Freeze2();
+        }
+
+        protected abstract void Freeze2();
```
ENgateman: Come on, `Freeze2`? What kind of a name is that. Is there a problem with making `Freeze` virtual?

exyi: The problem is that it would allow you to override the `Freeze` method in a derived class, which would allow you to suppress the freeze. I don't care about the name of `Freeze2`, maybe we could also name it `Freeze_IHATEINHERITANCE`

src/DotVVM.Framework/Compilation/Styles/StyleRepository.cs
exyi: > The exception message is (apart from the type name) the same for all freezable types, can we unify it as a single format string?

Yea, we can. But there is the obvious problem where to put that shared function :)

> Note: DotvvmConfiguration itself has a different, older message.

I know, it on purpose - it also explain when does the configuration get frozen. The message for other types is more general as they can be also use in different context.

src/DotVVM.Framework/Routing/RouteBase.cs
quigamdev: @exyi  Please, name it OnFrozen(). We can at least use .net conventions. When it's called OnFrozen you know that method is invoked after configuration is frozen. Freeze2 is nothing-saying.

quigamdev: Or maybe the better name is even OnFreezing()

tomasherceg: `OnFreezing` or `FreezeCore` looks better than `Freeze2`.

tomasherceg:
LGTM

src/DotVVM.Framework/Routing/RouteBase.cs
exyi: `FreezeCore` is a bad name, because it's not a core of freeze.

`OnFreezing` is bad name, because it's not an event you can optionally handle - it's a method you must implement and make sure that all properties are frozen.

I think that `Freeze2` will trigger a potential implementer of a derived class, which increases the probability it will get implemented correctly.

Also, if the implementer would want to allow inheritance from his class, he'd have to create another `Freeze` method for them. That would be named `Freeze3`, which is much better than `OnOningFreezing` or `TheRealCoreOfFreeze` :)

quigamdev: Still, I don't like it. But now I understand your POV.

Issue #745, opened 2019-08-23
Collect runtime warnings from DotVVM


exyi:
We could add a mechanism to collect runtime warnings from DotVVM and display them in the browser console (so the dev can actually read it). On ASP.NET Core, these warnings should be propagated to the `ILogger`.

The warnings could include HTML warnings from HtmlWriter (like `table` can not contain `div`)

Then, we could add optional "sanitizers" to help debug common problems with postbacks:
- add virtual command binding to all controls in the page
- execute postback and find all the command bindings
- warn if some command is missing or some is new

Or with control ids (leading to problems `Postback.Update` or commands in markup controls):
- add a virtual `Postback.Update` to all controls (force a generated id)
- execute postback
- warn if some control is missing or some is new

Such warning could be pretty expensive, so it should be done only on demand, but it could help find glitches in custom controls.

exyi:
We should also have a sanitizer that checks during client-side serialization, that all objects have `...$options` and `$type`. It is quite easy to miss that when objects are created manually in JS.

PR #750, opened 2019-09-01
API for runtime warnings

exyi:
RuntimeWarningCollector is a service, where a new warning will can be
added using the `Warn` method. Warning is an instance of
DotvvmRuntimeWarning, contains a message and may contain a reference to
an exception and a bindable object.

By default, warnings are collected in Debug mode. They are displayed in
browser JS console and forwarded to Asp.Net Core logging.

IDotvvmWarningSink is extensibility point for people that want to
collect these warnings on their own.

As a demo, HtmlWriter reports warnings when elements that should be
always self-closing contain some content, as the browser will interpret
the content as sibling.

related to #745, but does not close it since this PR introduces only the infrastructure.

Also note that I had to move rendering of resources in postback after rendering of `PostBack.Update` controls. Since we support resource registration in the `Render` phase, this would need to be done anyway. I have altered `IHtmlWriter` and `IViewModelSerializer`, but these interfaces are so bound to the implementations in framework, that I don't believe somebody has implemented them...

The public API looks like this:

```C#
namespace DotVVM.Framework.Runtime
{
    public interface IDotvvmWarningSink
    {
        void RuntimeWarning(DotvvmRuntimeWarning warning);
    }
    public class DotvvmRuntimeWarning
    {
        public DotvvmRuntimeWarning(string message, Exception relatedException = null, DotvvmBindableObject relatedControl = null);

        public string Message { get; }
        public Exception RelatedException { get; }
        public DotvvmBindableObject RelatedControl { get; }

        public override string ToString();
    }

    public class RuntimeWarningCollector
    {
        public bool Enabled { get; }

        public RuntimeWarningCollector(IEnumerable<IDotvvmWarningSink> sinks, DotvvmConfiguration config);

        public void Warn(DotvvmRuntimeWarning warning);


        public List<DotvvmRuntimeWarning> GetWarnings() => warnings.ToList();
    }
}
```

### DIFF
--- a/src/DotVVM.Framework/Controls/HtmlWriter.cs
+++ b/src/DotVVM.Framework/Controls/HtmlWriter.cs
@@ -204,7 +204,7 @@ namespace DotVVM.Framework.Controls
             writer.Write("/>");
 
             if (this.enableWarnings && !IsSelfClosing(name))
-                Warn($"Element {name} is not self closing but is rendered as so. It may interpreted as a start tag without the end tag by the browsers.");
+                Warn($"Element {name} is not self-closing but is rendered as so. It may be interpreted as a start tag without an end tag by the browsers.");
         }
 
         private Dictionary<string, string> attributeMergeTable = new Dictionary<string, string>(23);
@@ -368,7 +368,7 @@ namespace DotVVM.Framework.Controls
                 writer.Write(">");
 
                 if (this.enableWarnings && IsSelfClosing(tag))
-                    Warn($"Element {tag} is self closing but contains content. The browser may interpret the start tag as self-closing and put the 'content' into it's parent.");
+                    Warn($"Element {tag} is self-closing but contains content. The browser may interpret the start tag as self-closing and put the 'content' into its parent.");
             }
             else
             {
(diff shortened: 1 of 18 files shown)


src/DotVVM.Framework/Controls/HtmlWriter.cs
```diff
@@ -191,6 +202,9 @@ public void RenderSelfClosingTag(string name)
         {
             RenderBeginTagCore(name);
             writer.Write("/>");
+
+            if (this.enableWarnings && !IsSelfClosing(name))
+                Warn($"Element {name} is not self closing but is rendered as so. It may interpreted as a start tag without the end tag by the browsers.");
```
ENgateman: You're missing the "be" in "be interpreted".

If start tag has indefinite article, end tag should most likely have it as well: "an end tag".

Also, the tiniest of nitpicks: "self-closing".

src/DotVVM.Framework/Controls/HtmlWriter.cs
```diff
@@ -352,6 +366,9 @@ public void RenderEndTag()
                 writer.Write("</");
                 writer.Write(tag);
                 writer.Write(">");
+
+                if (this.enableWarnings && IsSelfClosing(tag))
+                    Warn($"Element {tag} is self closing but contains content. The browser may interpret the start tag as self-closing and put the 'content' into it's parent.");
```
ENgateman: Come on, you even have it here (still looking at "self-closing").

~"it's parent"~ -> "its parent".

exyi: Thanks, fixed that.

src/DotVVM.Framework/Controls/HtmlWriter.cs
exyi: Thanks, fixed that.

PR #751, opened 2019-09-01
Allowed bindings in attributes without quotes

exyi:
Binding can be now written without quotes in attributes:

```
MyProperty={value: OtherProperty}
```

The quotes are basically redundant in the syntax, as the binding is
already "quoted" by the curly braces. IMHO, it makes more sense to write
them without quotes as it's just noise. And React (JSX) also has them without
quotes :)


tomasherceg:
@Mylan719 How difficult will this be to support in the VS Extension?

exyi:
@tomasherceg This should work by default, as extension uses the same tokenizer and parser.

tomasherceg:
@exyi I know, but I expect many issues with intercepting the IntelliSense and all the black magic we are doing with artifacts.

quigamdev:
@tomasherceg We can publish an internal release with this PR and try it.

exyi:
@quigamdev or @Mylan719 you wanted to try this? Could you please have a look at it?

PR #752, opened 2019-09-06
Allow rename of a IRowEditOptions.PrimaryKeyPropertyName property through JsonProperty

ENgateman:
The behaviour so far was that inline editing of a GridView broke when a `JsonProperty` attribute renamed the primary key of the `GridViewDataSet`.


ENgateman:
`[Bind(Name = ...)]` should work fine. `DefaultPropertySerialization` seems to handle it as well.

How would say a `PrimaryKeyBinding` work on `GridViewDataSet`? Is there a way to set such a binding from inside the ViewModel like you set the name in `PrimaryKeyPropertyName`?

Anyway, I'm merging this.

exyi:
> How would say a |PrimaryKeyBinding| work on |GridViewDataSet|? Is there
> a way to set such a binding from inside the ViewModel like you set the
> name in |PrimaryKeyPropertyName|?

Ah, it's on the GridViewDataSet, I thought it's on the grid :/ That way,
it could be Linq.Expression instead of a value binding, but that would
be harder to change and implement.

PR #753, opened 2019-09-06
Support presenter factories on route groups

### LINKED ISSUE #661: Route groups should allow to specify presenter factory
It is annoying to add `LocalizablePresenter.BasedOnParameter("Lang")` to 150 routes when adding multiple languages in the project - the route groups should support it.

ENgateman:
Closes #661.

Please note that I had to change the signature of `DotvvmRouteTable.Add(string routeName, string url, Func<IServiceProvider, IDotvvmPresenter> presenterFactory, object defaultValues = null)` to `DotvvmRouteTable.Add(string routeName, string url, Func<IServiceProvider, IDotvvmPresenter> presenterFactory = null, object defaultValues = null)` (presenterFactory is optional now) to make it work.

### DIFF
--- a/src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
+++ b/src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
@@ -2,11 +2,8 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using DotVVM.Framework.Configuration;
 using DotVVM.Framework.Hosting;
-using DotVVM.Framework.Runtime;
-using DotVVM.Framework.Security;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace DotVVM.Framework.Routing
@@ -17,22 +14,12 @@ namespace DotVVM.Framework.Routing
     public class DotvvmRouteTable : IEnumerable<RouteBase>
     {
         private readonly DotvvmConfiguration configuration;
-
-        private List<KeyValuePair<string, RouteBase>> list = new List<KeyValuePair<string, RouteBase>>();
-
-        /// <summary>
-        /// Dictionary for faster checking of duplicate entries when adding.
-        /// </summary>
-        private Dictionary<string, RouteBase> dictionary = new Dictionary<string, RouteBase>(StringComparer.OrdinalIgnoreCase);
-
-        /// <summary>
-        /// Dictionary for groups of RouteTables.
-        /// </summary>
-        private Dictionary<string, DotvvmRouteTable> routeTableGroups = new Dictionary<string, DotvvmRouteTable>();
-
-        /// <summary>
-        /// Contains information about the group of this RouteTable.
-        /// </summary>
+        private List<KeyValuePair<string, RouteBase>> list
+            = new List<KeyValuePair<string, RouteBase>>();
+        private Dictionary<string, RouteBase> dictionary
+            = new Dictionary<string, RouteBase>(StringComparer.OrdinalIgnoreCase);
+        private Dictionary<string, DotvvmRouteTable> routeTableGroups
+            = new Dictionary<string, DotvvmRouteTable>();
         private RouteTableGroup group = null;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -17,22 +14,12 @@ namespace DotVVM.Framework.Routing
     public class DotvvmRouteTable : IEnumerable<RouteBase>
     {
         private readonly DotvvmConfiguration configuration;
-
-        private List<KeyValuePair<string, RouteBase>> list = new List<KeyValuePair<string, RouteBase>>();
-
-        /// <summary>
-        /// Dictionary for faster checking of duplicate entries when adding.
```
exyi: Come on, why have you reformatted the code and removed some comments (that actually explained something)? If there is anything wrong with them, it would be better to fix that. This comment specifically is not redundant.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -77,7 +69,13 @@ public void AddGroup(string groupName, string urlPrefix, string virtualPathPrefi
...
 
             var newGroup = new DotvvmRouteTable(configuration);
-            newGroup.group = new RouteTableGroup(groupName, group?.RouteNamePrefix + groupName + "_", urlPrefix, virtualPathPrefix, Add);
+            newGroup.group = new RouteTableGroup(
+                groupName: groupName,
+                routeNamePrefix: group?.RouteNamePrefix + groupName + "_",
+                urlPrefix: urlPrefix,
+                virtualPathPrefix: virtualPathPrefix,
+                addToParentRouteTable: Add,
+                presenterFactory: presenterFactory);
```
exyi: Nit: you can specify only some of the argument names. So you can have `addToParentRouteTable: Add` as it has some information and then only `presenterFactory` as repeating the name twice won't help anyone.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -87,7 +85,7 @@ public void AddGroup(string groupName, string urlPrefix, string virtualPathPrefi
         /// Creates the default presenter factory.
         /// </summary>
         public IDotvvmPresenter GetDefaultPresenter(IServiceProvider provider) =>
-            provider.GetRequiredService<IDotvvmPresenter>();
+            group?.PresenterFactory(provider) ?? provider.GetRequiredService<IDotvvmPresenter>();
```
exyi: In this case you accept `null` returned from `PresenterFactory`, when the presenter is specified for a single route, null result is not accepted. I think this behavior should be unified.

src/DotVVM.Framework/Routing/RouteTableGroup.cs
```diff
@@ -1,25 +1,37 @@
...
+        public Action<string, RouteBase> AddToParentRouteTable { get; }
 
-        public string GroupName { get; private set; }
-        public string RouteNamePrefix { get; private set; }
-        public string UrlPrefix { get; private set; }
-        public string VirtualPathPrefix { get; private set; }
+        public string GroupName { get; }
 
-        public RouteTableGroup(string groupName, string routeNamePrefix, string urlPrefix, string virtualPathPrefix, Action<string, RouteBase> addToParentRouteTable)
+        public string RouteNamePrefix { get; }
```
exyi: Are you addicted to reformatting to existing code? Removing unused private setters and adding few newlines definitely won't help readability, it just creates more merge conflicts and drops history. Everyone would like these irrelevant details to be a bit different and it would not help anyone if I'd remove these newlines between properties and add one more newline before `string groupName` in my next commit.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
ENgateman: I don't see the point of documentation for private fields, especially when it's relatively obvious what they're for. I agree, however, that this comment is less redundant than the rest, so I put it back.

src/DotVVM.Framework/Routing/RouteTableGroup.cs
ENgateman: Maybe a bit.

It does help readability. Inconsistent formatting is distracting and the formatting in this class was inconsistent even with itself. I could either add the property and make the mess worse or reformat it. If you really want me to revert this change and stop reformatting everything I touch; fine, just say the word. This is not gonna fix itself, though.

exyi: Yes, I want you to stop reformatting the existing code, as I have explained before (it's just harder to read the diff and makes the history less useful). I don't know what was inconsistent with this class, it was just a few properties and looked the same. Maybe they had the redundant `private set;`, but who cares about that (well, in some cases Newtonsoft.Json does, but this is not the case). And (code style) consistency is overrated, you'll never achieve it even if you would, this is not the only project you'll meet in your career. I'd better figure out how to read code that different spacing than what you'd write.

ENgateman: Fair enough. I reverted the formatting changes.

If you have no other comments, please approve the PR.

PR #756, opened 2019-09-10
Add support for InvalidCssClass on GridViewTextColumn

### LINKED ISSUE #493: Validator.InvalidCssClass not respected on GridViewTextColumn
Client-side validation works for inline editing, but only shows the specified InvalidCssClass for template columns.

ENgateman:
Closes #493.

Copies the value of `Validator.InvalidCssClass` to the inner `TextBox` of `GridViewTextColumn`.


exyi:
This was an easy fix, but it actually only makes the problem worse :/ Let me explain. The problem is that you can set any property on any bindable object and it only sometimes actually does something. In this case, `Validator.Value` and friends work on all DotvvmControls (unless other strange things happen). And that's because `Validator.Value` is an Active Property and all active properties are supposed to work on all DotvvmControls (unless other strange things happen). This rule is fairly simple to understand, maybe it's not the first thing you'll explain to a "dotvvm noob", but as a control/framework/"advanced" developers, this is the thing we keep in mind. Any exception to this rule makes things more complicated and our jobs get harder. Even today, there are many exceptions... I'm not aware of most of them and unfortunately probably nobody is, which makes it basically impossible to tell how will the code behave just by looking at it. So, why don't we simply run the code to see how it works? Well, you can do that (kindof...) when you are developing a standalone web page with DotVVM, since there aren't many factors affecting the behavior that may differ from what you are testing. Unfortunately, there is vary many of these factors in DotVVM when you are developing something generic and we can't run the code in all these cases to see how it behaves. This is not about testing, this is about understanding what will my change/feature affect, so we can't even automate it in any way.

This change may solve a single problem, but makes the system a little bit more complicated. Now we have controls where all validators work and the text column where only `InvalidCssClass` works. BP text column will have different rules, because somebody else will be programming that. We'll add few more similar helpers, some properties will be applied to the table cell, some to the textbox and some to the header. Now happy programming. And god bless you if you'd try to write some generic logic on top of that, regardless if it would be something similar to VS Extension or Dynamic Data.

By the way, the `Validator...` properties are inherited (except for `Validator.Value`), which is also sometimes funny/glitchy:

This works:
```
<dot:GridViewTextColumn Validator.Value="{...}" Validator.InvalidCssClass="error" />
```


This does not:
```
<dot:GridViewTextColumn Validator.Value="{...}" Validator.SetToolTipText />
```

But this does!
```
<dot:GridView Validator.SetToolTipText />
    <dot:GridViewTextColumn Validator.Value="{...}" />
</dot:GridView>
```

Why doesn't it get inherited from `GridViewTextColumn` to the `TextBox`? Because, GridViewTextColumn is not in the control tree, so nobody cares about it's properties. Maybe this is the thing that's wrong.

Anyway, I agree that validation and inline-edit is a real problem and I don't have better a solution, but we should come up with a systematic solution to this problem family. These things should be considered:
* Validators on `GridViewTextColumn`
    - and generalized Validators on all `GridViewColumn`s
    - and any Active Properties on `GridViewColumn`s (for example `Events.Click`)
* `Postback.Handler`s on `OnChanged` command
    - or `Postback.Handler`s on any "indirect" command
    - concurrency settings and render mode are probably the same story
* `FormatString`, `Type`, `SelectAllOnFocus` and other useful properties can't be set
* `DataContext` is ignored (I think) on some non-controls

... And more

tomasherceg:
I agree with @exyi - this change is unsystemmatic and creates more confusion.

* If the user needs to apply Validator properties on a row, we have `RowDecorator` for this purpose. 
* If the user needs to apply the Validator properties inside a particular cell, there is `GridViewTemplateColumn` - I guess that in such cases the user will need more customizations anyway.
* Currently, we are not able to set this up on the `<td>` element itself which can be quite limiting.

What I think `GridViewTextColumn` should be able to do is to apply `Validator.Value` automatically to the `TextBox` generated in the edit mode, as there is no other way to set it up except for using `GridViewTemplateColumn`. But there might be cases when you don't want to apply the CSS class on the control itself, but on the table cell (e.g. for Bootstrap 3 - the `has-error` class doesn't work on `input` elements - it has to be set on their parent elements). So this behavior should be configurable somehow and we should discuss how we'll approach this.

But the other validator properties which only configure how the validators look like, should be set on the `GridView` level, or using the `RowDecorator`. I don't think that it makes much sense to configure them on particular columns as in most applications the validation should look consistent. And if not, there is the template column.

tomasherceg:
We've talked about this PR on the call and decided on adding a property that would cover the two common ways of doing validation:
1) Apply `Validator.Value` on the control itself and let the validation parameters be specified from the outside - RowDecorators, GridView or its ancestors.
2) Render `Validator` control after the control itself (and also let the validation parameters be specified from the outside - RowDecorators, GridView or its ancestors).

Originally I thought about creating an enum with three options (`None`, `ValidateControl` or `RenderStandaloneValidator`), but it might be useful to combine both options together (make the texbox red and render a validation error below), so maybe we should add two `bool` properties instead.
* `ValidateControl`
* `RenderStandaloneValidator`

I am not sure about the names, especially the first one. Do you have better ideas on how to name them?

Note: We shouldn't add these properties to the template column because the user is responsible to handle the validation manually, so there might be a need for a new base class that would be common for `GridViewCheckBoxColumn` and `GridViewTextBoxColumn` - I believe that we have a similar class in Business Pack so we can take inspiration from there.

ENgateman:
If we can, I think we should use yes/no questions for the names:
- `IsControlValidated`
- `IsValidatorStandalone`

tomasherceg:
The problem with `Is` convention is that we don't stick to it almost anywhere in the framework - we have `Enabled`, `Visible`, `AllowInlineInsert` etc. That's why it seems a better option to have `ValidateControl` than `IsControlValidated` and `Render...` or  `AddStandaloneValidator`.
But let's implement this and we can rename the properties if we find better names.

ENgateman:
I think I implemented the new properties sufficiently. What do you think about it?

Concerning the names, there's IMHO a crucial difference between e.g. `Enabled` and `ValidateControl`: `ValidateControl` looks like it ought to be a method to me whereas `Enabled` looks like it's a `bool` or `bool?` field or property.

src/DotVVM.Framework/Controls/GridViewValueColumn.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// A base class for GridView columns that present a value by wrapping a control.
+    /// </summary>
+    public abstract class GridViewValueColumn : GridViewColumn
+    {
+        /// <summary>
+        /// Gets or sets a binding which retrieves the value to display from the current data item.
+        /// </summary>
+        [MarkupOptions(AllowHardCodedValue = false, Required = true)]
+        public IValueBinding ValueBinding
```
exyi: I don't think it's a great idea to loose the type constraint of `ValueBinding` property.

src/DotVVM.Framework/Controls/GridViewValueColumn.cs
```diff
@@ -0,0 +1,82 @@
...
+using DotVVM.Framework.Binding;
+using DotVVM.Framework.Binding.Expressions;
+using DotVVM.Framework.Hosting;
+
+namespace DotVVM.Framework.Controls
+{
+    /// <summary>
+    /// A base class for GridView columns that present a value by wrapping a control.
+    /// </summary>
+    public abstract class GridViewValueColumn : GridViewColumn
```
exyi: I'm not a fan of making the inheritance hierarchy deeper. In this case, it may lead to quite a bit of problems:
* What if I want a column that renders a control but does have the one distinctive "Value". For example combobox could be validated according to the `SelectedValue`, but the property probably should not be called `Value` for consistency with the original control.
* I may also want to have column with Value where validation does not make much sense. (example might be any non-trivial read-only component)
* I may already want to have a custom base class. "value column" may be one of many traits of that control, not the main super-class. 
* `ValueBinding` property does not have type specified (which is a problem for `GridViewCheckBoxColumn`)

I know they have something similar in BP, but we don't have to repeat every mistake here. Creating more classes named the same also won't probably help.

src/DotVVM.Framework/Controls/GridViewValueColumn.cs
```diff
@@ -0,0 +1,82 @@
...
+        public bool IsControlValidated
+        {
+            get { return (bool)GetValue(IsControlValidatedProperty); }
+            set { SetValue(IsControlValidatedProperty, value); }
+        }
+        public static readonly DotvvmProperty IsControlValidatedProperty
+            = DotvvmProperty.Register<bool, GridViewValueColumn>(c => c.IsControlValidated, false);
+
+        [MarkupOptions(AllowBinding = false)]
+        public bool IsValidatorStandalone
```
exyi: What if this is `true` while `IsControlValidated` is `false`? I guess there should be four valid states:
1. validation is diabled
2. The control has a validator attached.
3. There is the validator next to that control
4. Both of the above: the control has validator attached and message is displayed next to it.

This patch only seems to implement the first three, while the last one seems quite practical to me.

ENgateman:
I removed the base class and replaced the two properties with an enum. @tomasherceg @exyi can you take another look and merge this PR if you're content with it, please?

PR #757, opened 2019-09-18
Static command metadata

quigamdev:
This PR wraps all command responses created for static command. It adds an ability to add extra data for DotVVM into the response. 

As a first step, this PR also adds an ability to make a redirect from static command services where IDotvvmReqeustContext can be injected. On the context, you can easily invoke RedirectToUrl() or RedirectToRoute().


src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -31,6 +31,12 @@ interface IDotvvmViewModelInfo {
 interface IDotvvmViewModels {
     [name: string]: IDotvvmViewModelInfo
 }
+interface IDotvvmStaticCommandResponse {
+    action: string;
+    url: string;
+    result: any;
```
exyi: The types are wrong, as all three properties are optional.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -391,7 +397,12 @@ class DotVVM {
             this.postJSON(<string>this.viewModels[viewModelName].url, "POST", ko.toJSON(data), response => {
                 try {
                     this.isViewModelUpdating = true;
-                    const result = JSON.parse(response.responseText);
+                    const responseJson = <IDotvvmStaticCommandResponse>JSON.parse(response.responseText);
+                    const result = responseJson.result;
+                    if (responseJson.action === "redirect") {
+                        // redirect
+                        this.handleRedirect(responseJson, viewModelName)
+                    }
```
exyi: It's unclear whether the method should continue execution in case of a redirect. When the redirect happens on a "normal page", it will abort the execution, since the page uploads (I think, haven't tested). If we are performing SPA redirect, the `this.handleRedirect` method returns and the flow continues.

* It is inconsistent with `postbackCore` which interrupts the method (by returning and resolving the promise).
* It will try to finish the rest of the static command, which may starts another request racing with the navigation.

exyi: I think it should not continue with the method execution and it should `reject` (although normal postbackCore `resolve`s :/ )

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
quigamdev: @exyi What are the correct ones? If they are optional, do you think they should not be declared?

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
quigamdev: Why we should reject it? If we want to be consistent, I would much more rather `resolve` it as postbackCore does.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
exyi: They should be declared as optional: `action?: string`. Or better, it seems that only these cases are valid `{ result: any }`, `{ action: "redirect"; url: string }`, so we can use:

```typescript
type DotvvmStaticCommandResponse =
      { result: any }
    | { action: "redirect"; url: string }
```

There also isn't (AFAIK) a reson to use `interface`, and even if there would, interfaces in typescript are usually not prefixed with `I`.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
exyi: Well, yes, but there will be (probably) nothing happening after postbackCore. When we `resolve` here, the static command execution will continue, which may cause race conditions and strange behavior in SPA pages.

quigamdev: Ok, I will add the return statement.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
quigamdev: I defined the type as 
```typescript 
type DotvvmStaticCommandResponse = { result?: any; action?: "redirect"; url?: string };
```
The problem is once you access the result property the compiler does not let you access the action property.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
exyi: Only the return statement is also not the best since it makes the promise run forever.

exyi: Could you please reject it by calling `error`.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
exyi: I have redefined that as a union.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
quigamdev: Do you mean `errorCallback({ xhr: response, error: null });`?

quigamdev: Oh, I see. You have already changed it.

exyi: I have only changed the type. But I have changed that now - the promise gets rejected when redirect should happen.

quigamdev:
All tests have passed. So I'm merging this PR.

PR #768, opened 2019-10-02
Fix failed XHR request failing not being reported to the caller

mrnustik:
DotVVM did not respond to correctly to failed XHR requests when the requests were made offline. This caused `dotvvm.events.error` subscribers not being notififed of those errors.


tomasherceg:
LGTM. When the tests pass, we can merge and publish an internal release.

src/DotVVM.Samples.Common/Views/ComplexSamples/SPAErrorReporting/site.dotmaster
```diff
@@ -0,0 +1,34 @@
...
+    <dot:SpaContentPlaceHolder ID="Container" />
+
+    <div>Number of logged errors: <span ID="numberOfErrors">0</span></div>
+
+    <dot:InlineScript Dependencies="dotvvm">
+        var numberOfErrors = 0;
+        dotvvm.events.error.subscribe(
+            function (error) {
+                var element = document.getElementById('numberOfErrors');
+                element.innerText = ++numberOfErrors;
```
exyi: I don't understand how this test works. When does the error occur? Does the UI test anything?

mrnustik: The issue that is fixed by this PR caused the dotvvm.events.error not be called when app was offline and the SPA navigation (or any other server connection). This UI only demonstrates that the event is called properly when you turn offline mode in the development tools of your browser. This page was supposed to be used as part of UI test, but I did not find a way to make the selenium driver change the page to think its offline. So I keeped it for manual testing. 

I can remove it if you want.

exyi: No, that's fine. Could you please just add this explanation to the test, so that we can actually test it sometime in the future?

mrnustik: Done.

PR #770, opened 2019-10-03
Error event reported correctly in lazy CSRF mode

tomasherceg:
The PR #768 fixed the error event to be triggered correctly when the application is offline, however it didn't worked in lazy CSRF mode - when the CSRF token couldn't be obtained, it crashed without invoking such event.

This PR should fix this in lazy CSRF mode and cover this functionality with UI test making sure that all kinds of client-server communication will either succeed or raise the error event:
* command
* staticCommand
* SPA navigation

We should consider renaming the UI test to OfflineMode or something like this as it is not related only to SPA navigation.


src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -543,8 +554,16 @@ class DotVVM {
     public postbackCore(options: PostbackOptions, path: string[], command: string, controlUniqueId: string, context: any, commandArgs?: any[]) {
         return new Promise<() => Promise<DotvvmAfterPostBackEventArgs>>(async (resolve, reject) => {
             const viewModelName = options.viewModelName!;
-            await this.fetchCsrfToken(viewModelName)
             const viewModel = this.viewModels[viewModelName].viewModel;
+            
+            var csrfToken;
```
exyi: This variable seems to be unused.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -386,12 +386,23 @@ class DotVVM {
...
+            try {
+                csrfToken = await this.fetchCsrfToken(viewModelName);
+            }
+            catch (err) {
+                this.events.error.trigger(new DotvvmErrorEventArgs(sender, this.viewModels[viewModelName].viewModel, viewModelName, null, null));
+                console.warn(`CSRF token fetch failed.`);
+                errorCallback({ error: err });
+                return;
+            }
```
exyi: Might have been cleaner to use `await this.fetchCsrfToken(viewModelName).catch(err => ...)`

tomasherceg: I didn't knew about that, but still - I need to use `return` in the catch clause so it doesn't help much here.

PR #771, opened 2019-10-03
Public API migrated to non-nullable references

exyi:
Most of DotVVM public API should be now annotated with C# 8 non-nullable references.

There is a few notable changes:
* DotVVM.Framework targets also `netcoreapp3.0`, which is strange, but it does not do the typechecking properly for us. I don't know if there is a `netstandard` distribution of .NET with the nullability annotations which would the use-case better.
* There is the `NullableAttributes.cs` file copy-pasted from here: `https://source.dot.net/#System.Private.CoreLib/shared/System/Diagnostics/CodeAnalysis/NullableAttributes.cs,d813075228759f7c` which makes it possible to use the annotations even on older target platforms.
* Some types not intended to be created by our API users now have constructors. This may theoretically break someone, but I have only made the change when it does make sense.
* GridViewDataSet requires the `T` to be nullable (`class` constraint).
* In some extreme cases DotVVM now checks for nulls earlier, which may theoretically break someone. For example you can not invoke lifetime events on controls without the `Internal.RequestContextProperty` set.

Overall, most of the DotvvmProperties must be nullable, since the bindings may return null whenever they want to. Most `IDotvvmRequestContext` properties must be also nullable, because they are filled during the request processing and not when the context is created. Maybe this will be a bit annoying, but it's late to do something about it :/, so everyone will have to use `context.Parameter!` every single time...

Migration of entire DotVVM will take a bit more effort, this only enables the null checks through `#nullable enable` for specific files.


exyi:
@tomasherceg Could you please review the API? The PR affects pretty much everything, so we don't really want to wait too long with merge.

ENgateman:
I had a quick look around the changes and while I have no feedback regarding the nullability, I edited a few documentation comments and created PR #782.

PR #774, opened 2019-10-08
Signed properties can now contain objects with other signed properties.

### LINKED ISSUE #773: Failed to deserialize viewModel encrypted values
Hi guys,
we decided to migrate project to DotVVM 2.x.x (none of them doesnt work) and found this issue. It worked in version 1.1.9.1 and lower.

Problem is that you can't do postback in example below:
```
public class DefaultViewModel : DotvvmViewModelBase
    {
        [Protect(ProtectMode.SignData)]
        public TypA Obj_A { get; set; } = new TypA();


        public void Postback()
        {

        }


        public class TypA
        {
            [Protect(ProtectMode.SignData)]
            public string ID { get; set; }
        }
    }
```
I know that this example doesnt make sense. 
But we have many classes with protected and unprotected properties together. And we want  to
decide with object implementation if data can be passed to unprotected properties or not.

```
[Protect(ProtectMode.SignData)]
public TypA Obj_A { get; set; } = new TypA(); //data CAN'T be passed to unprotected properties

[Protect(ProtectMode.None)]
public TypA Obj_B { get; set; } = new TypA(); //data CAN be passed to unprotected properties
```

Thanks for reply.

exyi:
It does not get double the security, but this case may happen when some objects are reused. (see #773)
The implementation is not optimal, still both of the properties will be included in the
`$encryptedValues` field, making it longer that needed. It would be more work to implement it
optimally and multiple nested signed properties should not be a frequent use-case.

resolves #773


tomasherceg:
1. What about if somebody turns on Encryption on a higher level, but turns it off explicitly in the child object? I think that we should throw an exception in such case.

2. Shall I add/extend the UI test for this?

exyi:
> What about if somebody turns on Encryption on a higher level, but turns it off explicitly in the child object? I think that we should throw an exception in such case.

That is a valid point, but it's not that trivial to fix that :/ There is few problems
* Since it worked fine (the `[Protect(None)]` was ignored), throwing an exception would break someone's code. It will happen at runtime and may require obscure situation to occur. I don't know how likely is that anyone actually used `[Protect(None)]`.
* We would have to emit a check in the serializer that we are not in the encrypted values context, which we don't know.

I actually think it will be much easier to define `[Protect(None)]` as "does not bring any protection to the object, apart the the protection inherited from the properties that contain this entire object".

tomasherceg:
Makes sense - let's use the "don't require signing/encryption" semantics for the `Protect.None` attribute.

What about the UI test? Shall I add it?

PR #775, opened 2019-10-10
Selenium Generator

quigamdev:
This PR adds a new tool, that is able to generate selenium page object classes. Each page object class represents a strongly-typed definition of manipulation object for selenium tests. 

- [x] Add OWIN project support
- [ ] Check that all controls have its own proxy class
- [ ] Add support for `ko` comment elements
- [ ] Add tests for ANC projects
- [ ] Add tests for OWIN projects


ENgateman:
I spent some time thinking about the selenium generator and related libraries and executables, found some problems with them, and came up with two proposals to solve these problems.

I don't expect us to implement either of them to the fullest extent and definitely not right away but perhaps we can solve at least some of those issues one way or another.

I realize that some of these problems are a consequence of there not being a better way to do things when the code was first written. All the more reason to fix it now when there is.

**The Problems**

* Custom parsing of command line arguments is error-prone. If at all possible we should use an existing solution instead of implementing our own.

* Custom parsing of MSBuild projects using the XML is also error-prone and on top of that presumes that our users are not MSBuild power users (e.g. they don't use *.props files).

* The command line logic is too distributed. It is contained within six projects (i.e. `DotVVM.CommandLine`, `DotVVM.CommandLine.Core`, `DotVVM.Compiler`, `DotVVM.Framework.Tools.SeleniumGenerator`, `DotVVM.Utils.ConfigurationHost`, and `DotVVM.Utils.ProjectService`.

* `DotVVM.Compiler.Light`. What is its purpose? Is it just dead code?

**Proposal One: DotVVM.Cli with MSBuildLocator**

* Let' use [System.CommandLine](https://github.com/dotnet/command-line-api) or [CommandLineUtils](https://github.com/natemcmaster/CommandLineUtils) to handle the subcommands and argument parsing.

* Let's use MSBuild itself to extract the data we need from the *.csproj files. That doesn't mean we'd have to package MSBuild ourselves. We can use existing MSBuild installations that the users must have installed with Visual Studio or the .NET Core SDK with [Microsoft.Build.Locator](https://github.com/microsoft/MSBuildLocator).

* Merge all six projects into a single executable project called `DotVVM.Cli` and structure it similarly to the [dotnet](https://github.com/dotnet/toolset) CLI. Make this executable the `dotnet-dotvvm` tool.

* Remove `DotVVM.Compiler.Light`?

**Proposal Two: DotVVM.Cli with automatic config export**

Proposal Two is the same as Proposal One with one major difference: How about we don't care about the *.csproj files and application assemblies at all? Can't we just make the config export itself upon construction when the application is run and just work with this automatically exported data?

exyi:
>     |DotVVM.Compiler.Light|. What is its purpose? Is it just dead code?

I think so. It was a dotnet Core alternative to DotVVM.Compiler when it
could only run on Windows. That should be solved by now.



>     Merge all six projects into a single executable project called
>     |DotVVM.Cli| and structure it similarly to the dotnet
>     <https://github.com/dotnet/toolset> CLI. Make this executable the
>     |dotnet-dotvvm| tool.

There is one glitch - the DotVVM.Compiler (the logic that basically runs
the application) must use exactly the same version of DotVVM as the
project. DotVVM CLI on the other hand should work with different
versions, so you don't run into issue if you have multiple DotVVM
projects and a global installation of the tool.


> Proposal Two is the same as Proposal One with one major difference: How
> about we don't care about the *.csproj files and application assemblies
> at all? Can't we just make the config export itself upon construction
> when the application is run and just work with this automatically
> exported data?

Maybe it would also work. However, it is already quite problematic that
you can build DotVVM related stuff after the project is built. This is
not a huge issue for the selenium generator as you are generating code
for a different project, but would be issue for the non-existent
Typescript generator.

Adding dependency on actually running the application, not only building
it would probably break even the "generating code for other project"
use-case.

In some cases it might be also quite hard to get the information from
running project. For example, if you have some specific dependecies in
the project and build/run/debug a Docker container directly, this
approach is going to fail.

cafour:
A lot has changed since any work has been done in this PR. Most of its commits have been merged into `main` through `cli`. This PR thus just adds more control proxies, control generators, and uses XPath to find nodes. A lot remains to be done if SeleniumGenerator is ever to function and be practical to use, but given that its already half-merged into `main`, I propose we merge this as well, and start a new discussion about it.

cafour (approved):

src/Tools/SeleniumGenerator/Helpers/RoslynHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace DotVVM.Framework.Tools.SeleniumGenerator.Helpers
+{
+    public static class RoslynHelper
+    {
+        public static ObjectCreationExpressionSyntax GetPathSelectorObjectInitialization(string propertyName)
+        {
+            return SyntaxFactory.ObjectCreationExpression(
```
exyi, line 12: I don't think we'd lose anything if we just used string concatenation instead of Roslyn AST... it does not solve any problem for us (not even correct parenthesis) and is super annoying to use.

src/Framework/Testing.SeleniumHelpers/PathSelector.cs
```diff
@@ -0,0 +1,28 @@
...
+            //    return Parent.ToString() + "_" + UiName;
+            //}
+
+            //return UiName;
+            if (Index == null)
+            {
+                return $"//*[@data-uitest-name='{UiName}']";
+            }
+
+            return UiName;
```
exyi, line 25: This does not seem like it would work...

src/Tools/SeleniumGenerator/DotvvmServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public static void AddSeleniumGenerator(this IDotvvmServiceCollection services,
+            Action<SeleniumGeneratorOptions> optionsBuilder = null)
+        {
+            services.Services.AddSingleton<SeleniumGeneratorOptions>(provider =>
+            {
+                var options = new SeleniumGeneratorOptions();
+                optionsBuilder?.Invoke(options);
+
+                return options;
```
exyi, line 17: Maybe we could use IOptions<T> here?

src/Tools/SeleniumGenerator/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace DotVVM.Framework.Tools.SeleniumGenerator.Extensions
+{
+    public static class DictionaryExtensions
+    {
+        public static Dictionary<TKey, TValue> AddRange<TKey, TValue>(this Dictionary<TKey, TValue> first,
```
exyi, line 10:
```suggestion
        public static Dictionary<TKey, TValue> Union<TKey, TValue>(this Dictionary<TKey, TValue> first,
```

src/Tools/SeleniumGenerator/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+            Dictionary<TKey, TValue> second)
+        {
+            return first.Union(second).ToDictionary(t => t.Key, t => t.Value);
+        }
+    }
+    public static class ServiceProviderExtensions
+    {
+        public static T TryGetService<T>(this IServiceProvider serviceProvider)
```
exyi, line 18: Is this needed? GetService will return null by default when the service is not registred.

exyi (approved):
feel free to merge this (I would slightly prefer squash merge as this has quite convoluted history).

cafour:
@exyi Thanks for the review. I addressed some of it. The rest should IMHO be fixed in a separate PR but at the very least I added some `TODO`s. And sure, I'll squash it. :)

PR #783, opened 2019-10-26
Support complex bindings of CSS classes

### LINKED ISSUE #754: Support complex binding scenarios on the class attribute
Currently there are these ways to bind the class attribute:

__`HtmlGenericControl.Attributes`__:
Binds a string value to the HTML attribute itself using `class="{value: SomeValue}"`.

__`HtmlGenericControl.CssClasses`__:
Adds `some-class` only if the `Class-some-class="{value: DisplaySomeClass}"` binding evaluates to true.

These two approaches are mutually exclusive and there's currently no way to have an element with both class constants and classes added dynamically using a binding.

Perhaps we could support these complex scenarios by finding a way to combine the two approaches or by introducing a new property (e.g. `AdditionalCssClasses` or something)?

ENgateman:
Changes the rendering of bindings to the `class` attribute of `HtmlGenericControl` to use the knockout `class` binding. This makes it possible to use a binding to the class attribute alongside the `Class-` property group and have them work together rather nicely. (A value binding to the class attribute overrides those to the `Class-` group right now.)

If a class is referenced by both the `class` binding and a `Class-` binding, the latest change wins.

Closes #754.


tomasherceg:
The UI tests report that it doesn't work correctly in GridViewRowDecorator case - @ENgateman can you look at it? I am not sure if the Knockout `class` binding is the right way.

PR #784, opened 2019-10-26
Fix a CheckedItems "bug" when set to null

ENgateman:
If a collection bound to the `CheckBox.CheckedItems` property is set to null during the initial request, the `checked` binding behaves as if there was a boolean. Thus when the user checks the input, it assigns `true` to the collection which crashes the deserialization process.

This PR attempts to fix that with a new binding, `dotvvm-checkedItems` that wraps the `checked` binding but preprocesses the value so that if null, it evaluates to an empty array.

@tomasherceg  @exyi Is this even something we should support or should we just throw an exception when `CheckedItems` is null?


tomasherceg:
We should throw client-side error when the property value is non-array.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -446,9 +446,9 @@ class DotVVM {
                 errorCallback({ xhr });
                 dotvvm.events.staticCommandMethodFailed.trigger({ ...data, xhr })
             },
-            xhr => {
-                xhr.setRequestHeader("X-PostbackType", "StaticCommand");
-            });
+                xhr => {
+                    xhr.setRequestHeader("X-PostbackType", "StaticCommand");
+                });
         })()
```
exyi: Could you please not reformat code? The same file was changed in #787 and we'll later have to go through all the changes in JS and merge them into `js-modules`. This just makes merging very annoying.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -1321,6 +1321,22 @@ class DotVVM {
...
+            init: ko.bindingHandlers.checked.init,
+            options: ko.bindingHandlers.checked.options,
+            update: function (element, valueAccessor, allBindings, viewModel, bindingContext) {
+                var value = valueAccessor();
+                if (!Array.isArray(valueAccessor()) && (!ko.isObservable(value) || !Array.isArray(value()))) {
+                    throw Error("The value of a `checkedItems` binding must be an array (i.e. not null nor undefined).");
+                }
+                if (ko.bindingHandlers.checked.update) {
+                    // although `checked` doesn't have an update, call it conditionally in case we ever add one
+                    return ko.bindingHandlers.checked.update(element, valueAccessor, allBindings, viewModel, bindingContext);
```
exyi: I don't know how knockout handles `this` and if anything may go wrong, but you are calling `ko.bindingHandlers.checked.update` function in a different context than `ko.bindingHandlers.checked.options`. I'd be careful with that.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -1321,6 +1321,22 @@ class DotVVM {
             }
         };
 
+        ko.bindingHandlers["dotvvm-checkedItems"] = {
+            after: ko.bindingHandlers.checked.after,
+            init: ko.bindingHandlers.checked.init,
+            options: ko.bindingHandlers.checked.options,
+            update: function (element, valueAccessor, allBindings, viewModel, bindingContext) {
+                var value = valueAccessor();
+                if (!Array.isArray(valueAccessor()) && (!ko.isObservable(value) || !Array.isArray(value()))) {
```
exyi: Is this code equivalent to ``Array.isArray(ko.unwrap(value))`, just more ... intriguing?

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -1321,6 +1321,22 @@ class DotVVM {
             }
         };
 
+        ko.bindingHandlers["dotvvm-checkedItems"] = {
+            after: ko.bindingHandlers.checked.after,
+            init: ko.bindingHandlers.checked.init!.bind(this),
+            options: ko.bindingHandlers.checked.options,
```
exyi: Ahhh :fearful: `this` is now different for `init`, `checked` and `update`

ENgateman: I removed the calls to `bind` altogether since, after inspection of `applyBindingsToNodeInternal` in knockout a arrived at the conclusion that `this` is always the Window object both in `init` and `update`.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -555,7 +555,7 @@ class DotVVM {
         return new Promise<() => Promise<DotvvmAfterPostBackEventArgs>>(async (resolve, reject) => {
             const viewModelName = options.viewModelName!;
             const viewModel = this.viewModels[viewModelName].viewModel;
-            
+
```
ENgateman: @exyi I'll leave these two last reformatting changes in. I'm fine with trimming whitespace-only lines automatically.

exyi: Thats fine, I'm not likely to get in the history of this line :)

PR #789, opened 2019-11-22
Support running UI tests on Linux

ENgateman:


exyi:
Internet Explorer does seem to not support iterable, so we should be rather careful using it (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator). If you want to use it, we can probably add a polyfill for that..

I can't replicate the issue with `tsc`, current master branch compiles without any error on my PC. I have tried versions `3.7.2` (my system) and `3.3.4` (what DotVVM uses) on commit `a99e054628fc13c1956404c3ba22c9d6c3a00fd3`.

exyi:
Just tried it and found out that it depends on internal release of "Riganti.Selenium.*" packages. We should avoid using these version in the master branch of DotVVM, if we want to keep this project open to anybody, it should be possible (and easy) to build it without special credentials.

exyi:
And YES, it finally work on my machine :tada: Thank you @ENgateman!

One more question, why does it default to `http://localhost:16018` when the Asp.Net Samples start on port 5000? Is it possible to change the port from the command line or do I have to edit the json files?

ENgateman:
I didn't realize I was using an internal release of `Riganti.Selenium.*`. We should probably merge this PR after we release that.

AFAIK the only way to change the port number is through `seleniumConfig.json`. It would be nice if one could at least override that with an environment variable.

quigamdev:
@ENgateman Version  2.0.5 of Riganti.Selenium was released. Could you upgrade it?

ENgateman:
Updated 'em packages. Finally.

PR #790, opened 2019-11-22
Add es2015.iterable to tsconfig.json

ENgateman:
`tsc` doesn't seem to be willing to compile DotVVM otherwise.

### DIFF
--- a/src/DotVVM.Framework/tsconfig.json
+++ b/src/DotVVM.Framework/tsconfig.json
@@ -8,7 +8,7 @@
     "outFile": "Resources/Scripts/DotVVM.js",
     "declaration": true,
     "strictNullChecks": true,
-    "lib": ["dom", "es2015.promise", "es5"]
+    "lib": ["dom", "es2015.promise", "es2015.iterable", "es5"]
   },
   "files": [
     "Resources/Scripts/DotVVM.Polyfills.ts",


exyi:
Internet Explorer does seem to not support iterable, so we should be rather careful using it (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator). If you want to use it, we can probably add a polyfill for that..

I can't replicate the issue with `tsc`, current master branch compiles without any error on my PC. I have tried versions `3.7.2` (my system) and `3.3.4` (what DotVVM uses) on commit `a99e054628fc13c1956404c3ba22c9d6c3a00fd3`.

tomasherceg:
This is not necessary, just delete node_modules and run npm install and it should work.

S pozdravem
Tomáš Herceg

RIGANTI s.r.o. | DOTNETCOLLEGE s.r.o. | DotVVM
www.riganti.cz | www.dotnetcollege.cz | www.dotvvm.com

ENgateman:
@tomasherceg I tried `npm install && npm run build` on a clean dotvvm repo but compilation with `tsc` fails nonetheless.

@exyi I didn't want to use it for anything specific, I just needed the scripts to compile. It occurs on my machine with `tsc v3.7.2`, `npm v3.5.2`, and `nodejs v8.10.0`.

When I force npm to install `tsc v3.3.4000` by changing package.json like this:
```diff
-    "typescript": "^3.3.4",
+    "typescript": "3.3.4000",
```
...it works. It might be behaving differently on my machine because the `npm` and `nodejs` I use are rather outdated. What versions are you using?

exyi:
> When I force npm to install |tsc v3.3.4000| by changing package.json
> like this:
> 
> - "typescript": "^3.3.4",
> + "typescript": "3.3.4000",
> 
> ...it works. It might be behaving differently on my machine because the
> |npm| and |nodejs| I use are rather outdated. What versions are you using?



That's it, we use npm lock file to force it to use tsc 3.3.4000, but
that it a new feature (I think from version 5). They don't really do
breaking changes in npm, I'd recommend you to always upgrade to the
latest possible version.

ENgateman:
@exyi Yep. It works with npm 6.13.1 just fine. Closing.

PR #796, opened 2019-12-26
Made TokenBase mutable as the VS Extension needs that

exyi:
After discussion with @Mylan719 about migrating the entire parser and resolver to pure functional programming using very nice concepts such as zippers (also knows as Roslyn Red-Greed tree) and performance of this thing, we decided we have different plans for the next year so I added back the setters :)


exyi:
> I know that testing setters is ridiculous.

My intent was just to make sure that the setters don't disappear.

> Maybe we could build syntax tree out of the tokens, change tokens and check tree tokens are changed?

That's pretty good idea, it will prevent us from caching the text somewhere :)

PR #798, opened 2020-01-05
Fix GridView (with virtual elements)

### LINKED ISSUE #797: Evaluation of bindings crashes when new array is assigned to gridview datasource observable array
I was tasked to look at issue with GridView not updating on client side no matter what. After some investigation I traced the issue to an undesirable behavior inside `makeUpdatableChildrenContextHandler` binding handler. This handler is used in a custom dotvvm binding intended to update GridView control. 

Let's imagine we have viewmodel property 
```CSharp
List<string> Entries {get; set;} = new List<string> { "One", "Two" }
```
The property `Entries` is bound as a datasource for a GridView control in our markup. 
The GridView loads correctly in the html page. However when I try to update  `Entries` using javascript from the console like this:
```JavaScript
var newArray = [ko.observable("aaaa")];
dotvvm.viewModels.root.viewModel.ObjectInVirtualElement().Entries(newArray);
```
I get an error: `Error: You cannot apply bindings multiple times to the same element.`
I believe updating an observable array like this should be possible and supported.

Most likely cause of this problem is a computed property created from `makeUpdatableChildrenContextHandler`. It calls knockout API `applyBindingsToDescendants` to apply changes to main element of the gridview. `applyBindingsToDescendants` does not like to be called on an element which already has binding applied to it. And of course it makes sense that the element is already bound to the observable array `Entities`. Maybe this is not the right API to call? It seems to me `applyBindingsToDescendants` is designed to be called only one time on binding initialization. But here it is called multiple times from computed property callback.

exyi:
Fixed a problem with grid view that occurred in GridView.

The problem was, that when an observable was returned from `valueAccessor`, knockout did not count it as a dependency when this was checked by the handler. This problem should be fixed by the `ko.unwrap`.

I still don't know when exactly did the problem occur, which kinda triggers me... so if you have a explanation please post it to the PR :)

fixes #797


tomasherceg:
Strange, I also don't understand the reasons, but never mind - I've added a Selenium test to make sure this scenario works. I'll also give it a try on a real project to be sure that the issue is gone.
When the tests pass, we are ready to merge.

exyi:
> When the tests pass, we are ready to merge.

Sure. Could you also release public NuGet package reasonably quickly
after that? I think this bug was introduced in 2.4, so it may be
breaking existing code.

PR #804, opened 2020-01-26
Fixed passing of command arguments that are of non-primitive type

exyi:
The command arguments must be deserialized with knowledge of the expected type.

Apart from that, I also had to fix strange behavior of KnockoutHelper.GeneratePostbackExpression


src/DotVVM.Samples.Tests/Feature/CommandArgumentsTests.cs
```diff
@@ -22,7 +22,7 @@ public void Feature_CommandArguments_CommandArguments()
                 var text = browser.Single("[data-ui='value']");
                 AssertUI.InnerTextEquals(text, "Nothing here");
 
-                browser.Single("[data-ui='button'] button").Click();
+                browser.Single("[data-ui='button1'] button").Click().Wait();
```
quigamdev, line 25: I can't see any reason for using `Wait()` method here. Please use `WaitFor()`.

PR #805, opened 2020-01-29
Implemented LRU Cache and used it as default

exyi:
This PR should solve the problems with the default caching mechanism (with adapters):
* on OWIN it was completely broken, so I have removed it completely
* on Asp.Net Core it only had problems with cache retention, so I have only changed the default
    - in general, the MemoryCache is not a great fit for our use-case since we don't know the binding size, we don't have any expiration and we don't know how much memory can we use.

The new LRU cache will have a predefined size and predefined retention interval and will eject the items
that were unused for longest time (not created first as expiration time would do).

The default cache size is 1000...2000 (it should at most 5MB if it's full of bindings) and the default timeout is 10 minutes.

Note that I'm not sure whether we want to use LRU dictionary as implementation of the Cache adapter or rather use it strictly for the binding cache. In that case, however, the adapter mechanism probably becomes unused.


src/DotVVM.Framework/Runtime/Caching/SimpleLruDictionary.cs
```diff
@@ -0,0 +1,184 @@
...
+                return true;
+            lock (locker)
+            {
+                if (hot.TryGetValue(key, out value))
+                    return true;
+                if (cold.TryGetValue(key, out value))
+                {
+                    // add back to `hot` -- this way, we mark the item as used
+                    hot.TryAdd(key, value);
+                    return true;
```
tomasherceg, line 87: Is it OK that we copy the item to the hot dictionary while keeping it in cold?

exyi: That does not matter much. Clearing it from `cold` won't free any memory, though, so it's not useful for anything.

PR #813, opened 2020-03-14
Render adapter

MichalTichy:
Render adapter adds ability to provide custom implementation for control render methods.

This feature currently has to be turned on using experimental flag `ControlRenderAdapters`.
Custom implementation of render methods is provided to the controls as `IRenderAdapter` via DotvvmProperty `Internal.RenderAdapterProperty`.

### DIFF
--- a/src/DotVVM.Framework/Controls/DotvvmControl.cs
+++ b/src/DotVVM.Framework/Controls/DotvvmControl.cs
@@ -294,20 +294,20 @@ namespace DotVVM.Framework.Controls
             else
                 AddAttributesToRender(writer, context);
 
-            var renderBeginTagAdapter = renderAdapter?.AddAttributesToRender;
+            var renderBeginTagAdapter = renderAdapter?.RenderBeginTag;
             if (renderBeginTagAdapter != null)
                 renderBeginTagAdapter(this, writer, context);
             else
                 RenderBeginTag(writer, context);
 
 
-            var renderContentsAdapter = renderAdapter?.AddAttributesToRender;
+            var renderContentsAdapter = renderAdapter?.RenderContents;
             if (renderContentsAdapter != null)
                 renderContentsAdapter(this, writer, context);
             else
                 RenderContents(writer, context);
 
-            var renderEndTagAdapter = renderAdapter?.AddAttributesToRender;
+            var renderEndTagAdapter = renderAdapter?.RenderEndTag;
             if (renderEndTagAdapter != null)
                 renderEndTagAdapter(this, writer, context);
             else
@@ -467,7 +467,7 @@ namespace DotVVM.Framework.Controls
             if (!context.Configuration.ExperimentalFeatures.ControlRenderAdapters.IsEnabledForRoute(routeRouteName))
                 return null;
 
-            return (IRenderAdapter?)GetValue(Internal.RenderAdapterProperty);
+            return (IRenderAdapter?)GetValueRaw(Internal.RenderAdapterProperty);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/DotVVM.Framework/Controls/DotvvmControl.cs
```diff
@@ -435,6 +457,19 @@ public static bool IsNamingContainer(DotvvmBindableObject control)
...
+        /// Determines whether the specified control is a naming container.
+        /// </summary>
+        protected IRenderAdapter? GetRenderAdapter(IDotvvmRequestContext context)
+        {
+            var routeRouteName = context.Route.RouteName;
+
+            if (!context.Configuration.ExperimentalFeatures.ControlRenderAdapters.IsEnabledForRoute(routeRouteName))
+                return null;
+
+            return (IRenderAdapter?)GetValueRaw(Internal.RenderAdapterProperty);
```
exyi: If we want to merge, I'll have to resuscitate the benchmarks :/ This is one of the hottest paths in DotVVM and checking the `IsEnabledForRoute` for every control does not seem like a terribly good idea.

exyi: In general, I don't think we need to hide behind an experimental feature. It's not going to do anything, unless something starts using it.

src/DotVVM.Framework/Controls/DotvvmControl.cs
```diff
@@ -286,10 +286,32 @@ protected virtual void RenderControl(IHtmlWriter writer, IDotvvmRequestContext c
...
+            if (renderContentsAdapter != null)
+                renderContentsAdapter(this, writer, context);
+            else
+                RenderContents(writer, context);
+
+            var renderEndTagAdapter = renderAdapter?.RenderEndTag;
+            if (renderEndTagAdapter != null)
+                renderEndTagAdapter(this, writer, context);
+            else
+                RenderEndTag(writer, context);
```
exyi: This `RenderControl` method is virtual and some controls just override it and control the rendering on themselves. For these controls, this render adapter mechanism will never work, but this is private implementation detail of those controls that should not be observable from outside. This adapter breaks the encapsulation:

* Will make the behavior inconsistent across controls
* Will make any change in this implementation a breaking change
* Since the adapters will depend on the distribution of logic between those methods, moving anything around will also break compatibility
...
* Actually, even adding new properties to controls will start breaking code, because the adapters will not count with that. And new parts of the code base might depend on it :/

MichalTichy:
The main reason for this feature is to allow easy way how to render DotVVM control in entirely different way. More specifically, I'm trying to allow users to render DotVVM pages (in simplified form) as AMP pages.

We've determined that RenderAdapters would be the best option to implement this behavior.

src/DotVVM.Framework/Controls/DotvvmControl.cs
```diff
@@ -249,12 +249,12 @@ protected bool RenderBeforeControl(in RenderState r, IHtmlWriter writer, IDotvvm
...
             if (r.HasActives) foreach (var item in properties)
-            {
-                if (item.Key is ActiveDotvvmProperty activeProp)
                 {
-                    activeProp.AddAttributesToRender(writer, context, this);
+                    if (item.Key is ActiveDotvvmProperty activeProp)
+                    {
+                        activeProp.AddAttributesToRender(writer, context, this);
+                    }
                 }
```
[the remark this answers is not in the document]
MichalTichy: ups I've no clue how this happened.

src/DotVVM.Framework/Controls/DotvvmControl.cs
MichalTichy: Actually the main point for using ExperimentalFeatures flag was to limit this feature to only some routes because I was worried about performance of getting value of Internal.RenderAdapterProperty

exyi:
@MichalTichy I guess you have solved it in another way, so you can close this, right?

PR #814, opened 2020-03-29
Migrate client-side code to modern Javascript

exyi:
Specifically, it is now spitted into ES modules, uses async wherever possible.

It is bundled using Rollup into a single ES module, or System.js module to support old IE browsers.

SPAs and normal pages now use different bundles, so we got rid of few ugly runtime checks, but the impact on bundle size it not spectacular. However, the overall impact on bundle size due to this refactor is about 2x, which is quite cool, I'd say :)

The API surface of DotVVM is now *much* smaller, so we should be able to maintain this API as stable

There still are things to do:

- [ ] make the JS tests work, and maybe migrate them to `yarn test` or something
- [ ] write docs for the public API, when we want to stabilize it


exyi:
It is merged into `v3-master`

PR #816, opened 2020-03-29
Fixed race condition in queued postbacks

tomasherceg:
This PR fixes two issues:

* When there are multiple postbacks in a concurrency queue, more of them started executing simultaneously
* When the server responds with a redirect response, the next postbacks from queue were started (during the load of the new page)

### DIFF
--- a/src/DotVVM.Samples.Common/ViewModels/FeatureSamples/PostbackConcurrency/RedirectPostbackQueueViewModel.cs
+++ b/src/DotVVM.Samples.Common/ViewModels/FeatureSamples/PostbackConcurrency/RedirectPostbackQueueViewModel.cs
@@ -14,7 +14,7 @@ namespace DotVVM.Samples.Common.ViewModels.FeatureSamples.PostbackConcurrency
 
         public override async Task Init()
         {
-            if (Context.Query.ContainsKey("time"))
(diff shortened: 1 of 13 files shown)


src/DotVVM.Samples.Common/ViewModels/FeatureSamples/PostbackConcurrency/RedirectPostbackQueueViewModel.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace DotVVM.Samples.Common.ViewModels.FeatureSamples.PostbackConcurrency
+{
+    public class RedirectPostbackQueueViewModel : RedirectPostbackQueueMasterViewModel
+    {
+
+        public int Value { get; set; }
+
+        public override async Task Init()
+        {
+            if (Context.Query.ContainsKey("time"))
```
exyi: Could you please wait only for the first load. Otherwise, manual clicking on the sample may be quite annoying

tomasherceg: Fixed - I didn't expect anyone to click the buttons after the redirect - the UI tests use the buttons only before the redirect. But for humans, it'll be easier.

PR #817, opened 2020-03-30
Null propagation now passes nulls into methods instead of skipping execution

### LINKED ISSUE #808: Passing property of object which is null as parameter in command would not start the command but return 200 anyway.
Example:
Foo = null;
<dot:button Click="{command: Test(Foo.Bar}" />
This commands goes will not be run at the server but its request will return 200 regardless.
User is left without any hint why this happened.
We should probably throw exception in such scenarios.

exyi:
resolves #808 

Null propagation used to skip method if at least one of the arguments was null, but this is quite strange, so let's switch to allowing the nulls there. If the type is a value type, an exception is thrown.


quigamdev (approved):
I haven't found anything that would be preventing this PR from merging. Yesterday I found an issue with an indexer which is already fixed. So I think this PR is ready to go.

PR #820, opened 2020-04-10
Add IDotvvmControlLike marker interface

exyi:


quigamdev:
@exyi 
1) Why do we need this marker? Is there some issue explaining why we need this? 
2) There is a build error `src\DotVVM.Framework\Compilation\ControlTree\ControlResolverMetadataBase.cs(39,132): Error CS0246: The type or namespace name 'IDotvvmControl' could not be found (are you missing a using directive or an assembly reference?)
`. Can you please check whether the error is associated with the changes you have made?

exyi:
> Why do we need this marker? Is there some issue explaining why we need this?

As a replacement for `IDotvvmControl`, which does not make much sense.

Since people often want to define interfaces for controls, it's used on quite a bit of places in BS and BP. 

I'd be also happy to just remove the `IDotvvmControl` interface, if nobody needs such thing.

> There is a build error...

fixed, sorry

quigamdev:
Ok, I see. I have checked most of DotVVM projects I have on my PC and there is a negligible number of usages of IDotvvmControl interface. Based on this I think this change should not break anybody. 
Also, there will be needed some adjustments to DotVVM Academy and dynamic data package.

quigamdev:
CI & tests passed at last run.

exyi:
> CI & tests passed at last run.

Well, at this point we don't really care about that... I wanted to discuss how to control interfaces properly, there is quite a bit of unsolved problems:

* How should control properties be represented?
     - `string Text { get; }` does not allow you to get bindings
     - `ValueOrBinding<string> Text { get; }` will be very annoying to implement for controls that have the `string Text { get; }` property
     - somehow get the `DotvvmProperty TextProperty { get; }`... but how?
     - any other ?
* If we want this way of having control interfaces, or better figure out something else.


And this should have gone to `master-v3`, and was a draft for a reason... I'm revering that merge

PR #824, opened 2020-04-18
Add attributes to property groups

exyi:
so VS Extension has easier time doing the discovery


exyi:
I have fixed the issue with the registration, so it should be finally unblocked. Anyone for review?

PR #826, opened 2020-04-18
Add build dependency on reference assemblies for net451 builds

ENgateman:
Make building on linux as simple as `dotnet build` with a build dependency on the reference assemblies.  Based on https://github.com/dotnet/sdk/issues/826.

There's no easy way to tell the dotnet CLI to *not* build net451 targets on linux so those builds might as well not fail.

Btw. @exyi How were you getting around this? With `--framework netcoreapp*/netstandard*`?


exyi:
Very cool, I did not know this is possible. Yes, I was always using `--framework netcoreapp` when running the samples. I'll check tomorrow that it actually works and merge it :)

Issue #776, opened 2019-10-11
Allow use of a read-only variable in static command bindings


Mylan719:
Some time ago I've found myself  in a situation where I wanted to update multiple viewmodel properties in one button click. I wanted to load both user information and a shopping cart content to fill the top panel of my page when a button is clicked.
Let's have a viewmodel:
```csharp
public class ThePage
{
    public ShopingCartViewModel ShoppingCart { get; set; }
    public UserInfo { get; set; }
    //...
}
```
Of course the viewmodel contains lots of other properties, and I don't want to drag them all to the server side and back just because I need to update 2 properties...
...so I used a service with contract like this:
```csharp
public class PanelInfoDto{
    public ShopingCartViewModel ShoppingCart { get; set; }
    public UserInfo { get; set; }
}

public interface IPanelInfoService
{
    ShopingCartViewModel  GetShoppingCart();
    UserInfo  GetUserInfo();
    PanelInfoDto GetPanelInfo();
}
```
I realized, that there is no practical way to update two properties at once using single service call inside of static command.

Only 2 ways to solve this now:

1.  Using two calls: `staticCommand: ShoppingCart = service.GetShoppingCart(); UserInfo = service.GetUserInfo()`. This is less then ideal because I need to make 2 calls to the server which is slower. This method does not work for all cases. There are operations that require updating multiple properties which cannot be repeated (Save, Refresh of items,...).

2. Using temporary viewmodel property: `staticCommand: Temp = service.GetPanelInfo(); UserInfo = Temp.UserInfo(); ShoppingCart = Temp.ShoppingCart`. This method also works for operations that cannot be repeated. However I find it clunky and hard on an unfamiliar reader.

My proposed solution:
```html
<dot:Button Text="Update" 
            Click="{staticCommand: let info = service.GetPanelInfo(); 
                                              UserInfo = info.UserInfo(); 
                                              ShoppingCart = info.ShoppingCart}" />
```
Read-only binding local variable that would be defined in the first statement and could be used in all following statements  in the same binding.

This variable would be also useful in following use cases:

- saving data and getting updated repeater items back together with success/error message
- updating 2 grids at once
- updating grid and filling items of dependent combo-boxes at the same time
- updating multiple filters that depend on each others values

tomasherceg:
My first impression is that I don't like variables in the bindings. But on the second thought, it makes sense and it is certainly better than using a temp property in the viewmodel (which is possible right now).

[...]

PR #833, opened 2020-05-08
C# functions in DotHTML files

tomasherceg:
This is a proof of concept for a new feature - C# functions in DotHTML files that would be translated to JavaScript and called using static commands.

There is a new control `<dot:ClientModule>`. It must be used in the root context and must not be wrapped by any elements. 

It can contain public C# methods that can accept parameters and return values. The restrictions are equal to static commands - the C# code is translated in JS using static command compilation routine. 

```
<dot:ClientModule>
    public void Increment(ClientModuleViewModel module) 
    {
        module.Value1 = module.Value1 + 1;
        module.Value2 = module.Value2 + 2;        
    }    
</dot:ClientModule>
```

This block extends the binding system with the new `_client` variable that will contain the `Increment` method. It is then possible to call it like this: `{staticCommand: _client.Increment(_this)}`

When DotVVM compiles the view, it compiles the client module into a C# class (it is required for the JS translations to work - they depend on LINQ expression API and reflection), and then it generates JS representation of the module - right now it looks like this:

```
dotvvm.clientModules = dotvvm.clientModules || {};
dotvvm.clientModules['_root'] = {
 'Increment': (function(module) {
	ko.unwrap(module).Value1(ko.unwrap(module).Value1() + 1);
	ko.unwrap(module).Value2(ko.unwrap(module).Value2() + 2);
	return null;
  }),
}
```

I still have doubts about this feature and I am not sure if it is a good idea. It is confusing that it looks like C#, but the behavior is very different from C# and most of the C# features are not supported. However, if we add new features to the static commands (e.g. variables #776), they will be available here as well.

There are few things to solve & discuss:

* [ ] Explore more use cases (async, calls to server, calls to other client methods...)
* [ ] Support for content & master pages
* [ ] Tests
* [ ] Is one client-module per view enough?
* [ ] Shall we add state?
[ ] What about more complex syntax constructs? `if`, `while`, `for`, `try..catch`... Where does this stop?

### DIFF
--- a/src/DotVVM.Compiler/Compilation/ViewStaticCompiler.cs
+++ b/src/DotVVM.Compiler/Compilation/ViewStaticCompiler.cs
@@ -35,7 +35,7 @@ namespace DotVVM.Compiler.Compilation
         private IControlTreeResolver controlTreeResolver;
         private IViewCompiler compiler;
         private IMarkupFileLoader fileLoader;
-        private CSharpCompilation compilation;
+        private CompilationBuilder compilationBuilder;
         private CompilationResult result = new CompilationResult();
 
         private void InitOptions()
@@ -98,7 +98,7 @@ namespace DotVVM.Compiler.Compilation
             if (Options.FullCompile)
             {
                 compiler = configuration.ServiceProvider.GetService<IViewCompiler>();
-                compilation = compiler.CreateCompilation(Options.AssemblyName);
+                compilationBuilder = new CompilationBuilder(configuration.Markup, Options.AssemblyName);
             }
         }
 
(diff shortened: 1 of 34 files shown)


tomasherceg:
Just another crazy idea - we can add `Mode` property to the `ClientModule` control and there would be two options - `StaticCommands` (default) and `WebAssembly`. 

The latter would take the C# code and compile it into a WebAssembly. The invoking routine would have to be changed to something like postback (serialization & deserialization) because right now, the user can pass a part of the viewmodel in the function. 
But it could be a good thing for the user to be able to have a choice "I want a very limited C# functionality & small price to pay" vs "I want full C# & downloading a few megs doesn't matter".

exyi:
> we can add Mode property to the ClientModule control and there would be two options - StaticCommands (default) and WebAssembly.

ehh, having a Mode property that decides on such a fundamental thing is a terrible design... It will just make the two look very similar while they would actually be different in almost every aspect.

> The latter would take the C# code and compile it into a WebAssembly

Also, remember that we are compiling all of this at runtime and producing WebAssembly would create a lot of problems.

> The invoking routine would have to be changed to something like postback (serialization & deserialization) because right now, the user can pass a part of the viewmodel in the function.

`staticCommand`s solve this big problem of commands that they must revert the changes made in viewmodel. This would just take the problem back to static command, so please no.

Issue #744, opened 2019-08-23
Hide ValidationSummary when ModelState is valid


vaclavholusa:
Currently when `ValidationSummary` control is placed in page, it renders plain `<ul>`.
When e.g. bootstrap css class is added to it, it renders empty rectangle.
For example:
```html
<dot:ValidationSummary class="alert alert-danger" />
```

renders
```html
<ul class="alert alert-danger" data-bind="foreach: dotvvm.validation.getValidationErrors(dotvvm.viewModelObservables['root'], true, false)"></ul>
```

which displays empty red rectangle.

I can bypass this behavior by adding  `Visible="{value: !Context.ModelState.IsValid}"`, however that renders quite ugly piece of code
```html
<ul style="display:none" class="alert alert-danger" data-bind="foreach: dotvvm.validation.getValidationErrors(dotvvm.viewModelObservables['root'], true, false), visible: (function(a) {
	return !((a = Context() &amp;&amp; Context().ModelState()) == null ? null : a.IsValid());
}())"></ul>
```

My proposal is to add some property like `HideWhenValid` which will handle this in some better way.

PR #838, opened 2020-05-21
Added HideWhenValid property to ValidationSummary

acizmarik:
This PR implements `HideWhenValid` property for the `ValidationSummary` control as described in #744


tomasherceg:
LGTM.

The tests don't work because CI doesn't pass secrets (GitHub API key) to PRs from forked repos.

src/DotVVM.Framework/Resources/Scripts/DotVVM.Validation.ts
```diff
@@ -304,6 +305,15 @@ class DotvvmValidation {
                         item.innerText = error.errorMessage;
                         element.appendChild(item);
                     }
+
+                    if (binding.hideWhenValid == true) {
```
exyi, line 309: I'd just treat all truthy values as `true` (as we do for `includeErrorsFromChildren`, ...)
```suggestion
                    if (binding.hideWhenValid) {
```

or, if you would really want to accept only `true`, then you must use `===`.
```suggestion
                    if (binding.hideWhenValid === true) {
```

PR #840, opened 2020-05-26
Added support for custom sorting strategy for GridViewDataSet

### LINKED ISSUE #764: GridViewDataSet sorting logic cannot be changed
The default sorting logic of `GridViewDataSet` is [hard-coded in GridView](https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Controls/GridView.cs#L168) and it is not possible to change it easily.

We should either:
1. Change the condition so the `SortCommand` will have priority (right now it is not used when GridViewDataSet is used). 
2. Move the sorting logic to `ISortingOptions`. This would break the API, but since it would be in 2.4 and not many people has implemented their own `ISortingOptions`, it should not be such a big issue.

acizmarik:
This PR adds `ISortingStrategy` to `ISortingOptions`. This allows to customize the default sorting logic of `GridViewDataSet`. Resolves #764

### DIFF
--- a/src/DotVVM.CommandLine/DotVVM.CommandLine.csproj
+++ b/src/DotVVM.CommandLine/DotVVM.CommandLine.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <AssemblyTitle>DotVVM.CommandLine</AssemblyTitle>
     <AssemblyName>dotnet-dotvvm</AssemblyName>
-    <VersionPrefix>2.4.0.2-preview02-74160</VersionPrefix>
+    <VersionPrefix>2.4.0.2-preview02-76679</VersionPrefix>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <AssemblyOriginatorKeyFile>dotvvmwizard.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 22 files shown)


tomasherceg:
This is basically an extensibility point for a column click behavior in `GridView` - until now, the logic was hard-coded in the `GridView` control. It's true that the same thing can be achieved with `SortChanged`, but it may be easier to implement a universal strategy and reuse it when initializing the data set rather than specifying SortCommand on all pages and having a method in the viewmodel. With this strategy, it is also quite easy to override the logic just for a specific column while keeping the original behavior for the others.

Conceptually it is similar to `INearPageIndexesProvider` which tells `DataPager` how to calculate the page numbers. 
There is no need to serialize the strategy - it will be the user's responsibility to set it on each request (similarly to near page indexes).

The idea of adding multiple sort criteria is interesting - we should make a separate issue for that.

exyi:
> With this strategy, it is also quite easy to override the logic just for a specific column while keeping the original behavior for the others.

Sure, but why can't I just read the data from the `ISortingOptions` after it's applied by the default handler? It seems to me, that legit use cases are going quite rare, and thus better served by a simpler API

tomasherceg:
I can look into ISortingOptions but I don't know what has changed. I'd need to look there in `Load`, save the column I used for sorting, and then compare it in the `PreRender` phase to see that the column was clicked.
This strategy allows me to "respond" to the event of "sort expression change". I could achieve the same result with `SortChanged`, but this is more universal and it doesn't depend on a specific UI control.

My (hypothetical) use case is this:
- When I click a specific column in GridView (e.g. "Priority"), I want it to be sorted descending by default. 
- When I click any other column, I want ascending order. (default behavior)
- When I click the same column again, the order will reverse. (default behavior)

What I want to achieve is something like this:
```
public class MySortingStrategy : DefaultSortingStrategy {
    public override void Apply(ISortableGridViewDataSet dataSet, string sortExpression) {
        if (sortExpression == "Priority" && dataSet.SortingOptions.SortExpression != "Priority") {
           dataSet.SortingOptions.SortDescending = true;
           dataSet.SortingOptions.SortExpression = sortExpression;
        }
        else {
           base.Apply(dataSet, sortExpression);
        }
    }
}
```

Also, if we extend the sorting functionality for multi-criteria sorting, this strategy makes even more sense as the users might want to apply some specific sorting strategy everywhere (I've seen something like "remember last two columns I clicked").

acizmarik:
Thanks for feedback. Comments regarding the implementation were addressed in the last commit.

acizmarik:
Closing in favor of #887

PR #846, opened 2020-06-03
Added configuration for view precompilation.

MichalTichy:
This PR adds configuration for view compilation modes.
```
/// Gets or sets the mode under which the view compilation (pages, controls, ... ) is done.
configuration.Markup.ViewCompilation.Mode = ViewCompilationMode.AfterStartup;

/// Gets or sets the delay before view compilation will be done. This compilation delay can be set only in precompilation modes.
configuration.Markup.ViewCompilation.BackgroundCompilationDelay = TimeSpan.FromMinutes(1);
```

Compilation modes are:
- Lazy 
Current mode of compilation.
- DuringApplicationStart
Compilation will run during application startup. Application will start after compilation is done.
- AfterApplicationStart
Compilation will run after application startup.

### DIFF
--- a/src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
+++ b/src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
@@ -52,7 +52,7 @@ namespace DotVVM.Framework.Configuration
 
         public void Validate()
         {
-            if (BackgroundCompilationDelay.HasValue && (Mode == ViewCompilationMode.Lazy || Mode==ViewCompilationMode.DuringStartup))
+            if (BackgroundCompilationDelay.HasValue && (Mode == ViewCompilationMode.Lazy || Mode==ViewCompilationMode.DuringApplicationStart))
             {
                 throw new Exception($"{nameof(BackgroundCompilationDelay)} must be null in {nameof(ViewCompilationMode.Lazy)} {nameof(Mode)}.");
             }
--- a/src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
+++ b/src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
@@ -7,6 +7,9 @@ namespace DotVVM.Framework.Configuration
     {
         private bool isFrozen = false;
         private ViewCompilationMode mode;
+        /// <summary>
+        /// Gets or sets the mode under which the view compilation (pages, controls, ... ) is done.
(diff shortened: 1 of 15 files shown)


src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
```diff
@@ -0,0 +1,73 @@
...
+                ThrowIfFrozen();
+                compileInParallel = value;
+            }
+        }
+
+        public void Validate()
+        {
+            if (BackgroundCompilationDelay.HasValue && Mode == ViewCompilationMode.Lazy)
+            {
+                throw new Exception($"{nameof(BackgroundCompilationDelay)} is not supported in {nameof(ViewCompilationMode.Lazy)} {nameof(Mode)}.");
```
exyi, line 57: BackgroundCompilationDelay must be null in ...

src/DotVVM.Framework/Controls/UpdateProgress.cs
```diff
@@ -93,7 +93,7 @@ public static IEnumerable<ControlUsageError> ValidateUsage(ResolvedControl contr
             {
                 if ((int)delayProperty.Value < 0)
                 {
-                    yield return new ControlUsageError("Delay cannot be set to negative number.");
+                    yield return new ControlUsageError("BackgroundCompilationDelay cannot be set to negative number.");
```
exyi, line 96: No, this is wrong :sweat_smile:

src/DotVVM.Framework/Configuration/ViewCompilationConfiguration.cs
exyi: We should also check that the delay is null when Mode is DuringStartup. That would only stop the startup, which does not make much sense.

MichalTichy: Fixed

src/DotVVM.Framework/Controls/UpdateProgress.cs
MichalTichy: Ups :O
I've reverted the change and modified all the messages in UpdateProgress to use nameof

tomasherceg:
It seems like that integration tests are failing - @MichalTichy please look at it.

MichalTichy:
Tests seem to be ok now.
Can we merge it?
@tomasherceg

PR #847, opened 2020-06-05
Added support for multiple SpaContentPlaceHolders per page

### LINKED ISSUE #719: It's not possible to have two SpaContentPlaceHolder on page
It's not able to have two `SpaContentPlaceHolder` in one page. I think this is not a bug it's design decision. But I except some exception when I have a page with two `SpaContentPlaceHolder`.

acizmarik:
This PR lifts the restriction on having only a single `SpaContentPlaceHolder` per page. Resolves #719

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
@@ -13,12 +13,10 @@ export function init(): void {
         throw new Error("No SpaContentPlaceHolder control was found!");
     }
 
-    spaPlaceHolders.forEach(function (spa) {
-        window.addEventListener("hashchange", event => handleHashChangeWithHistory(spa, false));
-        handleHashChangeWithHistory(spa, true);
+    window.addEventListener("hashchange", event => handleHashChangeWithHistory(spaPlaceHolders, false));
+    handleHashChangeWithHistory(spaPlaceHolders, true);
 
-        window.addEventListener('popstate', event => handlePopState(event, spa != null));
-    });
+    window.addEventListener('popstate', event => handlePopState(event, spaPlaceHolders[0] != null));
 }
 
 function getSpaPlaceHolders(): NodeListOf<HTMLElement> | null {
@@ -47,7 +45,7 @@ function handlePopState(event: PopStateEvent, inSpaPage: boolean) {
     }
 }
 
-function handleHashChangeWithHistory(spaPlaceHolder: HTMLElement, isInitialPageLoad: boolean) {
+function handleHashChangeWithHistory(spaPlaceHolders: NodeListOf<HTMLElement>, isInitialPageLoad: boolean) {
     if (document.location.hash.indexOf("#!/") === 0) {
         // the user requested navigation to another SPA page
         navigateCore(
@@ -56,7 +54,9 @@ function handleHashChangeWithHistory(spaPlaceHolder: HTMLElement, isInitialPageL
         );
     } else {
         isSpaReady(true);
-        spaPlaceHolder.style.display = "";
+        spaPlaceHolders.forEach(function (element) {
+            element.style.display = "";
+        });
 
         const currentRelativeUrl = location.pathname + location.search + location.hash
         replacePage(currentRelativeUrl);
--- a/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
@@ -9,7 +9,7 @@ export const isSpaReady = ko.observable(false);
 
 export function init(): void {
     const spaPlaceHolders = getSpaPlaceHolders();
-    if (spaPlaceHolders == null || spaPlaceHolders.length == 0) {
+    if (spaPlaceHolders.length == 0) {
         throw new Error("No SpaContentPlaceHolder control was found!");
     }
 
@@ -19,12 +19,12 @@ export function init(): void {
     window.addEventListener('popstate', event => handlePopState(event, true));
 }
 
-function getSpaPlaceHolders(): NodeListOf<HTMLElement> | null {
+function getSpaPlaceHolders(): NodeListOf<HTMLElement> {
     return document.getElementsByName("__dot_SpaContentPlaceHolder");
 }
 
 export function getSpaPlaceHoldersUniqueId(): string {
-    const spas = Array.from(getSpaPlaceHolders()!);
+    const spas = Array.from(getSpaPlaceHolders());
     const identifiers = spas.map((element) =>
     {
         return element.getAttribute("data-dotvvm-spacontentplaceholder")?.valueOf()
--- a/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
@@ -8,27 +8,30 @@ import { DotvvmPostbackError } from '../shared-classes';
 export const isSpaReady = ko.observable(false);
 
 export function init(): void {
-    const spaPlaceHolder = getSpaPlaceHolder();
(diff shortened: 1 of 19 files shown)


src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,28 @@ import { DotvvmPostbackError } from '../shared-classes';
...
+        throw new Error("No SpaContentPlaceHolder control was found!");
     }
 
-    window.addEventListener("hashchange", event => handleHashChangeWithHistory(spaPlaceHolder, false));
-    handleHashChangeWithHistory(spaPlaceHolder, true);
+    window.addEventListener("hashchange", event => handleHashChangeWithHistory(spaPlaceHolders, false));
+    handleHashChangeWithHistory(spaPlaceHolders, true);
 
-    window.addEventListener('popstate', event => handlePopState(event, spaPlaceHolder != null));
+    window.addEventListener('popstate', event => handlePopState(event, spaPlaceHolders[0] != null));
```
exyi, line 19: I don't see a purpose of the `spaPlaceHolders[0] != null` condition. Can it even be null?

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,28 @@ import { DotvvmPostbackError } from '../shared-classes';
 export const isSpaReady = ko.observable(false);
 
 export function init(): void {
-    const spaPlaceHolder = getSpaPlaceHolder();
-    if (spaPlaceHolder == null) {
-        throw new Error("The SpaContentPlaceHolder control was not found!");
+    const spaPlaceHolders = getSpaPlaceHolders();
+    if (!spaPlaceHolders) {
```
exyi, line 12: Negation of empty array is always false, we should check for `...length == 0` (or `!...length`)

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,28 @@ import { DotvvmPostbackError } from '../shared-classes';
...
+function getSpaPlaceHolders(): NodeListOf<HTMLElement> | null {
+    return document.getElementsByName("__dot_SpaContentPlaceHolder");
 }
 
-export function getSpaPlaceHolderUniqueId(): string {
-    return getSpaPlaceHolder()!.getAttribute("data-dotvvm-spacontentplaceholder")!;
+export function getSpaPlaceHoldersUniqueId(): string {  
+    let identifier = "";
+    getSpaPlaceHolders()!.forEach(function (element) {
+        identifier += element.getAttribute("data-dotvvm-spacontentplaceholder")?.valueOf();
```
exyi, line 29: Instead of the loop, you can  probably use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join. I also think we should a separator in the generated ID (it may be anything except for `[ca0-9_]`.

src/DotVVM.Framework/Runtime/DefaultDotvvmViewBuilder.cs
```diff
@@ -70,11 +70,8 @@ protected void VerifySpaRequest(IDotvvmRequestContext context, DotvvmView page)
             if (context.IsSpaRequest)
             {
                 var spaContentPlaceHolders = page.GetAllDescendants().OfType<SpaContentPlaceHolder>().ToList();
-                if (spaContentPlaceHolders.Count > 1)
-                {
-                    throw new Exception("Multiple controls of type <dot:SpaContentPlaceHolder /> found on the page! This control can be used only once!");   // TODO: exception handling
-                }
-                if (spaContentPlaceHolders.Count == 0 || spaContentPlaceHolders[0].GetSpaContentPlaceHolderUniqueId() != context.GetSpaContentPlaceHolderUniqueId())
+
+                if (spaContentPlaceHolders.Count == 0 || !context.GetSpaContentPlaceHolderUniqueId().Contains(spaContentPlaceHolders[0].GetSpaContentPlaceHolderUniqueId()))
```
exyi, line 74: We should split the ID by the separator and then call `Contains`. Maybe, we could even use a proper array, we do not care too much about backwards compatibility in the DotVVM protocol.

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,28 @@ import { DotvvmPostbackError } from '../shared-classes';
...
-    return null;
+function getSpaPlaceHolders(): NodeListOf<HTMLElement> | null {
+    return document.getElementsByName("__dot_SpaContentPlaceHolder");
 }
 
-export function getSpaPlaceHolderUniqueId(): string {
-    return getSpaPlaceHolder()!.getAttribute("data-dotvvm-spacontentplaceholder")!;
+export function getSpaPlaceHoldersUniqueId(): string {  
+    let identifier = "";
+    getSpaPlaceHolders()!.forEach(function (element) {
```
exyi, line 28:
```suggestion
    getSpaPlaceHolders()!.forEach((element) => {
```

We generally prefer arrow functions, it is shorter and behaves more predictably when it comes to `this`

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
acizmarik: You are right. It should not be null at this point

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
acizmarik: Fixed

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
acizmarik: Changed the implementation to use the join function. Also added ';' as delimiter.

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
acizmarik: OK. Will be using arrow function from now on.

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,29 @@ import { DotvvmPostbackError } from '../shared-classes';
 export const isSpaReady = ko.observable(false);
 
 export function init(): void {
-    const spaPlaceHolder = getSpaPlaceHolder();
-    if (spaPlaceHolder == null) {
-        throw new Error("The SpaContentPlaceHolder control was not found!");
+    const spaPlaceHolders = getSpaPlaceHolders();
+    if (spaPlaceHolders == null || spaPlaceHolders.length == 0) {
```
exyi, line 12: `spaPlaceHolders` will never be null, see https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByName

src/DotVVM.Framework/Resources/Scripts/spa/spa.ts
```diff
@@ -8,27 +8,29 @@ import { DotvvmPostbackError } from '../shared-classes';
...
-    }
-    return null;
+function getSpaPlaceHolders(): NodeListOf<HTMLElement> | null {
+    return document.getElementsByName("__dot_SpaContentPlaceHolder");
 }
 
-export function getSpaPlaceHolderUniqueId(): string {
-    return getSpaPlaceHolder()!.getAttribute("data-dotvvm-spacontentplaceholder")!;
+export function getSpaPlaceHoldersUniqueId(): string {
+    const spas = Array.from(getSpaPlaceHolders()!);
```
exyi, line 27: Remove `!`, it can not be null.

exyi (approved):

PR #812, opened 2020-03-14
[Do not merge] Added tracing for application startup

tomasherceg:
I have added a simple mechanism to trace startup events and send them along with request performance traces.

I thought that the slow startup of DotVVM apps in Azure is caused by `InvokeAllStaticConstructors` that loads all assemblies, but it looks that the slow-down is earlier than this and that it happens in `DotvvmStartup.Configure`.

This PR doesn't solve the issues, it just gathers timings from the startup code in order to provide a foundation for further analysis.

The results are from `Dotvvm.Samples.BasicSamples.Owin` and `Dotvvm.Samples.BasicSamples.AspNetCore`.
* `Total` means the number of milliseconds since the app startup (to be precise, from the very first entry into DotVVM code).
* `Duration` represents the number of milliseconds since the previous event.

## Local - OWIN

```
Event Name                                                              Duration [ms]     Total [ms]
----------------------------------------------------------------------------------------------------
AddDotvvmStarted                                                                 0 ms           0 ms
DotvvmConfigurationUserServicesRegistrationStarted                              12 ms          12 ms
DotvvmConfigurationUserServicesRegistrationFinished                              5 ms          17 ms
DotvvmConfigurationUserConfigureStarted                                         58 ms          75 ms
DotvvmConfigurationUserConfigureFinished                                       617 ms         692 ms
UseDotvvmStarted                                                                 7 ms         699 ms
UseDotvvmFinished                                                                6 ms         705 ms
InvokeAllStaticConstructorsStarted                                             325 ms        1030 ms
InvokeAllStaticConstructorsFinished                                            142 ms        1172 ms
```

## Local - ASP.NET Core

```
Event Name                                                              Duration [ms]     Total [ms]
----------------------------------------------------------------------------------------------------
AddDotvvmStarted                                                                 0 ms           0 ms
DotvvmConfigurationUserServicesRegistrationStarted                              23 ms          23 ms
DotvvmConfigurationUserServicesRegistrationFinished                              3 ms          26 ms
DotvvmConfigurationUserConfigureStarted                                        202 ms         228 ms
DotvvmConfigurationUserConfigureFinished                                       431 ms         659 ms
UseDotvvmStarted                                                                 4 ms         663 ms
UseDotvvmFinished                                                               10 ms         673 ms
InvokeAllStaticConstructorsStarted                                             724 ms        1397 ms
InvokeAllStaticConstructorsFinished                                             89 ms        1486 ms
```

## Azure App Service - OWIN

```
[...]

PR #850, opened 2020-06-14
Explicit assembly loading

tomasherceg:
Since PR #812 contained a lot of changes and was difficult to merge, I've extracted part of the changes as a separate PR.

* Some static methods (`FindType`, `GetAllAssemblies` etc.) were moved from `ReflectionUtils` to `CompiledAssemblyCache` and changed to be instance methods. The logic inside these methods was not changed.
* Added experimental flag `ExplicitAssemblyLoading` which disables aggressive loading of all referenced assemblies (including transitive references) and uses only assemblies explicitly listed in `config.Markup.Assemblies`.
* Added simple performance tracking mechanism `IStartupTracer` for the DotVVM startup code.
* Since various tests and custom JSON converters relied on static `ReflectionUtils`, there is `CompiledAssemblyCache.Instance` which contains a singleton assembly cache. In tests, it is possible that multiple instances of this class are created, which is unfortunate, but tolerable. In real DotVVM web apps, there should be always only instance created by DI.


exyi:
> Since various tests and custom JSON converters relied on static ReflectionUtils, there is CompiledAssemblyCache.Instance which contains a singleton assembly cache. ... In real DotVVM web apps, there should be always only instance created by DI.

No, the reason why we have everything in the DI is to allow multiple instances of DotVVM (and entire AspNetCore in a process). This approach will lead to collisions and fail horribly if I'd run multiple servers in the process.

However, it will be mostly fine to register all allowed assemblies in a single "cache" - have a union of both configurations.

exyi:
The name `CompiledAssemblyCache` is terrible for a class that is used everywhere for doing reflection. I think should named SomethingContext, maybe DotvvmReflectionContext, but I don't know. But it is certainly not a cache of compiled assemblies, how **"compiled"**?

tomasherceg:
> The name `CompiledAssemblyCache` is terrible for a class that is used everywhere for doing reflection. I think should named SomethingContext, maybe DotvvmReflectionContext, but I don't know. But it is certainly not a cache of compiled assemblies, how **"compiled"**?

I am not sure why it was named that - it is quite old. Basically it maintains a list of assemblies that are used in Roslyn compilation, and caches Roslyn MetadataReferences. 
`DotvvmReflectionContext` seems fine to me, or maybe `DotvvmCompilationContext`.

tomasherceg:
> No, the reason why we have everything in the DI is to allow multiple instances of DotVVM (and entire AspNetCore in a process). This approach will lead to collisions and fail horribly if I'd run multiple servers in the process.
Is there a way how to get things from DI from JsonConverter attributes?https://github.com/riganti/dotvvm/pull/850/files#diff-216e10f9b6e51de337bb86e303e2c5a4R19

Running two instances of DotVVM in a single app would be a nice thing, but I'm not sure how many people would even consider doing this.

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
```diff
@@ -11,6 +24,79 @@ public class CompiledAssemblyCache
...
+                Instance = this;
+            }
+        }
+
+        private IEnumerable<Assembly> BuildReferencedAssembliesCache()
+        {
+            var diAssembly = typeof(ServiceCollection).Assembly;
+
+            var references = diAssembly.GetReferencedAssemblies().Select(Assembly.Load)
+                .Concat(configuration.Markup.Assemblies.Select(e => Assembly.Load(new AssemblyName(e))))
```
exyi, line 58: Shall we load the referenced assemblies recursively?

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
```diff
@@ -11,6 +24,79 @@ public class CompiledAssemblyCache
...
+#if DotNetCore
+                    Assembly.Load(new AssemblyName("System.Runtime")),
+                    Assembly.Load(new AssemblyName("System.Collections.Concurrent")),
+                    Assembly.Load(new AssemblyName("System.Collections")),
+#else
+                    typeof(List<>).Assembly
+#endif
+                });
+
+            var netstandardAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name == "netstandard");
```
exyi, line 76: Hmm, so what is the point when we get all assemblies anyway here? Oo

src/DotVVM.Framework/Compilation/DefaultViewCompiler.cs
```diff
@@ -104,62 +99,13 @@ protected virtual CSharpCompilation AddToCompilation(CSharpCompilation compilati
...
-            return references.Distinct().ToList();
+                    .Select(a => assemblyCache.GetAssemblyMetadata(a.Key).WithAliases(ImmutableArray.Create(a.Value, "global"))));
         }
-
+        
         public virtual CSharpCompilation CreateCompilation(string assemblyName)
         {
             return CSharpCompilation.Create(assemblyName, options: new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary))
-                .AddReferences(referencedAssembliesCache.Value.Select(a => assemblyCache.GetAssemblyMetadata(a)));
+                .AddReferences(assemblyCache.GetAllAssemblies().Select(a => assemblyCache.GetAssemblyMetadata(a)));
```
exyi, line 108: This used to reference only some assemblies and now we reference everything when the experimental option is disabled, right? I don't know if that will break, but I'd guess there is a reson why the logic used to be so complex.

src/DotVVM.Framework/Compilation/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -167,7 +169,7 @@ public string EmitCreateObject(string typeName, object[] constructorArguments =
             }
 
             var typeSyntax = ReflectionUtils.IsFullName(typeName)
-                ? ParseTypeName(ReflectionUtils.FindType(typeName))
+                ? ParseTypeName(compiledAssemblyCache.FindType(typeName))
```
exyi, line 172: When I see this, the type parsing should not be done by emitter, can we move it to the caller?

src/DotVVM.Framework/Compilation/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -769,7 +771,7 @@ public IEnumerable<SyntaxTree> BuildTree(string namespaceName, string className,
             UseType(BuilderDataContextType);
 
             var controlType = ReflectionUtils.IsFullName(ResultControlType)
-                ? ParseTypeName(ReflectionUtils.FindType(ResultControlType))
+                ? ParseTypeName(compiledAssemblyCache.FindType(ResultControlType))
```
exyi, line 774: Same here, type parsing should not be done by emitter.

src/DotVVM.Framework/Configuration/DotvvmGlobalExperimentalFeatureFlag.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                ThrowIfFrozen();
+                _enabled = value;
+            }
+        }
+        private bool _enabled = false;
+
+        public void Enable()
+        {
+            ThrowIfFrozen();
```
exyi, line 23: This is redundant.

src/DotVVM.Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
```diff
@@ -36,7 +37,7 @@ protected virtual IEnumerable<Type> ResolveAllTypesDerivedFromIResource(string d
 
         protected virtual IEnumerable<Assembly> GetAllAssembliesLoadedAssemblies()
         {
-            return ReflectionUtils.GetAllAssemblies();
+            return AppDomain.CurrentDomain.GetAssemblies();
```
exyi, line 40: It seems to be unused, anyway. I don't know why it's virtual, though.

exyi:
> Running two instances of DotVVM in a single app would be a nice thing, but I'm not sure how many people would even consider doing this.

I did. It is what we do in the benchmarks. It is also a convenient way to run two separate apps that are on different domains in a single process (like admin + the actual app).

tomasherceg:
> I did. It is what we do in the benchmarks. It is also a convenient way to run two separate apps that are on different domains in a single process (like admin + the actual app).

Cool, good to know.

src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -11,11 +11,15 @@
...
 using DotVVM.Framework.Runtime;
 using DotVVM.Framework.Hosting.AspNetCore;
+using DotVVM.Framework.Diagnostics;
+using DotVVM.Framework.Runtime.Tracing;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static class ServiceCollectionExtensions
     {
+        private static readonly DiagnosticsStartupTracer startupTracer = new DiagnosticsStartupTracer();
```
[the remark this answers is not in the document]
tomasherceg: It is added to the `IServiceCollection` container later (at the end of the file). 
In ASP.NET Core, it needs to be a static field because we need it before the `IServiceCollection` is ready and because it is called from both `ConfigureServices` and `Configure` methods in `Startup`.
In OWIN it can be just a local variable as there is only one method for initialization of DotVVM.

src/DotVVM.Framework/DependencyInjection/DotvvmBuilderExtensions.cs
```diff
@@ -94,12 +94,19 @@ public static IDotvvmServiceCollection AddDiagnosticServices(this IDotvvmService
...
-            services.Services.AddScoped<DiagnosticsRequestTracer>(s => new DiagnosticsRequestTracer(s.GetRequiredService<IDiagnosticsInformationSender>()));
+
+            services.Services.AddScoped<DiagnosticsRequestTracer>();
             services.Services.AddScoped<IRequestTracer>(s => {
                 var config = s.GetRequiredService<DotvvmConfiguration>();
                 return (config.Debug ? (IRequestTracer)s.GetService<DiagnosticsRequestTracer>() : null) ?? new NullRequestTracer();
             });
 
+            services.Services.AddScoped<DiagnosticsStartupTracer>();
+            services.Services.AddScoped<IStartupTracer>(s => {
```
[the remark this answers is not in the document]
tomasherceg: I removed this - I haven't noticed that this method is meant for the Diagnostic Window feature in VS extension that is not finished and probably doesn't work.

src/DotVVM.Framework/Diagnostics/DiagnosticsStartupTracer.cs
```diff
@@ -0,0 +1,95 @@
...
+            startupCompleted = true;
+
+            // create a timer to report after-startup events regularly
+            this.timer = new Timer(state =>
+            {
+                lock (locker)
+                {
+                    if (sentEventsCount < events.Count)
+                    {
+                        informationSender.SendInformationAsync(BuildDiagnosticsInformation(events.Skip(sentEventsCount).ToList()));
```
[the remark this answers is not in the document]
tomasherceg: It is a loop so it would never end. Or maybe I don't understand.

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
tomasherceg: `GetAssemblies` doesn't load the assemblies, but it looks like that this line is not needed anyway.

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
tomasherceg: Not sure - the only related issue I am aware of was with calling the static constructors (the static constructors on base types are not called automatically) which I'll handle elsewhere. But I think we don't need to explicitly load referenced assemblies.

src/DotVVM.Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
tomasherceg: Removed together with `ResolveAllTypesDerivedFromIResource` that was unused too.

src/DotVVM.Framework/Compilation/DefaultViewCompiler.cs
tomasherceg: Good catch - I've distinguished between the referenced assemblies and "all" assemblies, and work only with referenced assemblies here.

src/DotVVM.Framework.Hosting.Owin/AppBuilderExtensions.cs
```diff
@@ -56,7 +57,11 @@ public static DotvvmConfiguration UseDotVVM<TStartup>(this IAppBuilder app, stri
 
         private static DotvvmConfiguration UseDotVVM(this IAppBuilder app, string applicationRootPath, bool useErrorPages, bool debug, IDotvvmServiceConfigurator configurator, IDotvvmStartup startup, Func<IServiceCollection, IServiceProvider> serviceProviderFactoryMethod = null, Action<DotvvmConfiguration> modifyConfiguration = null)
         {
-            var config = DotvvmConfiguration.CreateDefault(s => {
+            var startupTracer = new DiagnosticsStartupTracer();
```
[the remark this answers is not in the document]
tomasherceg: This one is correct, I removed the other one.

src/DotVVM.Framework/Compilation/Binding/TypeRegistry.cs
```diff
@@ -10,15 +10,15 @@ namespace DotVVM.Framework.Compilation.Binding
     public class TypeRegistry
     {
         ImmutableDictionary<string, Expression> registry;
-        ImmutableList<Func<string, Expression>> resolvers;
+        ImmutableList<Func<string, CompiledAssemblyCache, Expression>> resolvers;
```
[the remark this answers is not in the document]
tomasherceg: Fixed, hopefully it is now better - had to change some static method to instance ones.

src/DotVVM.Framework/Compilation/ViewCompilingVisitor.cs
```diff
@@ -39,21 +42,25 @@ public override void VisitView(ResolvedTreeRoot view)
 
             var createsCustomDerivedType = view.Metadata.Type == typeof(DotvvmView);
 
+            string resultControlType;
             if (createsCustomDerivedType)
             {
-                emitter.ResultControlType = className + "Control";
-                emitter.EmitControlClass(view.Metadata.Type, emitter.ResultControlType);
+                resultControlType = className + "Control";
```
exyi, line 48: In this case, `resultControlType` is always relative name, so we will end up using expression `SyntaxFactory.ParseTypeName(resultControlType)`.

src/DotVVM.Framework/Compilation/ViewCompilingVisitor.cs
```diff
@@ -39,21 +42,25 @@ public override void VisitView(ResolvedTreeRoot view)
...
+                emitter.EmitControlClass(view.Metadata.Type, resultControlType);
             }
             else
-                emitter.ResultControlType = view.Metadata.Type.FullName;
-
+            {
+                resultControlType = view.Metadata.Type.FullName;
```
exyi, line 53: In this case, the type is always a full name. It would be simpler to call `emitter.ParseTypeName(view.Metadata.Type.)` directly instead of finding the type in the compilation again.

src/DotVVM.Framework/Diagnostics/DiagnosticsStartupTracer.cs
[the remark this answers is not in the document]
exyi: Ok, I don't understand if the timer fires only once or every 10sec. In any case, why not use `Task.Delay` in an async method? We could await the task returned from `SendInformationAsync`

tomasherceg:
> Ok, I don't understand if the timer fires only once or every 10sec. In any case, why not use Task.Delay in an async method? We could await the task returned from SendInformationAsync

I've got rid of the timer completely. When the DotVVM initialization ends, the events are sent using the `DiagnosticsInformationSender`. In case of additional events reported after this sending, the `LateInfoReported` event is triggered and there is a handler which sends these late events. And it returns task in canse anyone is interested.

src/DotVVM.Samples.Common/TextFileDiagnosticsInformationSender.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var messages = FormatUnwrittenMessages(information);
+
+            if (!File.Exists(logFilePath))
+            {
+                messages = $@"{"Event Name",-70}{"Duration [ms]",15}{"Total [ms]",15}
+{new string('-', 70 + 15 + 15)}
+{messages}";
+            }
+            File.AppendAllText(logFilePath, messages);
```
exyi, line 35: This may run in parallel, we should have a lock around this line (I know it's just samples, but still)

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
```diff
@@ -24,12 +102,44 @@ private Assembly DefaultOnResolving(System.Runtime.Loader.AssemblyLoadContext as
...
+                return DependencyContext.Default.GetDefaultAssemblyNames().Select(Assembly.Load).ToArray();
+#else
+                // this doesn't load new assemblies, but it is done in InvokeStaticConstructorsOnAllControls
+                return AppDomain.CurrentDomain.GetAssemblies();
 #endif
+            }
+        }
+
+        public bool IsAssemblyNamespace(string fullName)
+            => GetAllNamespaces().Contains(fullName, StringComparer.Ordinal);
```
exyi, line 135: Hmm, I know this is just copy paste from the old utils. However, it would make more sense to create the `HashSet` once with the `Ordinal` comparer, and then cache it in a `Lazy` field.

src/DotVVM.Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
quigamdev: Compiler overrides this method for its purposes.

src/DotVVM.Framework/Compilation/CompiledAssemblyCache.cs
tomasherceg: Fixed.

src/DotVVM.Samples.Common/TextFileDiagnosticsInformationSender.cs
tomasherceg: You're right - fixed.

PR #853, opened 2020-06-16
Serialization: KeyValuePair and ValueTuple are handled by Newtonsoft.Json

### LINKED ISSUE #829: Collection of KeyValuePair cannot be deserialized
There is a bug when a public property of type `List<KeyValuePair<int, string>>` is used in ViewModel, then deserialization of this property fails on postback with exception `Expression of type 'System.Collections.Generic.KeyValuePair2[System.Int32,System.String]' cannot be used for return type 'System.Object`.
Serialization and sending the serialized view model to the client works.

exyi:
Also, I have deduplicated the code that checks if we can serialize a value using our converter.

resolves #829


exyi:
@tomasherceg I have added test for a viewmodel in the tuple. Could you please review it?

PR #857, opened 2020-06-30
Added CancellationToken to IDotvvmRequestContextExtensions

### LINKED ISSUE #810: Add CancellationToken to IDotvvmRequestContext
it would be nice to add CancellationToken to IDotvvmRequestContext. Similiar concept has a asp.net core mvc. 

It'll be a helper for `Context.GetAspNetCoreContext().RequestAborted`

acizmarik:
This PR adds `CancellationToken` to `IDotvvmRequestContextExtensions`. Resolves #810

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
@@ -28,6 +28,14 @@ public static class DotvvmRequestContextExtensions
         return DotvvmPresenter.DetermineSpaContentPlaceHolderUniqueId(context.HttpContext);
     }
 
+    /// <summary>
+    /// Gets cancellation token for the request
+    /// </summary>
+    public static CancellationToken GetCancellationToken(this IDotvvmRequestContext context)
+    {
+        var cancellationService = context.Services.GetService<IRequestCancellationTokenProvider>();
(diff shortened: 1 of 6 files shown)


src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -28,6 +28,14 @@ public static string GetSpaContentPlaceHolderUniqueId(this IDotvvmRequestContext
         return DotvvmPresenter.DetermineSpaContentPlaceHolderUniqueId(context.HttpContext);
     }
 
+    /// <summary>
+    /// Gets cancellation token for the request
+    /// </summary>
+    public static CancellationToken GetCancellationToken(this IDotvvmRequestContext context)
+    {
+        var cancellationService = context.Services.GetService<IRequestCancellationTokenProvider>();
+        return cancellationService.GetCancellationToken(context);
```
exyi, line 37: We should either call `GetRequiredService` or return `default(CancellationToken)` when the service is null. This may produce a NullReferenceException when the `IRequestCancellationTokenProvider` is missing

acizmarik: Fixed by calling the `GetRequiredService`

PR #858, opened 2020-07-02
Added helper methods for redirection

### LINKED ISSUE #462: RouteTable.AddRedirection helper method
Often, the URLs on a website change, but you need to maintain compatibility with the old system. It could be quite helpful to have a helper method, that would simply register a route with a redirecting presenter. The usage could look like that:
```C#
routes.AddRedirection("article/{Id}/edit", "admin/article-edit/{Id}");
```
or redirection to route name:
```
routes.AddRouteRedirection("article/{Id}/edit", "Admin_EditArticle");
```

The redirection should be permanent by default, but there should be a optional `permanent: bool = true` parameter. And the name of the route is basically irrelevant (you probably don't want RouteLinks to a redirecing route), so they should also be optional.

acizmarik:
This PR adds helper methods for redirection. Resolves #462.
Compared to the original proposal in #462, we decided to implement the following changes after an internal discussion:

- `routeName` will be always mandatory. This is due to the fact that this property is used internally as a key in a lookup table. We can change this behaviour in future releases.
- `permanent` is an optional parameter with its default value set to `false`.


tomasherceg:
I think that we should add an optional parameter to `AddRouteRedirection` that will allow transforming the route parameters, and maybe an overload of `AddUrlRedirection` that can compose the URL dynamically. 

```
void AddUrlRedirection(string routeName, string from, Func<Dictionary<string, object>, string> targetUrlProvider, object? defaultValues = null, bool permanent = false)

void AddRouteRedirection(string routeName, string from, string targetRoute, object? defaultValues = null, bool permanent = false, Func<Dictionary<string, object>, Dictionary<string, object>> parameterProvider)
```

Also, I'd rename `from` to `urlPattern`, `toRoute` to `targetRoute` and `toUrl` to `targetUrl`.

src/DotVVM.Framework/Routing/RedirectingPresenter.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using DotVVM.Framework.Hosting;
+
+namespace DotVVM.Framework.Routing
+{
+    public class RedirectingPresenter : IDotvvmPresenter
```
exyi, line 10: I understand the presenter is used for redirection, but it actually does not do any redirect on itself. Let's not bikeshed about a better name and make the class internal instead.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -131,6 +135,49 @@ public void Add(string routeName, string? url, Func<IServiceProvider, IDotvvmPre
...
+        /// <summary>
+        /// Adds the specified URL redirection entry.
+        /// </summary>
+        /// <param name="routeName">Name of the redirection.</param>
+        /// <param name="urlPattern">URL pattern to redirect from.</param>
+        /// <param name="urlProvider">URL provider to obtain context-based redirection targets.</param>
+        public void AddUrlRedirection(string routeName, string urlPattern, Func<IDotvvmRequestContext, string> urlProvider, object? defaultValues = null, bool permanent = false)
+        {
+            ThrowIfFrozen();
+            Func<IServiceProvider, IDotvvmPresenter> presenterFactory = (serviceProvider) => GetRedirectionPresenter((context) =>
```
exyi, line 147:
```suggestion
            IServiceProvider presenterFactory(IDotvvmPresenter serviceProvider) => GetRedirectionPresenter((context) =>
```

You can use a local function, that is a bit less verbose.

src/DotVVM.Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -131,6 +135,49 @@ public void Add(string routeName, string? url, Func<IServiceProvider, IDotvvmPre
...
+        /// Adds the specified route redirection entry.
+        /// </summary>
+        /// <param name="routeName">Name of the redirection.</param>
+        /// <param name="urlPattern">URL pattern to redirect from.</param>
+        /// <param name="routeNameProvider">Route name provider to obtain context-based redirection targets.</param>
+        public void AddRouteRedirection(string routeName, string urlPattern, Func<IDotvvmRequestContext, string> routeNameProvider,
+            object? defaultValues = null, bool permanent = false, Func<IDictionary<string, object?>, Dictionary<string, object?>>? parametersProvider = null)
+        {
+            ThrowIfFrozen();
+            Func<IServiceProvider, IDotvvmPresenter> presenterFactory = (serviceProvider) => GetRedirectionPresenter((context) =>
```
exyi, line 169:
```suggestion
            IServiceProvider presenterFactory(IDotvvmPresenter serviceProvider) => GetRedirectionPresenter((context) =>
```

You can use a local function, that is a bit less verbose.

src/DotVVM.Framework/Routing/RedirectingPresenter.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using DotVVM.Framework.Hosting;
+
+namespace DotVVM.Framework.Routing
+{
+    internal class RedirectingPresenter : IDotvvmPresenter
```
tomasherceg, line 10: Change to `public`
Rename to `DelegatePresenter` since it's a generic presenter implementation

PR #872, opened 2020-08-14
Command-line interface

cafstep:
(Builds off of #775.)

This is a major-ish rewrite of the command-line interface and the stand-alone compiler executable. It tries to bring them closer to style of the SDK's dotnet executable.

The Compiler executable is no longer embedded into the DotVVM NuGet package anymore. Instead, DotVVM.CommandLine itself includes two builds of the Compiler, one for .NET Framework and one for .NET Core (and .NET 5).

In addition to that, the Compiler is now able to load the compiled project's assembly correctly (i.e. with all its dependencies) using AppDomain and DependencyContext magics.

However, not everything has been solved. The Compiler is still not actually a compiler. It's more of a linter really, since even though it can create an assembly for each page it can't merge them together. None of the IL merging tools I tried work, so I conclude that to get this to work, we'll have to make some changes in DotVVM.Framework.

The process of the CLI's `lint` command goes a little something like this:
1. Find a .csproj
2. Using a custom .targets file, figure out what framework does the .csproj target
3. Build the .csproj, unless the --no-build options was passed
4. Pick the proper build of the compiler (i.e. net461 or netcoreapp3.1)
5. Invoke the compiler with a path to the project and its built assembly file
  a. If this is the netcoreapp3.1 Compiler, use DependencyContext and AssemblyLoadContext to load the assembly with its dependencies.
  b. If this is the net461 Compiler, create a new AppDomain with the project's app.config (thus loading its dependencies first). If an assembly resolution fails, the original AppDomain is used in the AssemblyResolve event.
  c. Create a DotvvmConfiguration
  d. Compile all views and markup controls more-or-less the same way DotVVM does at runtime.


cafstep:
* Add api client regen back.
* Add command to migrate from .dotvvm.json.
* Move bulk of the Compiler code to Framework, so that Compiler is just a wrapper.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+            return ConfigurationInitializer.GetConfiguration(dotvvmProjectAssembly, dotvvmProjectDir, services =>
+            {
+                services.AddSingleton<IControlResolver, StaticViewControlResolver>();
+                services.TryAddSingleton<IViewModelProtector, FakeViewModelProtector>();
+                services.AddSingleton(new RefObjectSerializer());
+                // Yes, this is here so that there can be a circular dependency in StaticViewControlResolver.
+                // I'm not happy about it, no, but the alternative is a more-or-less complete rewrite.
+                services.AddSingleton(p => new StaticViewCompiler(
+                    p.GetRequiredService<DotvvmConfiguration>(),
+                    dotvvmProjectAssembly));
```
exyi, line 70: How is DotvvmConfiguration dependent on `StaticViewCompiler`?

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+
+                // HACK: IDotvvmCacheAdapter is not in v2.0.0 that's why it's hacked this way.
```
exyi, line 72: Actually, this is why we had Dotvvm.Compiler built into the dotvvm package. This way, we will either have to be binary forward compatible on the interface (which is documented?) or have a lot of cases like this.

Can we test for the compatibility? I guess we will have to test the current version against all recent DotVVM versions.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+        public StaticView GetView(string viewPath)
+        {
+            if (viewCache.ContainsKey(viewPath))
+            {
+                return viewCache[viewPath];
+            }
+
+            var view = CompileView(viewPath);
+            // NB: in the meantime, the view could have been compiled on another thread
+            return viewCache.GetOrAdd(viewPath, view);
```
exyi, line 99: Why not `viewCache.GetOrAdd(viewPath, CompileView)`?

Also, you might want to use `ConcurrentDictionary<..., Lazy<...>>` to prevent parallel compilation of one view (it is fairly expensive, so it seems worth it to me)

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+
+        private ImmutableArray<StaticView> CompileViews(IEnumerable<string?> paths)
+        {
+            return paths
+                .Where(p => !string.IsNullOrWhiteSpace(p))
+                .Select(p => GetView(p!))
+                .ToImmutableArray();
+        }
+
+        private StaticView CompileView(string viewPath)
```
exyi, line 117: Oh, this looks like a partially broken duplication of `DefaultViewCompiler` - could we merge the logic or is it unused code anyway? In particular, this probably breaks on recursive markup controls, and it does not run all visitors.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+            //       which is bollocks.
+            if (resolvedView.Metadata.Type != typeof(DotvvmView))
+            {
+                emitter.ResultControlTypeSyntax = emitter.ParseTypeName(resolvedView.Metadata.Type);
+            }
+
+            if (resolvedView.Directives.ContainsKey("masterPage"))
+            {
+                // make sure that the masterpage chain is already compiled
+                _ = GetView(resolvedView.Directives["masterPage"].Single().Value);
```
exyi, line 212: In the coming version (#834), this won't be needed.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
```diff
@@ -0,0 +1,277 @@
...
+            return view.WithAssembly(assembly)
+                .WithViewType(resolvedView.Metadata.Type)
+                .WithDataContextType(resolvedView.DataContextTypeStack.DataContextType);
+        }
+
+        /// <summary>
+        /// HACK: Because as of 2.4.0, DotVVM gets a list of all assemblies only from the Default DependencyContext,
+        ///       a problem arises, because in this Compiler, DotVVM itself isn't in the Default DependencyContext, thus
+        ///       I need to invoke the static constructors of controls myself.
+        /// </summary>
```
exyi, line 242: You claim to be invoking constructors on the Dotvvm.Framework assembly, but it actually runs though many more assemblies, right?

exyi (commented):
It seems near impossible to review the entire PR, so I only commented on the files in DotVVM.Framework

.gitignore
```diff
@@ -125,7 +125,7 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
-# TODO: Comment the next line if you want to checkin your web deploy settings 
+# TODO: Comment the next line if you want to checkin your web deploy settings
```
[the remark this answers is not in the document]
cafstep: Not sure, why this even happened. Probably VSCode's overzealous line trimming.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
cafstep: There was the same circular dependency as in `DefaultControlBuilderFactory`. Since `StaticViewCompiler` no longer duplicates `DefaultViewCompiler` this is resolved.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
cafstep: This is why I moved the bulk of DotVVM.Compiler to Framework. This way the binary interface that actually needs to remain forward-compatible is limited to the `StaticViewCompiler.CompileAll` method and the `CompilationReport` class. This should not be a problem for us, I hope.

This of course means, that this version of the "compiler" (still just a linter) is absolutely unusable for any version of DotVVM prior to 3.0.0 (or whenever this PR makes it).

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
cafstep: Fixed by virtue of the `DefaultControlBuilderFactory` doing it this way.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
cafstep: I agree that the duplication was really bad, so I removed it. However, not the linter is limited to one parser error per file. We should look into that at some point.

src/DotVVM.Framework/Compilation/Static/StaticViewCompiler.cs
cafstep: I removed this particular hack in favor of a better one (or a much worse one, depending on your point of view) in `DependencyContextCompilerExecutor`.

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
```diff
@@ -100,6 +101,8 @@ public static IServiceCollection RegisterDotVVMServices(IServiceCollection servi
                 o.TreeVisitors.Add(() => new LifecycleRequirementsAssigningVisitor());
             });
 
+            services.TryAddSingleton<IDotvvmCacheAdapter, DefaultDotvvmCacheAdapter>();
```
cafstep, line 104: I moved this line from the aspnetcore-specific and owin-specific ServiceCollection extensions, since both `IDotvvmCacheAdapter` and `DefaultDotvvmCacheAdapter` are in Framework. Not sure if this is an issue. @exyi, is this an issue?

src/DotVVM.Framework/Configuration/DotvvmConfiguration.cs
```diff
@@ -269,6 +270,18 @@ public static DotvvmConfiguration CreateDefault(IServiceProvider serviceProvider
...
+        /// </summary>
+        internal static DotvvmConfiguration CreateInternal()
+        {
+            return CreateDefault(services =>
+            {
+                services.TryAddSingleton<IViewModelProtector, FakeViewModelProtector>();
+                services.TryAddSingleton<ICsrfProtector, FakeCsrfProtector>();
+            });
+        }
+
```
cafstep, line 284: We have methods like this all over the place. I think it's time we added it into Framework itself.

src/DotVVM.Framework/DependencyInjection/DotVVMServiceCollectionExtensions.cs
quigamdev: This should be ok. What's new for me is I didn't know that DefaultDotvvmCacheAdapter is used as the default for both platforms.

src/DotVVM.Framework/Configuration/DotvvmConfiguration.cs
quigamdev: It sounds legit.

quigamdev:
Finally, I have tested the compiler against the net461 framework and it works. Also, I have checked all files that are directly part of DotVVM.Framework or DotVVM.Core assembly. All seem good to me.

quigamdev (approved):

PR #876, opened 2020-08-19
Deny postbacks during SPA navigation + fixed SPA navigation for default page

### LINKED ISSUE #873: SPA redirect didn't change url to empty
SPA redirect didn't change url if is empty

Repro steps
1. Create DotVVM sample project
1. Add SecondPage.dothtml with `<dot:RouteLink RouteName="Default" Text="Home" />` 
1. Use `SpaContentPlaceHolder` in masterpage
1. Redirect from SecondPage to home. Redirect is successful but url is /SecondPage instead of /

```
private void ConfigureRoutes(DotvvmConfiguration config, string applicationPath)
{
    config.RouteTable.Add("Default", "", "Views/Default.dothtml");
    config.RouteTable.AutoDiscoverRoutes(new DefaultRouteStrategy(config));    
}
```

tomasherceg:
We've discovered that SPA navigation is cancelled when the user issues a postback or a static command (with server-side call) during the navigation because the same postback counter is used.
To fix the issue, the postback and static commands are now disabled when the SPA navigation is in progress (and restored when it fails). The solution is not very nice and I'd like to discuss it with @exyi.

Also, this PR fixes #873 which caused the old URL stay in the address bar when navigating back from some page to the default one (`/`).

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/DotVVM.SpaHistory.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/DotVVM.SpaHistory.ts
@@ -5,6 +5,9 @@
 class DotvvmSpaHistory {
 
     public pushPage(url: string) {
+        // pushState doesn't work when the url is empty
+        url = url || "/";
+
         history.pushState(new HistoryRecord('SPA', url), '', url);
     }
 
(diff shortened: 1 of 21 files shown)


exyi (approved):
Good hack :)

PR #879, opened 2020-08-24
Resource optimization

MichalTichy:
- Added option to specify whether script is async or defer
- Added preload support for stylesheets


exyi (changes requested):
Defer/Async:

1. This breaks dependency resolution - `async` script must have no dependencies and nothing must depend on it. Everything that depends on deferred scripts must be also `defer`red.
2. On v3, scripts are `defer`red by default. Do you have use case for `async`?

Preload for styles:

1. Why, though, when styles are in the header anyway?
2. I don't think it will be rendered - only preloads for `ResourceRenderPosition.Body` show up.

exyi:
@MichalTichy, if there is no interest in this change or discussion, I'm closing this. Defer is implemented anyway in v3

PR #880, opened 2020-08-25
Fixed inline editing for nested GridViews

### LINKED ISSUE #864: Issue with InlineEditing in nested GridViews
One of our Dev Advocates found out that the ` InlineEditing` feature doesn't work in nested `GridView` controls.
Here is a repo with the problem: https://github.com/miroprocessor/dotvvm-nestedgrids

acizmarik:
This PR fixes inline editing for nested `GridView`s. Resolves #864.


tomasherceg (approved):

src/DotVVM.Framework/Configuration/UserColumnMappingCache.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading.Tasks;
+using DotVVM.Framework.Controls;
+using DotVVM.Framework.ViewModel;
+
+namespace DotVVM.Framework.Configuration
+{
+    public class UserColumnMappingCache
+    {
+        private ConcurrentDictionary<Type, Dictionary<string, string>> mappingCache;
+        private IPropertySerialization propertySerialization;
```
exyi, line 15: `IPropertySerialization` should not be used to figure out the property names. It can be renamed later using this API https://github.com/riganti/dotvvm/pull/262.

You should call `IViewModelSerializationMapper.GetMap` to get the real mapping of all properties.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -1447,7 +1447,18 @@ class DotVVM {
...
-                (bindingContext, value) => bindingContext.extend({ $gridViewDataSet: value, [foreachCollectionSymbol]: dotvvm.evaluator.getDataSourceItems(value) }),
+                (bindingContext, value, allBindings) => bindingContext.extend({
+                    $gridViewDataSet: value,
+                    [foreachCollectionSymbol]: dotvvm.evaluator.getDataSourceItems(value),
+                    $gridViewDataSetHelper: {
+                        columnMapping: allBindings.get("gridViewDataSetColumnMapping"),
+                        isInEditMode: function ($context) {
+                            let columnName = $context.$gridViewDataSet().RowEditOptions().PrimaryKeyPropertyName();
+                            columnName = this.columnMapping[columnName] || columnName;
+                            return ko.unwrap($context.$gridViewDataSet().RowEditOptions().EditRowId) === ko.unwrap($context.$data[columnName]);
```
exyi, line 1458: Is there a case when `EditRowId` is not observable? That would clearly be a bug - we should not use `ko.unwrap` here

src/DotVVM.Framework/Configuration/UserColumnMappingCache.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public UserColumnMappingCache(IPropertySerialization propertySerialization)
+        {
+            this.propertySerialization = propertySerialization;
+            this.mappingCache = new ConcurrentDictionary<Type, Dictionary<string, string>>();       
+        }
+
+        public Dictionary<string, string> GetMapping(Type type)
```
exyi, line 23: We should return `IReadOnlyDictionary` to prevent mutation of the cached value.

src/DotVVM.Framework/Resources/Scripts/DotVVM.ts
```diff
@@ -1408,7 +1408,7 @@ class DotVVM {
                     if (!isInitial) {
                         ko.virtualElements.setDomNodeChildren(element, ko.utils.cloneNodes(savedNodes!));
                     }
-                    ko.applyBindingsToDescendants(makeContextCallback(bindingContext, rawValue), element);
+                    ko.applyBindingsToDescendants(makeContextCallback(bindingContext, rawValue, _allBindings), element);
```
exyi, line 1411: `_allBindings` was named with the underscore since the parameter was unused. As we use it now, the name should not start with underscore.

exyi (changes requested):

exyi (approved):

PR #883, opened 2020-09-04
Migrate JS tests to Jest

exyi:
Tests are run in the `DotVVM.Framework` project by `yarn jest`, if you'd like to see code coverage run `yarn jest --coverage` (it's going to end up in `coverage` directory).

This PR migrates the old tests to the new infrastructure and also adds a few new tests. Notably, I have added a [fast-check library](https://github.com/dubzzz/fast-check) that does random testing. It is used in the postback tests to schedule concurrent postbacks randomly and try to hit same races.


tomasherceg (approved):
Nice one!

PR #884, opened 2020-09-04
Use immutable object instead of knockout observables

exyi:
This PR changes the way we store viewmodel on the client-side. We used to have everything in a hierarchy of knockout observables, so knockout could update the UI when something changed. It was however cumbersome to work with in code that did not want to care about knockout.

With this change, we will isolate the knockout observable hierarchy from an immutable view model that is easier to work with. We automatically synchronize changes both ways:

* changes from knockout observables are immediately propagated into the immutable state
* changes in the state are propagated to knockout on the next animation frame or when stateManager.doUpdateNow is invoked. This has a potential to speed up scenarios when data is changed often, because every change in ko.observable immediately leads to DOM updates.

The observable hierarchy is created lazily, so we have some potential to save on memory usage when most of the viewmodel is actually used in the view (only for postbacks or so).

This is only the first state, we can (and should) migrate other components of the framework to prefer using the state directly instead of going through the knockout observables. notably:

* static commands
* validation
* ReactBridge

We will do that in a following PRs, after we check that this change is OK (so, please review before we can move)


quigamdev:
I have used this PR on one of my projects. I will report as soon as I get some meaningful results.

PR #887, opened 2020-09-24
Customizable GridViewDataSets

exyi:


tomasherceg:
I've played with this a little bit - here are my observations:

1. I've noticed that sample API projects were broken - we had methods accepting SortingOptions and PagingOptions. This was never working well so it's not an issue, however renaming `SortingOptions` to `Sorter` and `PagingOptions` to `Pager` properties is quite unconvenient - we'll have to make this easy for the users.
Also, the `SortingOptions` and `PagingOptions` types were changed to `DefaultGridSorter` and `DefaultGridPager`, which is understandable, but again - ideally we should do something that'll help the users to fix these changes quickly (and safely).
I'll look how to create Roslyn analyzers and distribute them via NuGet package.

2. I suggest adding a variant of `DefaultGridPager` with just one type parameter - the second one would be the default `DistanceNearPageIndexesProvider<T>` which is used in 99% of cases. If someone was using `PagingOptions` in their controls (I've seen some custom pagers and stuff like that), it should help them - it is annoying to write `DefaultGridPager<MyCompany, DistanceNearPageIndexesProvider<MyCompany>>`.

3. I've created a sample (`/FeatureSamples/GenericGridViewDataSet/Paging`) with a custom pager with Prev/Next functionality. I am not sure how to implement `Apply` method for queryables - maybe using `SkipWhile`/`TakeWhile`, but I guess it won't be supported anywhere, so right now it throws `NotSupportedException`. 

4. I've create extension methods `TwitterExtensions.GoToNextPage` and `TwitterExtensions.GoToPreviousPage` (inspired by `GoTo...` from the `GridViewDataSetExtensions`) which make the transition and refresh the dataset, however I am not able to call them from the markup - probably because of the generic arguments or interface used in arguments (even if I call them as a static method, DotVVM cannot find the correct overload). 
I've made helper methods in the viewmodel to call these extension methods, but it complicates things when you want to build a custom pager as a markup control. 
I am not sure how difficult is to fix this (I am not requesting extension methods which I know are complicated - right now I'd be happy with calling generic methods with interface arguments).

5. I've defined an extension method to load datasets with Prev/Next paging - `LoadData`. However, I am not able to easily assign to `Items` and `IsRefreshRequested`. I had to use `dynamic`. 

6. I've been able to load the dataset and go to next/prev page using static command - I've just defined a method which accepts the current Pager object, reconstructs the dataset using it, calls `GoToNextPage`, loads and returns a new dataset which replaces the old one. This is as good start for being able to work with DataSets using REST APIs.

7. The test `TypeDescriptorUtils_GetPropertyOnInterface` is broken and I am not sure where the issue is.

Issue #889, opened 2020-09-29
Default Newtonsoft.Json settings in generated REST API Bindings


acizmarik:
We should generate code for REST API Bindings in such a way that we always explicitly pass `JsonSerializerSettings` whenever calling `JsonConvert.SerializeObject` or `JsonConvert.DeserializeObject`. This ensures consistency even when users decide to change the default serialization settings for Newtonsoft.Json

In the meantime users will be advised (in the dotvvm-docs project) to not change the default settings if they work with generated REST API Bindings.

quigamdev:
@acizmarik PR #890 was already merged. Can we close this issue or is there something else to be solved?

acizmarik:
There is something else to be solved. 

PR #890 fixed the issue #875. But someone might run into a similar issue when generating REST API bindings. This was not fixed and together with @tomasherceg we decided to postpone the fix. We wanted to look into other options on how to generate REST API bindings. However, this is not a change that we wanted to introduce into version 2.5

Also right now it seems that someone else run into an unrelated issue but also with generated REST API bindings #911. So it seems that there are more problems with the code generation mechanism.

PR #890, opened 2020-09-29
Fixed DotVVM.Framework to not rely on the default Newtonsoft.Json settings

### LINKED ISSUE #875: File Upload Files collection not set
Following the guidance for the FileUpload control usage here [File Upload Control](https://www.dotvvm.com/docs/controls/builtin/FileUpload/2.0) I'm seeing a weird issue.   The content of the UploadedFilesCollection has empty/default values.   

DotvvmStartup:
   
     public void ConfigureServices(IDotvvmServiceCollection options)
     {
        options.AddDefaultTempStorages("App_Data\\temporary");
     }

View Markup:

     <div class="modalBody">
       <div class="center">
                        <dot:FileUpload AllowMultipleFiles="false"
                                        class="btn btn-cl btn-oval"
                                        NumberOfFilesIndicatorText=""
                                        SuccessMessageText=""
                                        UploadCompleted="{command: OnUploadCompleted()}"
                                        UploadButtonText="Browse"
                                        UploadErrorMessageText=""
                                        UploadedFiles="{value: Files}" />
       </div>
     </div>

View Model:

     // ctor
     public CourseViewModel(IUploadedFileStorage uploadedFileStorage)
     {
         _uploadedFileStorage = uploadedFileStorage;
         Files = new UploadedFilesCollection();
     }
     public void OnUploadCompleted()
     {
       var file = Files.Files.First();
       UploadFileDialog = false;
     }

Debugging the file results in a FileId of "00000000-0000-0000-0000-000000000000", FileName is null, FileSize is 0B, IsAllowed is true, IsFileTypeAllowed is true and IsMaxSizeExceeded is false.  However when I go to the uploadedFiles folder in App_Data\temporary I see the uploaded files. 

Please advise on what I'm doing wrong with this project.   I have this near identical code working in a different project.   I'm using a Web Forms hybrid approach targeting .NET Full Framework 4.8.
FileUpload

acizmarik:
Resolves #875.

The issue, however, still remains in generated REST API Bindings. This is tracked as a new issue #889 and will be addressed in future releases. In the meantime, users will be notified to not change these settings when working with generated bindings.

### DIFF
--- a/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
@@ -72,8 +72,9 @@ namespace DotVVM.Framework.Tests.Runtime
             config1.Resources.Register("rs7", new PolyfillResource(){ RenderPosition =  ResourceRenderPosition.Head});
             config1.Resources.Register("rs8", new ScriptResource(new JQueryGlobalizeResourceLocation(CultureInfo.GetCultureInfo("en-US"))));
 
-            // serialize & deserialize
-            JsonSerializerSettings settings = new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.Auto };
+            // serialize & 
+            var settings = DefaultSerializerSettingsProvider.Instance.Settings;
+            settings.TypeNameHandling = TypeNameHandling.Auto;
             var config2 = JsonConvert.DeserializeObject<DotvvmConfiguration>(JsonConvert.SerializeObject(config1, settings), settings);
 
             //test 
--- a/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
@@ -72,7 +72,6 @@ namespace DotVVM.Framework.Tests.Runtime
             config1.Resources.Register("rs7", new PolyfillResource(){ RenderPosition =  ResourceRenderPosition.Head});
             config1.Resources.Register("rs8", new ScriptResource(new JQueryGlobalizeResourceLocation(CultureInfo.GetCultureInfo("en-US"))));
 
-            // serialize & 
             var settings = DefaultSerializerSettingsProvider.Instance.GetSettingsCopy();
             settings.TypeNameHandling = TypeNameHandling.Auto;
             var config2 = JsonConvert.DeserializeObject<DotvvmConfiguration>(JsonConvert.SerializeObject(config1, settings), settings);
--- a/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
@@ -73,7 +73,7 @@ namespace DotVVM.Framework.Tests.Runtime
             config1.Resources.Register("rs8", new ScriptResource(new JQueryGlobalizeResourceLocation(CultureInfo.GetCultureInfo("en-US"))));
 
             // serialize & 
-            var settings = DefaultSerializerSettingsProvider.Instance.Settings;
+            var settings = DefaultSerializerSettingsProvider.Instance.GetSettingsCopy();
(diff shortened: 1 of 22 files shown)


src/DotVVM.Framework.Tests.Common/Runtime/ResourceManagerTests.cs
```diff
@@ -72,8 +72,9 @@ public void ResourceManager_ConfigurationDeserialization()
             config1.Resources.Register("rs7", new PolyfillResource(){ RenderPosition =  ResourceRenderPosition.Head});
             config1.Resources.Register("rs8", new ScriptResource(new JQueryGlobalizeResourceLocation(CultureInfo.GetCultureInfo("en-US"))));
 
-            // serialize & deserialize
-            JsonSerializerSettings settings = new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.Auto };
+            // serialize &
```
exyi, line 75: fix the comment (or remove), please

src/DotVVM.Framework.Tests.Common/ViewModel/SerializerTests.cs
```diff
@@ -28,20 +29,22 @@ static ViewModelJsonConverter CreateConverter(bool isPostback, JObject encrypted
             };
         }
 
-        static string Serialize<T>(T viewModel, out JObject encryptedValues, bool isPostback = false)
+        string Serialize<T>(T viewModel, out JObject encryptedValues, bool isPostback = false)
```
exyi, line 32: It can still be static, right?

src/DotVVM.Framework/Configuration/DefaultSerializerSettingsProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Threading.Tasks;
+using DotVVM.Framework.ViewModel.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace DotVVM.Framework.Configuration
+{
+    public sealed class DefaultSerializerSettingsProvider
+    {
+        public JsonSerializerSettings Settings { get; private set; }
```
exyi, line 14: Are settings public and mutable on purpose?

acizmarik: I am changing it to a `readonly` field

Public modifier is currently not that necessary (it is accessed only from DotVVM.Framework and DotVVM.Framework.Tests). But in future it should be also accessed by generated REST API bindings.

Mutability is not desirable. However, it is hard to ensure immutability. This is mostly due to the fact that the class is designed to have mutable instances. I ran a simple test and measured the size of `JsonSerializerSettings` instance to be ~310 bytes

exyi: We could either make the field internal (and not modify it ourselves) or create new instance every time (preferably not by deserializing another instance, rather by calling the constructor and setting the properties)

exyi: > I am changing it to a readonly field

Nah, whatever, I was talking about the `JsonSerializerSettings` instance, not about the private setter - that it ok

acizmarik: > I was talking about the JsonSerializerSettings instance, not about the private setter - that it ok

I know. I just realized that a readonly field is sufficient in this case.

> We could either make the field internal (and not modify it ourselves)

I can make it internal for now (at least for the purposes of this specific bugfix). Not modifying it within the framework would be the best approach in my opinion.

> ...  or create new instance every time (preferably not by deserializing another instance, rather by calling the constructor and setting the properties)

The implementation is already changed to create a new instance by calling the constructor when requested.  However, we probably do not want to create a new instance every time someone accesses settings. While I realize that it would resolve this issue quite elegantly, there are problems with this approach:

-  instances are not trivial (~310 bytes)
-  they will be used once and immediately thrown away
-  many instances might be created for each request

While I have not really performed any perf-tests, I feel like it could negatively impact performance.

tomasherceg (approved):

PR #892, opened 2020-10-18
Recursive resolving of generic arguments types in registration of methods in javascript translation

quigamdev:
This PR rewrites resolving of generic arguments in ExpressionHelper.

Example 1: 
This registration is already allowed - this is how it was implemented so far.
```
T Unwrap(T value)
{
     return value.InneValue;
}
```

Example 2:  
This registration is not allowed in the current implementation. This PR makes it possible.
```
T Unwrap(TransferObject<T> value)
{
     return value.InneValue;
}
```


src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
             };
         }
 
+        private static Type ResolveGenericTypeFromGivenExpressions(int i, Type genericArgument, ParameterInfo[] parameters, Expression[] args)
```
exyi, line 262: `int i` is a bit confusing due to the cycle - could you please rename it, or use the genericParameter itself?

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+        {
+            for (var j = 0; j < parameters.Length; j++)
+            {
+                var parameter = parameters[j];
+                if (parameter.ParameterType.IsGenericParameter && parameter.ParameterType.GenericParameterPosition == i)
```
exyi, line 267: Why not `parameter.ParameterType == genericArgument`? (I haven't checked if that works, so sorry if it doesn't)

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+                {
+                    return args[i].Type;
+                }
+                if (parameter.ParameterType.IsArray)
+                {
+                    var elementType = parameter.ParameterType.GetElementType();
+                    var genericArgName = elementType.Name;
+                    var genericParameterPosition = elementType.GenericParameterPosition;
```
exyi, line 275: I doubt this branch does anything. Could you please add a test that checks that arrays also work? For example call this method:

```csharp
T[] Flatten<T>(List<T[]> x)
```

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+                    var value = GetGenericParameterType(genericArgument, parameter.ParameterType.GetGenericArguments(), args[j].Type);
+                    if (value is Type) return value;
+                }
+            }
+            return null;
+        }
+
+        private static Type GetGenericParameterType(Type genericArg, Type[] searchedGenericTypes, Type expressionType)
+        {
+            if (searchedGenericTypes is object && searchedGenericTypes.Length > 0)
```
exyi, line 288: When can be `searchedGenericTypes` null or empty? This method is only called when the type is generic, so it should contain at least one type argument

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+            {
+                var expressionGenericArgs = expressionType.GetGenericArguments();
+                for (var i = 0; i < searchedGenericTypes.Length; i++)
+                {
+                    var sgt = searchedGenericTypes[i];
+                    if (sgt.IsGenericParameter && sgt.GenericParameterPosition == genericArg.GenericParameterPosition)
```
exyi, line 294: It should be enough to check `sgt == genericArg`

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +259,53 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+                    {
+                        if (expressionGenericArgs.Length == 0) return expressionType;
+                        return expressionGenericArgs[i];
+                    }
+                    else if (sgt.IsGenericType)
```
exyi, line 299: And what if it's an array?

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -159,8 +160,12 @@ public static Expression CallMethod(Type target, BindingFlags flags, string name
         private static MethodRecognitionResult FindValidMethodOveloads(Type type, string name, BindingFlags flags, Type[] typeArguments, Expression[] arguments, IDictionary<string, Expression> namedArgs)
         {
             var methods = FindValidMethodOveloads(type.GetAllMembers(flags).OfType<MethodInfo>().Where(m => m.Name == name), typeArguments, arguments, namedArgs);
+            if (methods.Count() > 1)
+                // TODO: this behavior is not completed. Implement the same behavior as in roslyn.
+                throw new InvalidOperationException($"Found ambiguous overloads of method '{name}'.");
```
exyi, line 165: We do not have to implement all the rules C# uses to distinguish the operators, but this check breaks our existing overload resolution logic - we simply prefer less automatic conversions and then less implicit type arguments. Please add unit tests for these cases

* Calling `A(1)` will prefer `void A(int)` over `void A(object)`
* Calling `A(1)` will prefer `void A(int)` over `void A<T>(T)`
* Calling `A(1)` will prefer `void A<T>(T)` over `void A(object)`

Note that this most likely works now and this line "breaks" it. Instead of checking that there is only one matching method, we should check that there is only one with the max score that we use for sorting them.

exyi (changes requested):
Please see the comments in code and also please add unit tests for the behavior you have implemented - I don't see a reason to rely on UI tests for this. You can either add them as `BindingCompilationTests` or `JavascriptCompilationTests` (first option should be more stable, second also covers the translation to JS).

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
quigamdev: I had stuck to the solution that was used before my intervention. I have added several tests and can confirm that this comparison works.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
quigamdev: I have added support for this case. Also, I have added tests for it.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
quigamdev: I have added tests for the cases you have mentioned. Still, I am not a fan of just picking of method that is less automatically converted or just converted. I can't get off the feeling that someone will implement ambiguous methods and this logic will pick something else then C# would.

exyi (commented):
Very nice, and thanks for the tests. Could you please just have a look at the generic types in arrays?

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +269,56 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
...
+                if (parameterType == genericArgument)
+                {
+                    return args[j].Type;
+                }
+                if (parameterType.IsArray)
+                {
+                    parameterType = parameter.ParameterType.GetElementType();
+                    if (parameterType == genericArgument && args[j].Type.IsArray)
+                    {
+                        return args[j].Type.GetElementType();
```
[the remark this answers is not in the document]
quigamdev: I have added a test case for the method with a parameter of type `List<T>[]` and fixed the code to respect this case.

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -252,6 +269,56 @@ private static MethodRecognitionResult TryCallMethod(MethodInfo method, Type[] t
             };
         }
 
+        private static Type ResolveGenericTypeFromGivenExpressions(Type genericArgument, ParameterInfo[] parameters, Expression[] args)
```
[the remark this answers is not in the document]
quigamdev: I have removed this method and I have used GetGenericParameterType(). Thank you for pointing this out.

exyi (approved):
Nice

PR #896, opened 2020-10-30
Allow use custom instance of IDotvvmServiceConfigurator for AddDotVVM

martindybal:


src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -27,10 +27,20 @@ public static class ServiceCollectionExtensions
...
+            var configurator = new TServiceConfigurator();
+            return services.AddDotVvm(configurator);
+        }
+
+        /// <summary>
+        /// Adds DotVVM services with authorization and data protection to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="configurator">Configurator instance</param>
+        public static IServiceCollection AddDotVvm<TServiceConfigurator>(this IServiceCollection services, TServiceConfigurator configurator) where TServiceConfigurator : IDotvvmServiceConfigurator, new()
```
tomasherceg, line 40: The casing should be `AddDotVVM`.

tomasherceg (changes requested):
* For consistency reasons, we should enable the same thing in `UseDotVVM`
* We also need to add the same thing to `Hosting.Owin`

src/DotVVM.Framework.Hosting.AspNetCore/ServiceCollectionExtensions.cs
martindybal: There is such a method, but it's private. I can make it public. `applicationRootPath`, `useErrorPages` and `modifyConfiguration` should be null as default

https://github.com/riganti/dotvvm/blob/78c2f51db091f991414f04748e2946f36f3b1b01/src/DotVVM.Framework.Hosting.Owin/AppBuilderExtensions.cs#L61

tomasherceg (approved):
Now it seems good to me.
Just one question for @exyi - we've enforced the default constructor for both `TDotvvmStartup` and `TDotvvmServiceConfigurator` using constraints - even in the overloads that accept an instance of these types. 
Technically, these overloads don't need such constraints because they already get the instance, but since the compiler will need a default constructor on both types, it looks to me as a good idea to enforce it via constraint. What do you think?

src/DotVVM.Framework.Hosting.AspNetCore/ApplicationBuilderExtensions.cs
```diff
@@ -29,16 +29,18 @@ public static class ApplicationBuilderExtensions
         /// A value indicating whether to show detailed error page if an exception occurs. It is enabled by default
         /// if <see cref="HostingEnvironmentExtensions.IsDevelopment" /> returns <c>true</c>.
         /// </param>
+        /// <param name="modifyConfiguration">Delegate is called before configuration is freeze</param>
```
exyi, line 32: frozen

src/DotVVM.Framework.Hosting.AspNetCore/ApplicationBuilderExtensions.cs
```diff
@@ -47,12 +49,9 @@ public static DotvvmConfiguration UseDotVVM<TStartup>(this IApplicationBuilder a
         /// A value indicating whether to show detailed error page if an exception occurs. It is enabled by default
         /// if <see cref="HostingEnvironmentExtensions.IsDevelopment" /> returns <c>true</c>.
         /// </param>
-        public static DotvvmConfiguration UseDotVVM(this IApplicationBuilder app, string applicationRootPath, bool? useErrorPages, Action<DotvvmConfiguration> modifyConfiguration = null)
-        {
-            return UseDotVVM(app, applicationRootPath, useErrorPages, null, modifyConfiguration);
-        }
-
-        private static DotvvmConfiguration UseDotVVM(this IApplicationBuilder app, string applicationRootPath, bool? useErrorPages, IDotvvmStartup startup, Action<DotvvmConfiguration> modifyConfiguration)
+        /// <param name="modifyConfiguration">Delegate is called before configuration is freeze</param>
```
exyi, line 52: frozen

exyi (approved):

exyi:
>  Just one question for @exyi - we've enforced the default constructor for both `TDotvvmStartup` and `TDotvvmServiceConfigurator` using constraints - even in the overloads that accept an instance of these types.
Technically, these overloads don't need such constraints because they already get the instance, but since the compiler will need a default constructor on both types, it looks to me as a good idea to enforce it via constraint. What do you think?

I'm not sure what is actually better, but I have two points agains the constants:
* Not everyone uses the compiler, since it does not really work. When you decide to use it, our error message could be clearer than "we need a default constructor" - it could also explain why and what you should use it for.
* Some common arguments (like hosting env) could be atomaticaly faked by the compiler.

exyi:
One more thing, could you please just squash the commits when merging? We don't need that many build fix commits :D

PR #903, opened 2020-11-09
Add support for lambdas in command bindings

### LINKED ISSUE #118: Lambdas in command bindings
Since parameters can be passed to command bindings, it should also be possible to use them directly. Like:

``` html
<dot:GridView DataSource="{value: Items5}" SortChanged="{command: column => Sort('Items5', column)}" />
```

acizmarik:
Resolves #118.

Binding parser supports basic lambda definitions; empty lambdas `() => ;`, single-parameter lambdas `arg => expr` or `(arg) => expr`, multiple parameter lambdas `(arg1, arg2) => expr`, lambdas with type information `(string arg) => expr`. 

When compiling to C#'s expressions there are following limitations:
- Lambdas need to provide information about parameter types
- Parameter identifiers can not collide with already registered symbols within `TypeRegistry`. In other words, identifiers can not use names such as `_this` or `_parent`

Expressions to javascript translation. PoC translations:
- `Enumerable.Where<TSource>(IEnumerable<TSource>, Func<TSource, bool>)` -> `filter`
- `Enumerable.Select<TSource, TResult>(IEnumerable<TSource>, Func<TSource, TResult>)` -> `map`


exyi:
Note that lambda function can contain any expression, not only a method invocation. These should be valid: `x => x + 1`, `x => x.Prop`, ... Also, our semicolon is actually a binary operator with a high priority, so the lambdas should also support "kinda blocks" `x => (Method1(); Method2())`. With #869 it would also support `x => (var y = Method1(x); Method2(y))`

src/DotVVM.Framework/Compilation/Parser/Binding/Parser/BindingParser.cs
```diff
@@ -340,7 +340,149 @@ private BindingParserNode ReadUnaryExpression()
...
+                return false;
+
+            type = ReadIdentifierExpression(true);
+            if (Peek()?.Type != BindingTokenType.Identifier)
+            {
+                name = type;
+                type = null;
+                return true;
+            }
+            name = ReadIdentifierExpression(true);
```
exyi, line 436: We should check that name is a SimpleNameBindingParserNode - it should not have generic arguments and so on.

exyi (commented):
Nice work :)

acizmarik:
Thanks for feedback! The parser is adjusted so that it accepts any expression (as a lambda body). Generated code seems to be working. Parameter types are currently required by the expression visitor.

I will add more tests and then mark it as ready for review

src/DotVVM.Framework.Tests.Common/Binding/BindingCompilationTests.cs
```diff
@@ -145,6 +145,62 @@ public void BindingCompiler_Valid_MethodCall()
...
+        }
+
+        [TestMethod]
+        [DataRow("() => ;", typeof(Action), null)]
+        [DataRow("() => \"HelloWorld\"", typeof(Func<string>), typeof(string))]
+        [DataRow("() => 11", typeof(Func<int>), typeof(int))]
+        [DataRow("() => 4.5f", typeof(Func<float>), typeof(float))]
+        [DataRow("() => 4.5", typeof(Func<double>), typeof(double))]
+        [DataRow("() => 11 + 4.5", typeof(Func<double>), typeof(double))]
+        [DataRow("() => new System.Collections.Generic.List<int>()", typeof(Func<List<int>>), typeof(List<int>))]
```
acizmarik, line 169: Parser currently does not support the `new` operator. Due to this fact, similar constructs currently fail. Should we support this? I think that someone could try to write something similar.

tomasherceg: Thanks to the type metadata in #905, we could support this in the future (with some limitations like default property values).

acizmarik:
### Type inference

1. Pass information about the expected type to `ExpressionBuildingVisitor`.
2. In `VisitLambda` we can enrich `LambdaBindingParserNode` with type information and check for any mismatches in parameters.
3. In `VisitLambdaParameter` we must resolve the type. If the type information was specified explicitly, we visit the `BindingParserNode`. If it was inferred, we already should have an instance of `System.Type` that we extracted from expected lambda type.
4. At the end of `VisitLambda` validate that the `Type` of lambda body expression matches the expected result type (if available)

The `expectedType` is already passed to `BindingExpressionBuilderExtensions::ParseWithLambdaConversion(...)`. Subsequently, this method calls `IBindingExpressionBuilder::Parse(...)` where the `ExpressionBuildingVisitor` is constructed.

### Motivation

This would allow the following construct:
`<dot:GridView DataSource="{value: DataSet}" SortChanged="{command: (string column) => CustomSort(column)}">`

To be simplified as follows:
`<dot:GridView DataSource="{value: DataSet}" SortChanged="{command: (column) => CustomSort(column)}">`

DotVVM already has the necessary type information for `SortChanged` and during compilation the `expectedType` is set to `Action<string>`

### Possible errors

- Parameter types can not be inferred
   - User did not specify types for lambda parameters and type inference is not available
- Parameter types mismatch
   - User specified types for lambda parameters but type inference expects different types
- Return type mismatch
   - Similar to the previous error but on return type
   - Also we should check that the type of lambda's body expression matches the provided return type

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
```diff
@@ -176,6 +176,18 @@ public void AddDefaultMethodTranslators()
...
+                .Where(m => m.Name == "Where" && m.GetParameters().Length == 2 && m.GetParameters().Last().ParameterType.GetGenericTypeDefinition() == typeof(Func<,>)).Single();
+            AddMethodTranslator(whereMethod, translator: new GenericMethodCompiler(
+                args => args[1].Member("filter").WithAnnotation(ResultIsObservableArrayAnnotation.Instance)
+                    .Invoke(args[2]).WithAnnotation(ResultIsObservableAnnotation.Instance)));
+
+            var selectMethod = typeof(Enumerable).GetMethods(BindingFlags.Public | BindingFlags.Static)
+                .Where(m => m.Name == "Select" && m.GetParameters().Length == 2 && m.GetParameters().Last().ParameterType.GetGenericTypeDefinition() == typeof(Func<,>)).Single();
+            AddMethodTranslator(selectMethod, translator: new GenericMethodCompiler(
+                args => args[1].Member("map").WithAnnotation(ResultIsObservableArrayAnnotation.Instance)
+                    .Invoke(args[2]).WithAnnotation(ResultIsObservableAnnotation.Instance)));
```
exyi, line 190: Why `.WithAnnotation(ResultIsObservableArrayAnnotation.Instance)`? The result is not obervable, or is it? If this is required for it to work, then there is something wrong in the JS translator.

src/DotVVM.Framework/Compilation/Binding/TypeRegistry.cs
```diff
@@ -121,7 +122,8 @@ public static Expression CreateStatic(Type type)
                .Add("Decimal", CreateStatic(typeof(Decimal)))
                .Add("Double", CreateStatic(typeof(Double)))
                .Add("Single", CreateStatic(typeof(Single)))
-               .Add("String", CreateStatic(typeof(String))),
+               .Add("String", CreateStatic(typeof(String)))
+            .Add("Enumerable", CreateStatic(typeof(Enumerable))),
```
exyi, line 126: Nitpick: a strange indent.

exyi (approved):
Ok, this is nice. I'm just not convinced that the `Select` and `Where` translator actually work, could you please add a unit test for that (just translate that to JS in `JavascriptCompilationTests`, not necessarily a UI test)

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
acizmarik: I also think that the result should not be observable. The created sample works also without the annotation.

Moreover, I added unit tests for the Javascript translation of `Where` and `Select` methods. Can you please check that the translations are correct? There is also a sample `LambdaExpressions` that you can try to play with these translations.

Issue #882, opened 2020-08-29
Client-side ViewModel Specification


tomasherceg:
This is not a single issue - its purpose is to track several steps that need to be done in order to allow the implementation of several planned features.

* [ ] Change the generator of validation maps to return complete type information
* [ ] Update client-side validation code to use the new `typeMetadata` structure
* [ ] Implement coercion and validation logic
* [ ] Update serializer/deserializer to use `typeMetadata` instead of `$options`
* [ ] Implement support for `TimeSpan`
* [ ] Implement support for the `createInstance` method

Thanks to the strongly-typed nature of DotVVM, we've found out that it would be useful to have a complete type information about the structure of the viewmodel on the client. 
In the first place, it will allow the users to work safely with DotVVM viewmodels on the client. Also, it will help us to bring new features like immutable-state viewmodels which will hopefully lead to a better client-side performance.

## Supported types in the viewmodels

The viewmodel of the page ($root) must always be an object. DotVVM doesn't assume that the root viewmodel would be an array or a primitive type - if this is intended, such type can be wrapped as the only member of an object type.

The viewmodel can contain properties of the following supported types:

* primitive types (see the table further)
* object types (that meet the same criteria as the viewmodel)
* arrays of supported primitive types or supported object types

### Supported primitive types

The following table lists the primitive types that are supported in DotVVM. 

| Server type | Client type | Notes                                                                 |
|-------------|-------------|-----------------------------------------------------------------------|
| bool        | boolean     |                                                                       |
| byte        | number      |                                                                       |
| sbyte       | number      |                                                                       |
| short       | number      |                                                                       |
| ushort      | number      |                                                                       |
| int         | number      |                                                                       |
| uint        | number      |                                                                       |
| long        | number      | Possible precision loss                                               |
| ulong       | number      | Possible precision loss                                               |
| float       | number      |                                                                       |
| double      | number      |                                                                       |
| decimal     | number      | Possible precision loss                                               |
| string      | string      |                                                                       |
| char        | string      |                                                                       |
| Guid        | string      |                                                                       |
| Enum        | string      |                                                                       |
| DateTime    | Date        | Until DotVVM 3.0, date values were stored as strings in the viewmodel |
| TimeSpan    | Date        | New in DotVVM 3.0                                                     |

#### Nullability

[...]

PR #905, opened 2020-11-11
Type metadata for viewmodels

tomasherceg:
A draft PR that implements proposed changes from #882


src/DotVVM.Framework.Tests.Common/ViewModel/ViewModelTypeMetadataSerializerTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                    'parameters': [
+                        0,
+                        10
+                    ]
+                }
+            ]
+        }
+    }
+}";
+            Assert.AreEqual(JToken.Parse(expected.Replace("'", "\"")).ToString(Formatting.None), result.ToString(Formatting.None));
```
exyi, line 125: You can use https://github.com/exyi/CheckTestOutput instead of these scary literals ;)

src/DotVVM.Framework.Tests.Common/ViewModel/ViewModelTypeMetadataSerializerTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        },
+        'property ONE': {
+            'type': 'Guid'
+        },
+        'property TWO': {
+            'type': [
+                {
+                    'type': 'nullable',
+                    'inner': {
+                        'type': 'enum',
```
exyi, line 76: Hmm, could we pre-declare the enums like we do for types? This could get quite long if you would use the enum often.

src/DotVVM.Framework/Resources/Scripts/global-declarations.ts
```diff
@@ -174,3 +175,29 @@ type ValidationRuleTable = {
...
+}
+
+type TypeMetadata = {
+    [prop: string]: PropertyMetadata
+}
+
+type PropertyMetadata = {
+    type: TypeDefinition;
+    post?: "always" | "pathOnly" | "no";
+    update?: "always" | "firstRequest" | "no";
```
exyi, line 190: Is this actually required? We can just trust the server not to send us these properties.

PR #912, opened 2020-11-26
Switch most of the view compiler to nullable reference types

exyi:


tomasherceg:
Looks good to me - there is one failing test but according to @quigamdev, it should be already fixed in #910.

tomasherceg (approved):

PR #914, opened 2020-11-27
Add initCompleted JavaScript event

mrnustik:
Adds `initCompleted` event to `dotvvm.events` which is triggered after `init` function is finished and bindings have been applied.

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/global-declarations.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/global-declarations.ts
@@ -122,6 +122,9 @@ type DotvvmInitEventArgs = {
     readonly viewModel: any
 }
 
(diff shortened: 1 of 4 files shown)


src/DotVVM.Framework/Resources/Scripts/global-declarations.ts
```diff
@@ -122,6 +122,9 @@ type DotvvmInitEventArgs = {
     readonly viewModel: any
 }
 
+type DotvvmInitCompletedEventArgs = {
```
mrnustik, line 125: @tomasherceg Is there anything you would like to be passed as an event argument for the `initCompleted` event?

tomasherceg (approved):

Issue #899, opened 2020-11-07
Invoking JS module functions from bindings


tomasherceg:
Allows to use `_js.Invoke<TResult>("functionName", arg1, arg2...)` in binding expressions.

See #749 for details.

exyi:
This is already implement.

PR #915, opened 2020-12-02
Implement support of @js directive

exyi:
It renders an inline script that imports the module and register it into dotvvm.

When in a page, it is added into the serialized JSON.
When in a control, it renders a binding handler with the specified module


quigamdev:
This PR also solves #899.

exyi:
Succeeded by #935

PR #920, opened 2020-12-18
Postbacks were incorrectly disabled on redirects

tomasherceg:
We didn't realize that returning files is commonly done via redirects, and we were disabling postbacks on redirects. 
This PR returns to the previous behavior. 

Also, I've found an issue that we should fix in 3.x branch - long running static commands can modify a viewmodel of another page if SPA navigation is triggered during the time static command is running.


exyi (approved):
Ok, fine. We should also fix in v3, there is likely going to be the same bug.

PR #923, opened 2021-01-04
Add overridable Active and Disabled css classes for DataPager

mrnustik:
Adds overridable properties for active page item and disabled page item to DataPager. This can be for easily overriding DataPager and using their own CSS classes.


exyi (approved):
On line 290, there is a direct usage of the "disabled" css class, could you also replace it with `DisabledItemCssClass`? Otherwise, it's fine

mrnustik:
@exyi should be fixed by now. I must have missed it. Thanks!

PR #928, opened 2021-01-17
Static Commands in MarkupControl

Mylan719:
I try to use static command as much as possible to make my pages faster. However I ren into a severe limitation. It is impossible to bind `Command` property to static command binding in the markup control.
Example:
Let's assume I have `DeviceDetail` markup control used in the page like so:
```html
<cc:DeviceDetail Save={staticCommand: _root.Detail = service.Save(Detail); _alert.Show('Item saved')} />
``` 
I wanted to implement the `DeviceDetail` control as follows:
```html
@viewModel MyProject.DeviceViewModel
@baseType MyProject.Device 
<div>
<!-- ... --/>
    <dot:Button Click={staticCommand: _control.Save()} Text="Save"/>
</div>
```
Code behind:
```csharp
public class Device : DotvvmMarkupControl {
public Command Save
        {
            get { return (Command)GetValue(SaveProperty); }
            set { SetValue(SaveProperty, value); }
        }
        public static readonly DotvvmProperty SaveProperty
            = DotvvmProperty.Register<Command, Device>(c => c.Save, null);
}
```
This use case does not work as it generates client side JavaScript `contetext.$control.Save()` for the save button in the markup control. The `Save` property is not present in the `$control` context and attempt to press the "Save" button fails silently. 

Since markup control properties of type `Command` generate no corresponding property into the `$control` client context no control commands can be called from staticCommand binding. 

We need to correct this.

### DIFF
--- a/src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
@@ -392,13 +392,13 @@ namespace DotVVM.Framework.Tests.Binding
         public static readonly DotvvmProperty LoadProperty
             = DotvvmProperty.Register<Func<string>, TestMarkupControl>(c => c.Load, null);
 
-        public Func<string, string> Chanege
+        public Func<string, string> Change
         {
-            get { return (Func<string, string>)GetValue(ChanegeProperty); }
-            set { SetValue(ChanegeProperty, value); }
+            get { return (Func<string, string>)GetValue(ChangeProperty); }
+            set { SetValue(ChangeProperty, value); }
         }
-        public static readonly DotvvmProperty ChanegeProperty
-            = DotvvmProperty.Register<Func<string, string>, TestMarkupControl>(c => c.Chanege, null);
+        public static readonly DotvvmProperty ChangeProperty
+            = DotvvmProperty.Register<Func<string, string>, TestMarkupControl>(c => c.Change, null);
 
 
         public static TestMarkupControl CreateInitialized()
@@ -406,7 +406,7 @@ namespace DotVVM.Framework.Tests.Binding
             var control = new TestMarkupControl();
             control.SetBinding(SaveProperty, new FakeCommandBinding(new ParametrizedCode("test"), null));
             control.SetBinding(LoadProperty, new FakeCommandBinding(new ParametrizedCode("test2"), null));
-            control.SetBinding(ChanegeProperty, new FakeCommandBinding(new ParametrizedCode("test3"), null));
+            control.SetBinding(ChangeProperty, new FakeCommandBinding(new ParametrizedCode("test3"), null));
             return control;
         }
 
--- a/src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
@@ -307,9 +307,9 @@ namespace DotVVM.Framework.Tests.Binding
             var control = new TestMarkupControl();
             control.SetBinding(TestMarkupControl.SaveProperty, new FakeCommandBinding(new ParametrizedCode("test"), null));
 
-
             var result = CompileBinding("_control.Save()", niceMode: true, new[] { typeof(object) }, typeof(Command), typeof(TestMarkupControl));
 
+            Assert.Fail("TODO: implement changes and create expected result.");
         }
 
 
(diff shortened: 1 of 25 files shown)


exyi (commented):
There is still the problem that the translator does not know that the command invocation will return a promise and thus won't wait for it. It works for the simplest cases when only one command is invoked, but will not work when the command result value would be used in a second step, or so.

src/DotVVM.Framework/Controls/DotvvmMarkupControl.cs
```diff
@@ -97,6 +97,19 @@ private PropertySerializeInfo GetPropertySerializationInfo(DotvvmProperty proper
                     valueBinding.GetKnockoutBindingExpression(this)
                 );
             }
+            else if (GetBinding(property) is StaticCommandBindingExpression staticCommand)
```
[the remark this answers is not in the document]
Mylan719: This was suprisingly easy to implement. It works like a charm and there is new UI sample for that.

Mylan719:
```There is still the problem that the translator does not know that the command invocation will return a promise and thus won't wait for it.
```
This issue has been resolved, now the transpator wraps the control property invocation result into `Promise.resolve`.  It also adds correct annotation top the result so the compiler takes the property call into account when deciding command execution order.

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
```diff
@@ -15,6 +15,25 @@
...
+            if (method == null)
+            {
+                return null;
+            }
+
+            if (method.Name == "Invoke" && typeof(Delegate).IsAssignableFrom(method.DeclaringType))
+            {
+                var invocationTargetExpresionCall = context.JsExpression().Invoke(arguments.Select(a => a.JsExpression()));
+                return new JsIdentifierExpression("Promise").Member("resolve").Invoke(invocationTargetExpresionCall)
+                    .WithAnnotation(new ResultIsPromiseAnnotation(a=> a));
```
exyi, line 31:
```suggestion
                return invocationTargetExpresionCall
                    .WithAnnotation(new ResultIsPromiseAnnotation(a=>new JsIdentifierExpression("Promise").Member("resolve").Invoke(a)));
```

It would be a bit nicer to wrap it inside promise only when it's going to be waited for.

exyi (approved):
Very nice, thank you for this patch. Would be nice if you resolve my nitpicks.

We can add support for arguments later (ideally, with the lambdas merged), but I think that is going to be relatively easy.

src/DotVVM.Framework/Controls/DotvvmMarkupControl.cs
```diff
@@ -97,6 +97,19 @@ private PropertySerializeInfo GetPropertySerializationInfo(DotvvmProperty proper
...
+            {
+                var call = KnockoutHelper.GenerateClientPostBackExpression(
+                    property.Name,
+                    staticCommand,
+                    this,
+                    new PostbackScriptOptions(false, true, false, "$element"));
```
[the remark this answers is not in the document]
Mylan719: Resolved in a commit

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
Mylan719: I did not realize it works this way, So the promise accessor is used only if someone wants to chain `.then` behind it? Ok than ;)

src/DotVVM.Framework/Controls/DotvvmMarkupControl.cs
```diff
@@ -97,6 +97,19 @@ private PropertySerializeInfo GetPropertySerializationInfo(DotvvmProperty proper
...
+
+                return new PropertySerializeInfo(
+                    property,
+                    $"function(){{return {call}}}"
```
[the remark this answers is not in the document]
Mylan719: Something similar is being done as part of named command pull request. I will implement the argument feature after lambdas have been merged to main.

Mylan719:
Nitpicks resolved ;)

src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
```diff
@@ -292,12 +301,117 @@ public void StaticCommandCompilation_PromiseReturningTranslatedCall_NeedsReorder
...
+
+        public Func<string> Load
+        {
+            get { return (Func<string>)GetValue(LoadProperty); }
+            set { SetValue(LoadProperty, value); }
+        }
+        public static readonly DotvvmProperty LoadProperty
+            = DotvvmProperty.Register<Func<string>, TestMarkupControl>(c => c.Load, null);
+
+        public Func<string, string> Chanege
```
tomasherceg, line 395: I guess it is a typo - `Change`

tomasherceg (approved):

src/DotVVM.Framework.Tests.Common/Binding/StaticCommandCompilationTests.cs
quigamdev: Fixed

quigamdev (approved):

PR #930, opened 2021-01-22
ScriptResource is `defer`ed by default

exyi:
This is technically a breaking change, but it should have a lower impact than without this change.
Before this change, the dotvvm resource was defered (as it was a module), which caused a lot of issues.

### DIFF
--- a/src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
+++ b/src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
@@ -21,7 +21,7 @@ namespace DotVVM.Framework.ResourceManagement
 
         /// <summary>Location property is required!</summary>
         public ScriptResource()
-            : base(ResourceRenderPosition.Body, "text/javascript")
+            : this(location: null!) // hack: people assign the Location property late, but it should non-nullable...
         { }
 
         public override void RenderLink(IResourceLocation location, IHtmlWriter writer, IDotvvmRequestContext context, string resourceName)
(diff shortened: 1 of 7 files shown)


tomasherceg (approved):

src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
```diff
@@ -13,7 +13,7 @@ namespace DotVVM.Framework.ResourceManagement
     public class ScriptResource : LinkResourceBase, IPreloadResource, IDeferableResource
     {
         public bool Defer { get; }
-        public ScriptResource(IResourceLocation location, bool defer = false)
+        public ScriptResource(IResourceLocation location, bool defer = true)
```
MichalTichy, line 16: defer should be also set to true when using parameterless constructor

src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
```diff
@@ -13,7 +13,7 @@ namespace DotVVM.Framework.ResourceManagement
     public class ScriptResource : LinkResourceBase, IPreloadResource, IDeferableResource
     {
         public bool Defer { get; }
```
MichalTichy, line 15: Other properties of resources have public setter (and those setters are heavily utilized). Is there a reason why we don't have it here?

MichalTichy (changes requested):

src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
exyi: Thanks, should be fixed.

src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
exyi: We set `RenderPosition` to Body or Anywhere based on defer's value. This would just break if we added setter to `Defer`. Generally, setters only cause problems and we keep them for legacy reasons.

MichalTichy (approved):

PR #931, opened 2021-01-29
Add support for using extension methods in bindings

acizmarik:
This PR allows using extension methods inside bindings. #903 should be merged first.

- Supported types to search for extension methods are defined within `TypeRegistry`
- Extension methods are searched only in case the basic method search fails
  - Regular methods are always preferred

Currently the only registered type to look for extension methods is `System.Linq.Enumerable`. This makes it easier to define lambdas when using LINQ.

* Initial lambdas support #903: `Enumerable.Where(Array, (int item) => item % 2 == 0)`
* This PR introduces syntactic sugar: `Array.Where((int item) => item % 2 == 0)`


src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -63,6 +64,12 @@ public static IEnumerable<MemberInfo> GetAllMembers(this Type type, BindingFlags
                 return type.GetMembers(flags);
         }
 
+        public static IEnumerable<MemberInfo> GetAllExtensions(this Type type, BindingFlags flags = BindingFlags.Public | BindingFlags.Static)
+        {
+            foreach (var registeredType in TypeRegistry.GetRegisteredTypesForExtensionMethodsLookup())
+                foreach (var member in registeredType.GetMembers(flags))
+                    yield return member;
```
exyi, line 71: Does this return something else than method?

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
```diff
@@ -147,36 +155,74 @@ public static Expression Call(Expression target, Expression[] arguments)
...
-                    // TODO: this behavior is not completed. Implement the same behavior as in roslyn.
-                    throw new InvalidOperationException($"Found ambiguous overloads of method '{name}'.");
+                    // Change to a static call
+                    var newArguments = CreateArgumentsForExtensionMethod(target, arguments);
+                    var extensions = FindValidMethodOveloads(type.GetAllExtensions().OfType<MethodInfo>().Where(m => m.Name == name), typeArguments, newArguments, namedArgs).ToList();
+
+                    // We found an extension method
+                    if (extensions.Count == 1)
+                    {
+                        extensions[0].IsExtension = true;
```
exyi, line 203: Shall we set `IsExtension = true` for all found extension methods?

src/DotVVM.Framework/Utils/ReflectionUtils.cs
acizmarik: No - it should return only methods. Changed to `MethodInfo`

src/DotVVM.Framework/Compilation/Binding/ExpressionHelper.cs
acizmarik: Yes - thanks I did not notice that :). Fixed

exyi (approved):
Very nice :)

acizmarik:
To use custom extension methods you can:

- Implement own `IExtensionsProvider` or
- Extend the `DefaultExtensionsProvider`

This needs to be then registered using dependency injection. If user registers no implementation, DotVVM uses the `DefaultExtensionsProvider` that has already the `Enumerable` type registered

src/DotVVM.Framework.Tests.Common/Binding/CustomExtensionMethodTests.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace DotVVM.Framework.Tests.Common.Binding
+{
+    [TestClass]
+    public class CustomExtensionMethodTests
+    {
+        private DotvvmConfiguration configuration;
+        private MemberExpressionFactory memberExpressionFactory;
+
+        [TestInitialize]
+        public void INIT()
```
tomasherceg, line 19: Rename to `Init` please

src/DotVVM.Framework/Compilation/Binding/MemberExpressionFactory.cs
```diff
@@ -0,0 +1,457 @@
...
+                GetDotvvmPropertyMember(target, name) is Expression result) return result;
+
+            var members = type.GetAllMembers(BindingFlags.Public | (isStatic ? BindingFlags.Static : BindingFlags.Instance))
+                .Where(m => ((isGeneric && m is TypeInfo) ? genericName : name) == m.Name)
+                .ToArray();
+
+            if (members.Length == 0)
+            {
+                // We did not find any match in regular methods => try extension methods
+                var extensions = extensionsProvider.GetExtensionMethods()
```
tomasherceg, line 51: Shall we cache this? I am afraid that this will browse the types and look for the methods every single time.

acizmarik: We probably should. The implementation now caches the result of the method call `GetMethods(..)` to avoid unnecessary reflection. Also the attribute `ExtensionAttribute` is now checked to ensure that we yield only extension methods

PR #933, opened 2021-02-05
Issue with async command execution

Mylan719:
There was en issue with executing async commands. In case en exception was thrown, the command result was set to the executing `Task` instead of null. This caused a seriqalization error.
Also now, if user tries to return custom awaitable, they will be warned that this is not supported. Before it would just fail quietly.

### DIFF
--- a/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
+++ b/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
@@ -392,23 +392,21 @@ namespace DotVVM.Framework.Hosting
                 await filter.OnCommandExecutingAsync(context, action);
             }
 
-            object? result = null;
             try
             {
-                Task? resultTask = null;
+                var commandResultOrNotYetComputedTask = action.Action();
 
-                result = action.Action();
-
-                resultTask = result as Task;
-                if (resultTask != null)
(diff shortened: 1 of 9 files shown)


src/DotVVM.Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -392,23 +393,25 @@ public async Task ProcessStaticCommandRequest(IDotvvmRequestContext context)
...
+                if (commandResultOrNotYetComputedAwaitable is Task commandTask)
                 {
-                    await resultTask;
+                    await commandTask;
+                    return TaskUtils.GetResult(commandTask);
                 }
 
-                if (resultTask != null)
+                var resultType = commandResultOrNotYetComputedAwaitable?.GetType();
+                var possibleResultAwaiter = resultType?.GetMethod(nameof(Task.GetAwaiter));
```
exyi, line 407:
```suggestion
                var possibleResultAwaiter = resultType?.GetMethod(nameof(Task.GetAwaiter), Array.Empty<Task>());
```

The awaiter method must have no parameters - I'd add that to the check to prevent false positives of the check,

exyi (approved):

tomasherceg (approved):

PR #936, opened 2021-02-09
Add support for basic type inference within lambda functions

acizmarik:
This PR adds support for type inference within lambda functions.

### Overview
We try to infer information about lambdas based on provided arguments and context in which the lambda is being used. Current biggest limitation is that we do not take into account lambda bodies. Still, in most cases, this simpler approach should work and infer correct type information. Following situations are covered by the type inferer:

1. **Lambdas as method arguments**
    - Gather usable method candidates
    - Based on other method arguments, resolve open generic types and refine method candidates (in each iteration we have more accurate information about the method call)
    - Based on refined method candidates and resolved generics, try infer lambda type
2. **Top-level lambdas**, such as `SortChanged` on grids
    - Here we must rely on the computed binding's expected type
3. **Compilation into Action<..> vs Func<..> vs Predicate<>**
    - Originally presented as #995 (abandoned)
    - DotVVM distinguishes 2 statements: assignment and method invocation
    - If user specified a valid statement as lambda body and lambda should be assigned into an `Action<..>`, return value gets discarded and as such the compilation result can be treated as an `Action<..>`
    - This implementation uses type inference to decide whether we need to discard return value or not

### Breaking change
Interface `IBindingExpressionBuilder` was altered by adding an optional parameter to its Parse method. Regardless, I do not think that this change could actually break someone


acizmarik:
I added a test that showcases this feature works with methods introduced by #965. It should work nicely together with method calls, but also with extension method calls.

src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -106,6 +106,29 @@ public static string ExtractMemberStringValue(object? item, string propertyName)
             return item + "";
         }
 
+        /// <summary>
+        /// Checks whether given instantiated type is compatible with the open generic type
+        /// </summary>
+        public static bool IsAssignableToGenericType(this Type givenType, Type genericType)
```
exyi, line 112: It would be better if this method returned the mapping of concrete types to the generic arguments from `genericType`

src/DotVVM.Framework/Compilation/Inference/Results/ITypeInferenceResult.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DotVVM.Framework.Compilation.Inference.Results
+{
+    internal interface ITypeInferenceResult
+    {
+        public bool Result { get; }
+        public Type? Type { get; }
```
exyi, line 13:
```suggestion
        bool Result { get; }
        Type? Type { get; }
```

exyi (approved):

src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
```diff
@@ -317,16 +360,49 @@ protected override Expression VisitLambda(LambdaBindingParserNode node)
...
+                var parameterType = Visit(node.Type).Type;
+                return Expression.Parameter(parameterType, node.Name.ToDisplayString());
+            }
+        }
+
+        private Expression CreateLambdaExpression(Expression body, ParameterExpression[] parameters, Type? delegateType)
+        {
+            if (delegateType != null)
+            {
+                if (delegateType.Name.StartsWith("Action"))
```
tomasherceg, line 388: This can potentially match things we don't want - I'd check namespace and also presence of a backtick.

src/DotVVM.Framework/Compilation/Inference/Inferers/LambdaInferer.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var delegateParameters = delegateType.GetMethod("Invoke", BindingFlags.Public | BindingFlags.Instance).GetParameters();
+            if (delegateParameters.Length != argsCount)
+                return false;
+
+            var generics = (context != null) ? context.Generics : new Dictionary<string, Type>();
+            if (!TryInstantiateDelegateParameters(delegateType, argsCount, generics, out parameters))
+                return false;
+
+            isFunc = delegateType.Name.StartsWith("Func");
```
tomasherceg, line 104: Remove

tomasherceg (approved):

PR #937, opened 2021-02-13
Fix generics resolving in method overloads search

acizmarik:
This PR fixes a bug that failed method overloads search, for example in the following case:

- Suppose user created a method with given signature: `T Method<T>(Enumerable<T> arg)`
- User called the method by passing an array `T[]` as the first and the only argument
- DotVVM tries to resolve generic argument `T` based on `typeof(T[]).GetGenericArguments()` which fails for arrays


exyi (approved):
Ok, this is not worse than before. However this logic actually does not take implicit conversions into account at all, maybe we should fix that more generally. For example, this wont work:

```csharp
interface A<T> { }
class B: A<int> { }

void Method<T>(I<T> x);
```

Invoking the `Method(new B())` will fail.

PR #938, opened 2021-02-14
Added urlSuffixProvider to AddRouteRedirection

tomasherceg:
I tried to use this new route redirection feature and found out that it doesn't have a way to work with query string parameters so I added this extensibility point.


exyi:
It would make sense to me to keep the query string by default.

tomasherceg:
@exyi Good idea, I've changed it so the query string is preserved by default. If somebody doesn't want it, they can specify a function that returns an empty string.

tomasherceg:
we need to add tests

PR #944, opened 2021-02-21
Fixed loading of the fallback script when a defer attribute is used

quigamdev:
Fixes usage of `defer` attribute on a fallback script.

### DIFF
--- a/src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
+++ b/src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
@@ -1,9 +1,11 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using DotVVM.Framework.Controls;
 using DotVVM.Framework.Hosting;
+using Newtonsoft.Json;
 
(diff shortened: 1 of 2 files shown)


src/DotVVM.Framework/ResourceManagement/ScriptResource.cs
```diff
@@ -44,5 +51,23 @@ public void RenderPreloadLink(IHtmlWriter writer, IDotvvmRequestContext context,
             writer.RenderBeginTag("link");
             writer.RenderEndTag();
         }
+        protected override void RenderFallbackLoadingScript(IHtmlWriter writer, IDotvvmRequestContext context, string resourceName, IResourceLocation fallback, string javascriptCondition)
+        {
+            var text = new StringWriter();
+            var hw = new HtmlWriter(text, context);
+            RenderLink(fallback, hw, context, resourceName, false);
```
exyi, line 58: No blocker, but actually, is it required to exclude the defer attribute from the fallback script? As the attributes are copied into a new script tag without the `defer` attribute, it should have no effect, right?

exyi (approved):

PR #946, opened 2021-02-23
Improved support for extension methods

acizmarik:
This PR improves experience when working with extension methods. The main changes are:

- Directive `@import` should be used to define which namespaces should be searched for extension methods
- Namespace `System.Linq` was added to default imports
- Introducing `ExtensionMethodsCache` to cache found extension methods for individual namespaces

Method overload resolving was not changed


src/DotVVM.Framework/Compilation/ExtensionMethodsCache.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.assemblyCache = assemblyCache;
+            this.methodsCache = new ConcurrentDictionary<string, List<MethodInfo>>();
+        }
+
+        public IEnumerable<MethodInfo> GetExtensionsForNamespace(string @namespace)
+        {
+            if (!methodsCache.TryGetValue(@namespace, out var extensions))
+                return CreateExtensionsForNamespace(@namespace);
+
+            return extensions;
```
exyi, line 28:
```suggestion
            return methodsCache.GetOrAdd(@namespace, CreateExtensionsForNamespace));
```

;)

src/DotVVM.Framework/Compilation/ExtensionMethodsCache.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace DotVVM.Framework.Compilation
+{
+    public class ExtensionMethodsCache
+    {
+        private readonly CompiledAssemblyCache assemblyCache;
+        private readonly ConcurrentDictionary<string, List<MethodInfo>> methodsCache;
```
exyi, line 15: I'd prefer to cache ImmutableArray to prevent accidental cache poisoning by changing the returned List.

exyi (approved):
Very nice :) Please, just change the cached type and it's perfect.

PR #947, opened 2021-02-23
Fixed empty string coercion in nullable types

quigamdev:
This PR fixes the quite unfortunate behavior of setters of Knockout Observables. If a value is set to observable then KO observable contains not coerced value for a moment. This causes problems in controls that expect valid types because `notifySubscribers()` is called before the coercion. The new implementation taking coerced value from a validator and then assigns it.


exyi (approved):
Good. Just please add tests that demonstrate the problems you are fixing.

PR #948, opened 2021-02-25
Dictionary serialization fix

quigamdev:
In previous versions of DotVVM, you could use `Dictionary<T1,T2>` in viewmodel and DotVVM was ok with it. It was not accessible from View, but it did work. At this moment, type validation on the javascript side prevents usage of it. 

This PR adds a translation of `Dictionary<,>` to `{ type: "dynamic" }`. So viewmodel type validation passes this case.


exyi:
It did not work and it will not work properly. If you what to support it, make sure this works:

* Removing an element (in a command and staticCommand). Also try with viewmodel server-side cache
* If you want to access it in the knockout observables, also try if a new element exists after it is added.

We could serialize it as base64 encoded string, that would be easy and would "work".

tomasherceg:
This is not the final solution - it is just a quick fix to prevent a breaking change in Business Pack and maybe other projects that use Dictionary (although it is not supported). 
We need the Dictionary to be readable on the client-side because some BP controls are accessing it. 

We've already discussed proper support for Dictionaries with @quigamdev - right now we just need to keep the serialized format on the client as is because someone may rely on it, and we just need the coercer to stay away from it.

I'll create a separate issue for the proper Dictionary support.

src/DotVVM.Framework/ViewModel/Serialization/ViewModelTypeMetadataSerializer.cs
```diff
@@ -163,6 +163,10 @@ internal JToken GetTypeIdentifier(Type type, HashSet<Type> dependentObjectTypes,
             {
                 return new JObject(new JProperty("type", "dynamic"));
             }
+            else if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Dictionary<,>))
```
tomasherceg, line 166: I'd rather check if the type implements `IDictionary<,>` - not sure if all people use concrete `Dictionary`.

exyi:
> I'll create a separate issue for the proper Dictionary support.

I don't think there is a relatively easy way to add support that is compatible with the current serialized format and is not broken. This change is just ensuring that the broken software stays broken.

We assume on multiple places (the state-observable mapper, diffing) that new properties do not appear and that properties don't disappear. Some use cases might still be broken due to new usages of this assumption in the state-observable mapper.

tomasherceg:
@exyi I agree that we'll have to change the serialization format - basically, we'll need to serialize `Dictionary` as `List<KeyValuePair<...>>` or something similar, and provide some JS translations so the users can use indexer, maybe `ContainsKey` and other methods. Some simple API on the JS side would also be nice.

And of course, we'll need to reimplement all BP components to reflect that change.

tomasherceg (approved):

src/DotVVM.Framework/ViewModel/Serialization/ViewModelSerializationMapper.cs
```diff
@@ -47,6 +47,7 @@ protected virtual ViewModelSerializationMap CreateMap(Type type)
         /// </summary>
         protected virtual IEnumerable<ViewModelPropertyMap> GetProperties(Type type)
         {
+            var isSetterSupported = type.IsGenericType && type.GetGenericTypeDefinition() == typeof(KeyValuePair<,>);
```
tomasherceg, line 50: I'd rather extract this as a separate method that would get a specific property - this decision is not made on the type level but on a property level.
Also, we should make this method virtual so it can be overridden if needed.

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
```
tomasherceg, line 1: Please add unit/integration tests on dictionary serialization - I know that there is a UI test, but it cannot check the exact serialization format. We should already have some serialization tests, so this should be another case.

src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -274,9 +274,43 @@ public static bool IsEnumerable(Type type)
...
+                    Type interfaceDefinition = type.GetGenericTypeDefinition();
+
+                    if (genericInterfaceDefinition == interfaceDefinition)
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            foreach (Type i in type.GetInterfaces())
```
exyi, line 299:
```suggestion
            foreach (Type i in type.GetInterfaces().Append(type))
```

src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -274,9 +274,43 @@ public static bool IsEnumerable(Type type)
...
+                    Type interfaceDefinition = i.GetGenericTypeDefinition();
 
+                    if (genericInterfaceDefinition == interfaceDefinition)
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
```
exyi, line 312: Or maybe

```suggestion
            return type.GetInterfaces().Append(type).
                       .Where(t => t.IsGenericType)
                       .Select(t => t.GetGenericTypeDefinition())
                       .Contains(genericInterfaceDefinition)
```

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                var dict = value as IDictionary;
+
+                var attrs = value.GetType().GetGenericArguments();
+                var keyValuePair = typeof(KeyValuePair<,>).MakeGenericType(attrs);
+                var listType = typeof(List<>).MakeGenericType(keyValuePair);
+
+
+                var keyEnumerator = dict.Keys.GetEnumerator();
+                var valuesEnumerator = dict.Values.GetEnumerator();
```
exyi, line 29: I don't think there is a guarantee that the order of keys and values must be the same. Why not dict.GetEnumerator?

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+
+
+                var list = Activator.CreateInstance(listType);
+                var invokeMethod = listType.GetMethod(nameof(List<object>.Add));
+                while (keyEnumerator.MoveNext() && valuesEnumerator.MoveNext())
+                {
+                    var item = Activator.CreateInstance(keyValuePair, keyEnumerator.Current, valuesEnumerator.Current);
+                    invokeMethod.Invoke(list, new[] { item });
+                }
```
exyi, line 38: I think this could be simplified using `dynamic` (not checked it works, though):

```suggestion
                KeyValuePair<T, U>[] toArray<T, U>(IEnumerable<KeyValuePair<T, U>> x) =>
                       x.ToArray()
                object list = toArray((dynamic) value);
```


It will resolve the generic parameters by itself.

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            else if (reader.TokenType == JsonToken.StartArray)
+            {
+
+                var attrs = objectType.GetGenericArguments();
+                var keyValuePair = typeof(KeyValuePair<,>).MakeGenericType(attrs);
+                var listType = typeof(List<>).MakeGenericType(keyValuePair);
+
+                var dict = existingValue as IDictionary;
+                dict ??= (IDictionary)Activator.CreateInstance(objectType);
```
exyi, line 58: This would be problematic if the `objectType` is `IDictionary<...>`.

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
quigamdev: I just didn't want to use another reflection, but it makes sense to use just dict.GetEnumerator

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
quigamdev: I have tried and unfortunately does not work.

src/DotVVM.Framework/ViewModel/Serialization/DotvvmDictionaryConverter.cs
quigamdev: I have added condition `typeof(IDictionary).IsAssignableFrom(objectType)` to `bool CanConvert(Type objectType)` that prevents (de)serialization of instance that implements only `IDictionary<,>`. I am not even sure whether that is something we want to support until standard Dictionary<,> is fully supported.

exyi (approved):
Nice

PR #949, opened 2021-02-25
Upgraded RIGANTI Selenium testing framework for UI tests 

quigamdev:
The new version of Riganti.Selenium has changed the behavior of all `AssertUI` functions. 
These functions have a `WaitFor` block inside of them now. So it's not required to write WaitFor block around every `AssertUI.Something()`.


quigamdev:
Also, I have removed `Microsoft.VisualStudio.TestTools.UnitTesting` dependency and used standard `xunit.Assert` functions.

exyi (approved):

PR #952, opened 2021-02-28
NPM package with typescript definitions 

quigamdev:
Adds npm package definition for publication of typescript definitions of DotVVM.


exyi (approved):
Ok.

PR #955, opened 2021-03-01
Fixed coercion issue on observable assign and added state/patch API to observables 

quigamdev:
This PR adds API to observables that allow direct manipulation with an immutable state.
Added: 
* state prop 
* patchState method
* setState method

Also, this issue fixes a coercion issue on observable assign.


src/DotVVM.Framework/Resources/Scripts/tests/stateManagement.test.ts
```diff
@@ -141,7 +142,7 @@ test("Dirty flag", () => {
 })
 
 test("Upgrade null to observableArray", () => {
-    s.update(x => ({...x, ArrayWillBe: [ { $type: "t5", B: "ahoj" } ] }))
+    s.update(x => ({ ...x, ArrayWillBe: [{ $type: "t5", B: "ahoj" }] }))
```
exyi, line 145: I know my rants on formatting just hide the important stuff, but these pointless formatting changes also hide the actual changes in the diff and clutter the history.

src/DotVVM.Framework/Resources/Scripts/state-manager.ts
```diff
@@ -240,19 +241,35 @@ function createObservableObject<T extends object>(initialObject: T, typeHint: Ty
 function createWrappedObservable<T>(initialValue: T, typeHint: TypeDefinition | undefined, updater: UpdateDispatcher<T>): DeepKnockoutWrapped<T> {
 
     let isUpdating = false
+    let isDirty = false
+    let isLastErrorUpdate = false
+
+    function triggerLastSetErrorUpdate(obs: KnockoutObservable<T>) {
+        try {
+            isLastErrorUpdate = true;
+            obs.valueHasMutated && obs.valueHasMutated();
```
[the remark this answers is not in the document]
tomasherceg: At least type definitions say so: https://github.com/riganti/dotvvm/blob/viewmodel-types/src/DotVVM.Framework/Resources/Scripts/typings/knockout/knockout.d.ts#L115
Maybe it's safe to assume that it'll be there always.

src/DotVVM.Framework/Resources/Scripts/state-manager.ts
```diff
@@ -240,19 +241,35 @@ function createObservableObject<T extends object>(initialObject: T, typeHint: Ty
 function createWrappedObservable<T>(initialValue: T, typeHint: TypeDefinition | undefined, updater: UpdateDispatcher<T>): DeepKnockoutWrapped<T> {
 
     let isUpdating = false
+    let isDirty = false
+    let isLastErrorUpdate = false
```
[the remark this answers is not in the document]
tomasherceg: The problem here was that the handler was called twice in some cases. Hopefully, there is a better solution.

src/DotVVM.Framework/Resources/Scripts/tests/stateManagement.test.ts
tomasherceg: Sorry about that, this wasn't intentional.

src/DotVVM.Framework/Resources/Scripts/binding-handlers/introduce-alias.ts
```diff
@@ -21,6 +21,11 @@ function createWrapperComputed<T>(accessor: () => KnockoutObservable<T> | T, pro
         }
     });
     computed["wrappedProperty"] = accessor;
+    Object.defineProperty(computed, "state", {
+        get: () => (accessor() as any || {})["state"],
+        configurable: false,
+        enumerable: false
+    })
```
exyi, line 28: We should probably also clone the other properties like `setState`.

exyi (approved):
See the comment. I don't really understand what is going on with the TextBox bindings, but that should at least be well covered by UI tests. Otherwise, it seems fine.

src/DotVVM.Framework/Resources/Scripts/binding-handlers/introduce-alias.ts
quigamdev: I will create a new PR that will add other props.

PR #956, opened 2021-03-01
Fixed typescript definition for SPA (#2)

quigamdev:


src/DotVVM.Framework/Resources/Scripts/dotvvm-root.ts
```diff
@@ -74,7 +74,7 @@ const dotvvmExports = {
     isPostbackRunning,
     events: (compileConstants.isSpa ?
              { ...events, ...spaEvents } :
-             events),
+             events) as (typeof events | (typeof spaEvents & typeof events)),
```
exyi, line 77: Just note that `Partial<typeof spaEvents> & typeof events` would also be an option, maybe it could be more comfortable to use (the `!` will suppress any warning that the spa event does not need to exist)

exyi (approved):

exyi (approved):

PR #957, opened 2021-03-02
Fix concurrency handler for staticCommands

exyi:
The problem was that the static command commit was always rejected,
because it was not started as postback

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/postback/internal-handlers.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/postback/internal-handlers.ts
@@ -95,8 +95,10 @@ export const suppressOnUpdating = (o: any) => ({
     }
 })
 
-export function isPostbackStillActive(id: number) {
-    return getLastStartedPostbackId() == id && !gate.isPostbackDisabled(id)
+export function isPostbackStillActive(options: PostbackOptions) {
+    const id = options.postbackId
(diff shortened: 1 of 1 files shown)


src/DotVVM.Framework/Resources/Scripts/postback/internal-handlers.ts
```diff
@@ -95,8 +95,10 @@ export const suppressOnUpdating = (o: any) => ({
     }
 })
 
-export function isPostbackStillActive(id: number) {
-    return getLastStartedPostbackId() == id && !gate.isPostbackDisabled(id)
+export function isPostbackStillActive(options: PostbackOptions) {
+    const id = options.postbackId
+    // For postback and SPA navigation, we reject it if another postback has already started to prevent flashes of new data that will be soon overridden anyway.
+    return (options.commandType == "staticCommand" || getLastStartedPostbackId()) == id && !gate.isPostbackDisabled(id)
```
quigamdev, line 101: In this case, StaticCommand will be still rejected due to the fact that  e.g. 12 != true. 
Didn't you mean ` (options.commandType == "staticCommand" || getLastStartedPostbackId() == id) && !gate.isPostbackDisabled(id)` ?

quigamdev (changes requested):

quigamdev (approved):

PR #962, opened 2021-03-05
 Try to use setState in deserialize 

exyi:


tomasherceg:
Looks like it won't be as easy as we thought.

Issue #831, opened 2020-05-01
Suggestion for JS translations available in command expressions


tomasherceg:
We've talked about adding more JS translations in the command expressions, so the users can do more especially in static command bindings - e.g. manipulate with arrays and collections, dates etc.

Here are some ideas of what we can add:

## DateTime
- `Year`, `Month`, `Day`, `Hour`, `Minute`, `Second`, `Millisecond`, `DayOfWeek`, `DayOfYear`
- `AddYears(double)`, `AddMonths(double)`, `AddDays(double)`, `AddHours(double)`, `AddMinutes(double)`, `AddSeconds(double)`, `AddMilliseconds(double)`

## String
- `Format(string, params object[])`
- `Join(string, params string[])`
- `Split(char)`
- `Replace(string, string)`
- `Contains(string)`, `StartsWith(string)`, `EndsWith(string)`
- `IndexOf(string)`, `IndexOf(string, int)`, `LastIndexOf(string)`, `LastIndexOf(string, int)`

## Math functions
- `Abs(double)`, `Abs(int)`
- `Ceiling(double)`, `Floor(double)`, `Round(double, int)`
- `Min(double, double)`, `Min(int, int)`, `Max(double, double)`, `Max(int, int)`
- `Sign(double)`, `Sign(int)`

## List<T>
- `Add(T)`, `AddRange(T)`
- `First()`, `Last()`
- `RemoveAt(int)`, `RemoveRange(int, int)`
- custom extension methods `AddOrUpdateItem<TSelectorValue>(string propertySelector, TSelectorValue, T)`, `RemoveItem(string propertySelector, TSelectorValue)`
- if we could add support for lambdas, we could use x => x.Prop instead of string selector in the previous methods, and we could also add quite a lot of LINQ methods in that case (this can be phase 2)

## WebUtility
- `UrlEncode(string)`, `UrlDecode(string)`
- `HtmlEncode(string)`, `HtmlDecode(string)`

Also, there are quite a lot of methods that we could support provided we can support new types - especially `DateTimeOffset`, `TimeSpan`, maybe `Dictionary<K, V>`.

exyi:
On 2020-05-01 17:54, Tomáš Herceg wrote:
>     DateTime
> 
>   * |Year|, |Month|, |Day|, |Hour|, |Minute|, |Second|, |Millisecond|,
>     |DayOfWeek|, |DayOfYear|
[...]

PR #965, opened 2021-03-10
Added more JavaScript translations

acizmarik:
As suggested by #831, this PR adds various translations for methods on `Enumerable`, `String` and `Math` classes.

### Enumerable

* `Enumerable.All(source, predicate)`
* `Enumerable.Any(source)`
* `Enumerable.Any(source, predicate)`
* `Enumerable.Concat(first, second)`
* `Enumerable.Distinct(source)` 
   * restricted to primitive types
   * attempt to distinct complex types ends with compilation exception
* `Enumerable.FirstOrDefault(source)`
* `Enumerable.FirstOrDefault(source, predicate)`
* `Enumerable.LastOrDefault(source)`
* `Enumerable.LastOrDefault(source, predicate)`
* `Enumerable.Max(source)`
* `Enumerable.Max(source, selector)`
   * restricted to primitive types
   * overload `Max<TSource,TResult>(IEnumerable<TSource>, Func<TSource,TResult>)` is not supported
* `Enumerable.Min(source)`
* `Enumerable.Min(source, selector)`
   * restricted to primitive types
   * overload `Min<TSource,TResult>(IEnumerable<TSource>, Func<TSource,TResult>)` is not supported
* `Enumerable.OrderBy(source, selector)`
   * restricted to primitive types
   * attempt to order by complex type ends with compilation exception
* `Enumerable.OrderByDescending(source, selector)`
   * restricted to primitive types
   * attempt to order by complex type ends with compilation exception
* `Enumerable.Skip(source, count)`
* `Enumerable.Take(source, count)`
* `Enumerable.ToArray(source)`
* `Enumerable.ToList(source)`

### String
* `string.Contains(value)`
* `string.EndsWith(value)`
* `string.IndexOf(value)`
* `string.IndexOf(value, startIndex)`
* `String.IsNullOrEmpty(value)`
* `string.Join(separator, stringArray)`
* `string.Join(separator, stringEnumerable)`
* `string.LastIndexOf(value)`
* `string.LastIndexOf(value, startIndex)`
* `string.Replace(old, new)`
* `string.Split(charSeparatorsArray)`
* `string.Split(separator, StringSplitOptions)`
   * `None` and `RemoveEmptyEntries` options are supported
* `string.StartsWith(value)`
* `string.ToLower()`
* `string.ToUpper()`

### Math
`Math.Abs`, `Math.Acos`, `Math.Asin`, `Math.Atan`, `Math.Atan2`, `Math.Ceiling`, `Math.Cos`, `Math.Cosh`, `Math.Exp`, `Math.Floor`, `Math.Log`, `Math.Log10`, `Math.Max`, `Math.Min`, `Math.Pow`, `Math.Round`, `Math.Sign`, `Math.Sin`, `Math.Sinh`, `Math.Sqrt`, `Math.Tan`, `Math.Tanh`, `Math.Trunc`


src/DotVVM.Framework/Resources/Scripts/collections/arrayHelper.ts
```diff
@@ -0,0 +1,122 @@
...
+    for (let i = 0; i < items.length; i++) {
+        action(items[i], i);
+    }
+}
+
+function lastOrDefault<T>(source: T[], predicate: (s: T) => boolean): T | null {
+    let lastSatisfyingElement = null;
+    for (let i = 0; i < source.length; i++) {
+        if (predicate(source[i])) {
+            lastSatisfyingElement = source[i];
```
quigamdev, line 72: Just from a performance standpoint, isn't it better to reverse the `for` statement and iterate backward and return the first result?

src/DotVVM.Framework/Resources/Scripts/collections/arrayHelper.ts
```diff
@@ -0,0 +1,122 @@
...
+function clear<T>(source: T[]): void {
+    source.splice(0, source.length);
+}
+
+function distinct<T>(source: T[]): T[] {
+    let r = [];
+    for (let i = 0; i < source.length; i++) {
+        let found = false;
+        for (var j = 0; j < r.length; j++) {
+            if (r[j] == source[i]) {
```
quigamdev, line 41: I am not sure about this comparison. I would rather pass a comparer to this function as an optional parameter. In case the comparer is not specified, default compared would be used.

src/DotVVM.Framework/Resources/Scripts/collections/arrayHelper.ts
acizmarik: Yes it is. I copied the implementation from `firstOrDefault` and did not realize I can easily optimize this. I will push a fix for this

src/DotVVM.Framework/Resources/Scripts/collections/arrayHelper.ts
acizmarik: Yes, there is no overload with lambda for `Distinct`. Moreover, I added a compilation check that ensures we pass only collections of primitive types. Hopefully, the default comparer should work on primitive types without any problems.

exyi:
Could you please cherry-pick the IE11 fixes to a separate PR, so we can merge it right away?

src/DotVVM.Framework/Resources/Scripts/collections/arrayHelper.ts
quigamdev: I know that there is not an overload that could provide the comparer. In fact, I have meant that when this is a public JS API, it would be nice to have the option to use also comparer. I think we can keep the `==` comparison for now. If there is a use-case for a more sophisticated comparison then I would add it.

quigamdev (approved):

quigamdev (approved):

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
```diff
@@ -187,23 +191,207 @@ public void AddDefaultMethodTranslators()
...
+            AddMethodTranslator(joinStringArrayMethod, translator: new GenericMethodCompiler(args => args[2].Member("join").Invoke(args[1])));
+            var joinStringEnumerableMethod = typeof(string).GetMethods(BindingFlags.Public | BindingFlags.Static)
+                .Where(m => m.Name == nameof(string.Join) && m.GetParameters().Length == 2 && m.GetParameters().Last().ParameterType == typeof(IEnumerable<string>) && m.GetParameters().First().ParameterType == typeof(string)).Single();
+            AddMethodTranslator(joinStringEnumerableMethod, translator: new GenericMethodCompiler(args => args[2].Member("join").Invoke(args[1])));
+
+            AddMethodTranslator(typeof(string), nameof(string.Replace), parameters: new[] { typeof(string), typeof(string) }, translator: new GenericMethodCompiler(
+                args => args[0].Member("split").Invoke(args[1]).Member("join").Invoke(args[2])));
+            AddMethodTranslator(typeof(string), nameof(string.Split), parameters: new[] { typeof(char), typeof(StringSplitOptions) }, translator: new GenericMethodCompiler(
```
quigamdev, line 242: This method is not available for .NET Framework.

quigamdev (approved):
Nice

PR #969, opened 2021-03-12
Better error message when state is used before init

exyi:
Previously this caused error like "v is undefined", now it says that "DotVVM is not initialized".


quigamdev (approved):

quigamdev:
Tests have passed, so I think we can merge it.

PR #970, opened 2021-03-14
Non-minified version of DotVVM in debug mode

tomasherceg:
This PR uses non-minified versions of DotVVM and Knockout when `configuration.Debug` is `true`.


exyi (approved):
Nice.

quigamdev (approved):

PR #976, opened 2021-03-16
Improved error reporting for corrupted master page

acizmarik:
This PR fixes weird DotVVM error reported by @martindybal. If user makes a mistake while specifying viewmodel (by referencing a non-existing class) for their master page, they receive a somewhat confusing error message.

image

The problem:
- DotVVM finds out that the specified viewmodel could not be found https://github.com/riganti/dotvvm/blob/7dc7e9d0cda757f953545f1f47e85ce0fb883662/src/DotVVM.Framework/Compilation/ControlTree/Resolved/ResolvedTreeBuilder.cs#L149
- However this information gets stripped and is not propagated further https://github.com/riganti/dotvvm/blob/7dc7e9d0cda757f953545f1f47e85ce0fb883662/src/DotVVM.Framework/Compilation/ControlTree/Resolved/ResolvedTreeRoot.cs#L29

If we were to add those errors there, I am not sure whether it could break something. In the meantime, I created a simple fix that should not break anything and provides users with a more specific error message.

### DIFF
--- a/src/DotVVM.Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
+++ b/src/DotVVM.Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
@@ -208,7 +208,14 @@ namespace DotVVM.Framework.Compilation.ControlTree
             if (masterPage == null)
                 return;
             var viewModel = root.DataContextTypeStack.DataContextType;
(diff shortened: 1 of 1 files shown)


exyi (approved):
Ok.

PR #978, opened 2021-03-18
Fix generics resolving in method overloads search (2)

### LINKED ISSUE #942: Resolving types in method overloads search
As @exyi pointed out in #937, the PR only solved one specific error but the bug still persists.
See an [example](https://github.com/riganti/dotvvm/pull/937#pullrequestreview-590039888) where the current implementation fails.

acizmarik:
Resolves #942. This is a continuation of #937.

### DIFF
--- a/src/DotVVM.Framework/Compilation/Binding/MemberExpressionFactory.cs
+++ b/src/DotVVM.Framework/Compilation/Binding/MemberExpressionFactory.cs
@@ -408,12 +408,14 @@ namespace DotVVM.Framework.Compilation.Binding
                 }
                 else if (sgt.IsGenericType)
                 {
-                    Type[] genericArguments;
+                    Type[] genericArguments = null;
                     var expression = expressionTypes[i];
 
-                    // Arrays need to be handled in a special way to obtain instantiation
                     if (expression.IsArray)
+                    {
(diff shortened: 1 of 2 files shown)


src/DotVVM.Framework/Compilation/Binding/MemberExpressionFactory.cs
```diff
@@ -408,17 +408,53 @@ private Type GetGenericParameterType(Type genericArg, Type[] searchedGenericType
...
+                    {
+                        if (expression.IsGenericType && sgt.GetGenericTypeDefinition() == expression.GetGenericTypeDefinition())
+                        {
+                            // We have exactly the same type => return generic arguments
+                            genericArguments = expression.GetGenericArguments();
+                        }
+                        else if (sgt.IsInterface)
+                        {
+                            // We must find the instantiation within an implemented generic interface
+                            var implementation = expression.GetInterfaces().Where(ifc => ifc.IsGenericType && ifc.GetGenericTypeDefinition() == sgt.GetGenericTypeDefinition()).Take(2).ToList();
```
cafstep, line 429: Is it actually possible for `expression` to implement more than one interface that is equal to the generic type definition of `sgt`? If such a thing ever happened, wouldn't it mean that something went horribly wrong and would warrant an exception?

Issue #926, opened 2021-01-15
IUploadedFileStorage should be in DotVVM.Core nuget package


martindybal:
`IUploadedFileStorage` is located in `DotVVM.Framework`. I think it should move moved to `DotVVM.Core`

https://github.com/riganti/dotvvm/blob/949d3b121be46069691ad133407dc6ff0c2128b7/src/DotVVM.Framework/Storage/IUploadedFileStorage.cs

tomasherceg:
Agree, there is also `IReturnedFileStorage`. 
I'll research if there are other things that we should move as well.

exyi:
This was done in the #981 PR

PR #981, opened 2021-03-21
IUploadedFileStorage and IReturnedFileStorage made async and moved to DotVVM.Core

### LINKED ISSUE #927: Add asynchronous support to IUploadedFileStorage
Actual implementation: 
```
    public interface IUploadedFileStorage
    {
        Task<Guid> StoreFile(Stream stream);
        void DeleteFile(Guid fileId);
        Stream GetFile(Guid fileId);
        void DeleteOldFiles(DateTime maxCreatedDate);
    }
```
https://github.com/riganti/dotvvm/blob/949d3b121be46069691ad133407dc6ff0c2128b7/src/DotVVM.Framework/Storage/IUploadedFileStorage.cs

I missing `CancellationToken` in `StoreFile`. And also other methods should be asynchronous, IO operation may slow down server. 
```
    public interface IUploadedFileStorage
    {
        Task<Guid> StoreFile(Stream stream, CancellationToken ct);
        Task DeleteFileAsync(Guid fileId, CancellationToken ct);
        Task<Stream> GetFile(Guid fileId, CancellationToken ct);
        Task DeleteOldFiles(DateTime maxCreatedDate, CancellationToken ct);
    }
```

It'll be breaking change, but it should be easy to fix and implement asynchronous version

tomasherceg:
Resolves #927 and #926 

I've added the `Async` suffix to all methods and made them async.

I've removed `DeleteOldFiles` method from the interfaces - it shouldn't have been there - it is implementation-specific and used only from filesystem implementation of these interfaces.


src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -232,7 +233,7 @@ public static void ReturnFile(this IDotvvmRequestContext context, Stream stream,
             AttachmentDispositionType = attachmentDispositionType ?? "attachment"
         };
 
-        var generatedFileId = returnedFileStorage.StoreFile(stream, metadata).Result;
+        var generatedFileId = returnedFileStorage.StoreFileAsync(stream, metadata).Result;
```
exyi, line 236: Could we add async ReturnFile and deprecate this method? This does not seem right, maybe it's not very problematic, but at least there should be more correct way.

exyi (approved):
Why do we use the `Async` suffix? Usually we don't and usually it's not used when there is not synchronous variant.

Otherwise fine and it should go to v3, whatever the naming

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
quigamdev: @exyi  I have removed the suffix.

quigamdev: @exyi What do you mean by the "deprecate this method"? When the signature of parameters is the same as before then we cannot add an overload with a different return type and deprecate it. So I am not sure what you mean.

exyi: Ah, shit. In this case we'd need the async suffix... :(

quigamdev: @exyi Ok, reverting the last commit and merging this PR.

quigamdev (approved):

src/DotVVM.Framework/Hosting/DotvvmRequestContextExtensions.cs
exyi: I don't understand how that solves this issue, there is still synchronous wait in this method

PR #983, opened 2021-03-23
EventValidator - error message

quigamdev:


src/DotVVM.Framework/Runtime/Commands/EventValidator.cs
```diff
@@ -151,7 +151,7 @@ public FindBindingResult ValidateCommand(string[] path, string commandId, Dotvvm
 
             return new FindBindingResult
             {
-                ErrorMessage = bindingInPath ? errorMessage : "Nothing was found in found inside specified DataContext check if ViewModel is populated",
+                ErrorMessage = bindingInPath ? errorMessage : "Nothing was found inside specified DataContext check if ViewModel is populated.",
```
exyi, line 154:
```suggestion
                ErrorMessage = bindingInPath ? errorMessage : "Nothing was found inside specified DataContext. Please check if ViewModel is populated.",
```

Maybe we could also include the data context (type name) in the message.

PR #987, opened 2021-04-01
Added support for string interpolation in bindings

### LINKED ISSUE #362: String interpolation in bindings
As discussed in #195 it would be useful to have string interpolation support in value bindings - `$"Test {Id} {CreatedDate:yyyy/M/d}"`.

acizmarik:
Resolves #362. This PR adds support for string interpolation in bindings.

Interpolated strings are always translated into C# `String.Format(...)` calls, unless a constant string is provided with no interpolation expressions. When compiling `String.Format(...)` calls to JavaScript, the function `dotvvm.globalize.format(...)` is used.

Implementation progress:
- [x] Expression interpolation support
- [x] String formatting component
- [ ] ~~String alignment component~~ (we decided that this will not be supported)

### DIFF
--- a/src/DotVVM.Framework.Tests.Common/Binding/BindingCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests.Common/Binding/BindingCompilationTests.cs
@@ -173,10 +173,19 @@ namespace DotVVM.Framework.Tests.Binding
         [DataRow(@"$""Interpolated {StringProp} {StringProp}""", "Interpolated abc abc")]
         [DataRow(@"$'Interpolated {StringProp} {StringProp}'", "Interpolated abc abc")]
         [DataRow(@"$'Interpolated {StringProp.Length}'", "Interpolated 3")]
-        [DataRow(@"$'{string.Join(', ', IntArray)}'", "1, 2, 3")]
         public void BindingCompiler_Valid_InterpolatedString_WithSimpleExpressions(string expression, string evaluated)
         {
-            var viewModel = new TestViewModel() { StringProp = "abc", IntArray = new[] { 1, 2, 3 } };
+            var viewModel = new TestViewModel() { StringProp = "abc" };
+            var binding = ExecuteBinding(expression, viewModel);
+            Assert.AreEqual(evaluated, binding);
+        }
+
+        [TestMethod]
+        [DataRow(@"$'{string.Join(', ', IntArray)}'", "1, 2, 3")]
+        [DataRow(@"$'{string.Join(', ', 'abc', 'def', $'{string.Join(', ', IntArray)}')}'", "abc, def, 1, 2, 3")]
+        public void BindingCompiler_Valid_InterpolatedString_NestedExpressions(string expression, string evaluated)
+        {
+            var viewModel = new TestViewModel { IntArray = new[] { 1, 2, 3 } };
             var binding = ExecuteBinding(expression, viewModel);
             Assert.AreEqual(evaluated, binding);
         }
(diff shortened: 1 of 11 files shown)


src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
```diff
@@ -105,6 +105,31 @@ protected override Expression VisitLiteralExpression(LiteralExpressionBindingPar
             return Expression.Constant(node.Value);
         }
 
+        protected override Expression VisitInterpolatedStringExpression(InterpolatedStringBindingParserNode node)
+        {
+            var target = new MethodGroupExpression() {
+                MethodName = nameof(String.Format),
+                Target = Registry.Resolve(nameof(String))
```
exyi, line 112:
```suggestion
                Target = typeof(string)
```

If you'd define your own string, we still want to use System.String (as C# does: https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGgQMrABOYAdgOYEDeOBzRhy6ADAQLIAU1wAgA8AlAyYtJyAOwEAJACJ6QgL4KA3BOYqcKoA=)

src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
```diff
@@ -105,6 +105,31 @@ protected override Expression VisitLiteralExpression(LiteralExpressionBindingPar
...
+            };
+
+            if (node.Arguments.Any())
+            {
+                // Translate to a String.Format(...) call
+                var arguments = new Expression[node.Arguments.Count];
+                for (var index = 0; index < node.Arguments.Count; index++)
+                {
+                    arguments[index] = HandleErrors(node.Arguments[index], Visit)!;
+                }
```
exyi, line 122:
```suggestion
                var arguments = node.Arguments.Select(arg => HandleErrors(node.Arguments[index], Visit))[.ToArray()];
```

;)

src/DotVVM.Framework.Tests.Common/Binding/BindingCompilationTests.cs
```diff
@@ -130,6 +130,64 @@ public void BindingCompiler_Valid_StringLiteralInSingleQuotes()
...
+                Assert.AreEqual(typeof(BindingCompilationException), current.GetType());
+                StringAssert.Contains(current.Message, errorMessage);
+            }
+        }
+
+        [TestMethod]
+        [DataRow(@"$""Interpolated {StringProp} {StringProp}""", "Interpolated abc abc")]
+        [DataRow(@"$'Interpolated {StringProp} {StringProp}'", "Interpolated abc abc")]
+        [DataRow(@"$'Interpolated {StringProp.Length}'", "Interpolated 3")]
+        public void BindingCompiler_Valid_InterpolatedString_WithSimpleExpressions(string expression, string evaluated)
```
exyi, line 173: Could you please also test that nested string interpolations work? For example `$"{ string.Join(", ", MyArray) }"`, I think it may be broken now.

src/DotVVM.Framework/Compilation/Parser/Binding/Parser/BindingParser.cs
```diff
@@ -883,6 +898,94 @@ private static string ParseStringLiteral(string text, out string? error)
             return sb.ToString();
         }
 
+        private static (string, List<BindingParserNode>) ParseInterpolatedString(string text, out string? error)
+        {
+            text = ParseStringLiteral(text, out error).Substring(1);
```
exyi, line 903: You are parsing it as a standard string literal while it may actually contain nested expressions which may also contain `"`.

exyi (changes requested):
Pretty cool :) Just please have a look at the case with nested expressions.

acizmarik:
Thanks for review. You were right - nested expressions were broken if they contained strings. It should be fixed now. Also formatting should now be supported.
I will add more tests, do some refactoring and then will request one more review of the changes :)

PR #988, opened 2021-04-02
FileUpload control doesn't need an iframe

### LINKED ISSUE #984: Do not use iframe in FileUpload
Hi,

I need to add the option to upload a file in a form.
The built-in FileUpload control renders an invisible iframe, which means that it does not work in our environment, because the server always returns the header "X-Frame-Options: DENY" due to security requirements which are non-negotiable.
1. Is it possible to change FileUpload so it does not require the use of an iframe? (In general, iframes are not ideal. AJAX could and should be used instead today)
2. Is there an alternative method to upload files in DotVVM? (I also have another issue with the built-in control so it makes sense)

Thanks

tomasherceg:
Resolves #984


src/DotVVM.Framework/Controls/FileUpload.cs
```diff
@@ -220,13 +219,23 @@ private void RenderUploadButton(IHtmlWriter writer)
...
+        private void RenderInputControl(IHtmlWriter writer, IDotvvmRequestContext context)
         {
-            // render iframe
-            writer.AddAttribute("class", "dotvvm-upload-iframe");
-            writer.AddAttribute("src", GetFileUploadHandlerUrl());
-            writer.RenderBeginTag("iframe");
-            writer.RenderEndTag();
+            writer.AddStyleAttribute("display", "none");
+            writer.AddAttribute("type", "file");
+
```
exyi, line 226: The `GetFileUploadHandlerUrl` does not need to contain all the query parameters now, right?

exyi (approved):
Nice

src/DotVVM.Framework/Controls/FileUpload.cs
tomasherceg: These parameters are read by the upload middleware which checks the file types and size on the server. 
I believe it is because IE9 we supported in the past - it couldn't verify these things on the client.

Since these values are not protected in any way, there is not much sense in checking that on the server - anyone can post the file to a URL and change the parameters.

I think we have two options here:
1. We can check these things on the client (it's possible to do in IE11)
2. We can protect the attributes so they cannot be tampered with. But there is no way to distinguish through which file upload control the files were uploaded, so anyone can still take the less restrictive "protected" upload URL and upload the file. 

I prefer the first way - it shouldn't break anything.

exyi: > We can protect the attributes so they cannot be tampered with. But there is no way to distinguish through which file upload control the files were uploaded, so anyone can still take the less restrictive "protected" upload URL and upload the file.

This makes little sense, only gives false feel of security.

PR #989, opened 2021-04-03
Run UI tests in headless Chrome

quigamdev:
I have added capabilities to the chrome testing profiles that ensure that tests run under the headless version of chrome.
This change should save about 5-10 minutes of CI run time.


quigamdev:
I am closing this issue in favor of #990 Removing unnecessary waits and waitfor.
It does not make much sense to keep these two separate.

PR #994, opened 2021-04-13
Breaking change in TimeSpan fixed

tomasherceg:
We've introduced a breaking change in DotVVM 3.0 which causes JavaScript errors when `TimeSpan` is present in the viewmodel.

`TimeSpan` was not officially supported (and it still isn't) - we don't support formatting of `TimeSpan` values, and other things.

However, this PR "fixes" the behavior so `TimeSpan` can be used in the viewmodel and is not broken when sent back to the server.

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
@@ -183,12 +183,12 @@ function validateTime(value: any) {
     if (typeof value === "string") {
         // strict DotVVM format parse
         const parsedValue = serializationParseTime(value);
-        if (parsedValue) {
-            return { value: serializeTime(parsedValue, false) };
+        if (parsedValue !== null) {
+            return { value: serializeTime(parsedValue) };
         }
     }
     
-    if (value instanceof Date) {
-        return { value: serializeTime(value, false), wasCoerced: true };
+    if (typeof value === "number") {
+        return { value: serializeTime(value), wasCoerced: true };
     }
 }
--- a/src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
@@ -1,5 +1,5 @@
 import { formatString, parseDate as globalizeParseDate } from "../DotVVM.Globalize";
-import { parseDate as serializationParseDate, parseTime as serializationParseTime, serializeDate, serializeTime } from "../serialization/date";
+import { parseDate as serializationParseDate, parseTimeSpan as serializationParseTimeSpan, serializeDate, serializeTimeSpan } from "../serialization/date";
 import { CoerceError } from "../shared-classes";
 
 type PrimitiveTypes = { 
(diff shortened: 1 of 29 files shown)


src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
```diff
@@ -178,3 +178,17 @@ function validateDateTime(value: any) {
         return { value: serializeDate(value, false), wasCoerced: true };
     }
 }
+
+function validateTime(value: any) {
```
exyi, line 182:
```suggestion
function validateTimeSpan(value: any) {
```

exyi: I think that calling it TimeSpan even on client-side will be clearer

src/DotVVM.Framework/Resources/Scripts/serialization/date.ts
```diff
@@ -1,14 +1,30 @@
 import { logWarning } from "../utils/logging";
 
+const timeSpanBaseDate = new Date("0001-01-01 00:00:00.0000");
```
exyi, line 3: I don't understand what is going on, but I'm afraid this may break in some cases:
image

exyi: Aha, the year 0001 is parsed in Node in non-UTC timezone as year 2001 :facepalm: Do we have this covered with (unit) tests?

src/DotVVM.Framework/Resources/Scripts/serialization/date.ts
```diff
@@ -61,9 +77,22 @@ export function serializeTime(date: string | Date | null, convertToUtc: boolean
...
-    const h = (date2.getTime() - new Date(1, 0, 1).getTime() / 1000 / 3600) | 0;
-    const mi = padNumber(date2.getMinutes(), 2);
-    const s = padNumber(date2.getSeconds(), 2);
-    const ms = padNumber(date2.getMilliseconds(), 3);
-    return `${h}:${mi}:${s}.${ms}0000`;
+    const ticks = Math.abs(date2.getTime() - timeSpanBaseDate.getTime()) | 0;
+    const hours = (ticks / 1000 / 3600) | 0;
+    const minutes = (ticks / 1000 / 60) | 0;
+    const seconds = (ticks / 1000) | 0;
+    const milliseconds = (ticks % 1000) | 0;
```
exyi, line 84: I'm not sure how often will this fail - esp. with regards to dayling saving time. Since it does not allow me to use 0001-01-01 as the base date, I use 2021-01-01. Your logic says that at 3:00 27.3. is h = 3, while at 3:00 28. 3. h = 2:

```


> (new Date("2021-03-28 03:00:00.0000").getTime() - new Date("2021-01-01 00:00:00.0000")) / 1000 / 3600 % 24
2
> (new Date("2021-03-27 03:00:00.0000").getTime() - new Date("2021-01-01 00:00:00.0000")) / 1000 / 3600 % 24
3
```

exyi: Actually I don't know what should be the result...

exyi (commented):
I hate date and time. I don't think this works the same as did before.

tomasherceg:
I've reworked this to not depend on Date - it was a stupid idea in general.
I've also added JS tests so valid formats of timespan that need to pass the validation, are now covered.

PR #995, opened 2021-04-14
Fixed issue with lambdas and System.Action

acizmarik:
This PR fixes an issue when compiling lambda expressions that need to be assigned to `System.Action<...>`. It is also necessary for the type inference feature #936.

### Current state
If user defines a lambda in a binding whose expected type is `Action<...>`, current implementation fails for all non-`void` expressions. DotVVM throws compilation exception because it is unable to find an implicit conversion to convert from `Func<...>` to `Action<...>`. This is due to the fact that every expression that has non-`void` return type gets compiled into a `Func<...>`.

This is different from C#, where it is possible to have the following:
```csharp
Action<string> action = (strArg) => MyStringProperty = strArg;
```

The expression above would be compiled into `Func<string, string>` in DotVVM because the assignment gets evaluated to a string.

### Proposed solution
It is necessary to propagate information about expected types to the `BindingExpressionBuilder`. This type information can be then used inside `ExpressionBuildingVisitor` to validate that the body is a valid statement and if so, make sure to generate `Action<...>` if required.

### Breaking change
Interface `IBindingExpressionBuilder` was altered by adding an optional parameter to its `Parse` method. Regardless, I do not think that this change could actually break someone.

### DIFF
--- a/src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
+++ b/src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 
 namespace DotVVM.Framework.Compilation.Binding
 {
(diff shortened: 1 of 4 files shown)


acizmarik:
I just realized there is a bigger problem with lambdas and `System.Action`. I am changing to draft until resolved.

src/DotVVM.Framework/Compilation/Binding/ExpressionBuildingVisitor.cs
```diff
@@ -329,6 +335,28 @@ protected override Expression VisitLambdaParameter(LambdaParameterBindingParserN
             return Expression.Parameter(parameterType, node.Name.ToDisplayString());
         }
 
+        private Expression CreateLambdaExpression(Expression body, ParameterExpression[] parameters)
+        {
+            // Check if we have type information available
+            if (expressionDepth == 1 && ExpectedType != null && ReflectionUtils.IsDelegate(ExpectedType))
```
exyi, line 341: The `ExpectedType` is the top level expected type. However, the lambda may be used somewhere inside of the expression. ExpectedType might be string, while the expression is `MyMethod(x => Prop = x)` and it will lead to the same problem with `Action` and `Func`

exyi (changes requested):

acizmarik:
Closing in favor of #936. As pointed out by @exyi, this implementation does not fully fix the problem. PR #936 should resolve all mentioned issues together with the introduction of lambdas type inference.

PR #1004, opened 2021-04-22
Fixed coercer errors when DateTimeOffset is used in the viewmodel

tomasherceg:
This PR fixes client-side errors in coercer that occurred when `DateTimeOffset` is used in the viewmodel. 

**Please note that DateTimeOffset still doesn't work correctly - it loses the offset information when deserialized.** 
However, there is no client-side error when the page is loaded, so the behavior is the same as it was in DotVVM 2.x.


src/DotVVM.Framework/Resources/Scripts/metadata/primitiveTypes.ts
```diff
@@ -192,3 +194,17 @@ function validateTimeSpan(value: any) {
...
+
+function validateDateTimeOffset(value: any) {
+    if (typeof value === "string") {
+        // strict DotVVM format parse
+        if (serializationParseDateTimeOffset(value)) {
+            return { value };
+        }        
+    }
+    
+    // TODO: support conversion to date
```
exyi, line 206: I don't think we should ever have support for conversion of Date to DateTimeOffset, since Date does not have the offset information

tomasherceg: Agree - we have to keep our string-based format in the viewmodel so the offset information is persisted. 
However, if anyone wants to do something with the value on the client, they'll probably need some function that will be able to parse this value - e. g. return Date and offset, or better return our DateTimeOffset representation.

PR #1005, opened 2021-04-23
Fixed issue with underlying enum values, introduced new API

acizmarik:
This PR contains following changes:

- `OrderBy` and `OrderByDescending` translations now correctly resolve enum values that do not have string identifiers
- **New API** to simplify working with metadata and enums in TypeScript
   - `function getMetadataInfo(viewModel: any, value: any): TypeMetadata | null`
   - `function getEnumMetadataInfo(viewModel: any, value: any): EnumTypeMetadata | null`
   - `function getEnumValue(viewModel: any, value: any): number | null`


quigamdev (approved):
Looks good to me.

acizmarik:
Closing in favor of #1012 and #1013

PR #1012, opened 2021-05-04
Fixed issues with OrderBy(Descending) and simplified their translations

acizmarik:
It was possible to simplify current implementation of `OrderBy` and `OrderByDescending`. This also resolved some previously discovered issues regarding comparing by value vs by reference.


exyi (approved):
Ok, this is good

PR #1013, opened 2021-05-04
Added helper functions for working with type metadata

acizmarik:
DotVVM previously lacked better support for accessing type metadata. To resolve this issue, this PR introduces some helper functions into `dotvvm.metadata.*`

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
@@ -1,0 +1,40 @@
+﻿import { getTypeInfo } from "./typeMap";
+
+export function getTypeId(viewModel: any): string {
+    if (typeof viewModel !== "object") {
+        throw Error("Expected object but received: " + typeof viewModel);
+    }
+
+    return ko.unwrap(viewModel.$type);
+}
+
+export function getTypeMetadata(viewModel: any): TypeMetadata {
+    let typeId = getTypeId(viewModel);
+    return getTypeInfo(typeId);
+}
+
+export function getEnumMetadata(enumMetadataId: string): EnumTypeMetadata {
+    let metadata = getTypeInfo(enumMetadataId);
+    if (metadata.type !== "enum") {
+        throw new Error("Expected enum, but received object");
+    }
+
+    return metadata as EnumTypeMetadata;
+}
+
+export function getEnumValue(identifier: string | number, enumMetadataId: string): number | null {
+    let metadata = getEnumMetadata(enumMetadataId);
+    if (typeof identifier === "string") {
+        return metadata.values[identifier];
+    }
+    else {
+        // Ensure this number is not already defined
+        for (const key in metadata.values) {
+            if (metadata.values[key] === identifier) {
+                return null;
+            }
+        }
+
+        return identifier;
+    }
+}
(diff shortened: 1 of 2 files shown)


src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
```diff
@@ -0,0 +1,40 @@
+﻿import { getTypeInfo } from "./typeMap";
+
+export function getTypeId(viewModel: any): string {
```
exyi, line 3: We should either return `string | undefined` or check that the $type property actually exists. It just may not be there for some objects (for example if it comes from API bindings).

exyi, line 3:
```suggestion
export function getTypeId(viewModel: object): string {
```

src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (typeof viewModel !== "object") {
+        throw Error("Expected object but received: " + typeof viewModel);
+    }
+
+    return ko.unwrap(viewModel.$type);
+}
+
+export function getTypeMetadata(viewModel: any): TypeMetadata {
```
exyi, line 11: It would be better to accept a type id to be consistent with `getEnumMetadata`. If you think it's helpful to also accept an object, we can accept both - `viewModel: string | object`.

src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (typeof identifier === "string") {
+        return metadata.values[identifier];
+    }
+    else {
+        // Ensure this number is not already defined
+        for (const key in metadata.values) {
+            if (metadata.values[key] === identifier) {
+                return null;
+            }
+        }
```
exyi, line 36: I don't understand this check - why is it a problem if the number **is** defined?

exyi (changes requested):

src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
acizmarik: I prefer the variant with `string | undefined`

src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
acizmarik: Originally, I did not like the idea of passing arbitrary numbers - especially those that have defined string identifiers. But I agree it probably does not make much sense so I took it out.

src/DotVVM.Framework/Resources/Scripts/metadata/metadataHelper.ts
```diff
@@ -0,0 +1,28 @@
...
+export function getEnumMetadata(enumMetadataId: string): EnumTypeMetadata {
+    let metadata = getTypeInfo(enumMetadataId);
+    if (metadata.type !== "enum") {
+        throw new Error("Expected enum, but received object");
+    }
+
+    return metadata as EnumTypeMetadata;
+}
+
+export function getEnumValue(identifier: string | number, enumMetadataId: string): number | null {
```
exyi, line 20:
```suggestion
export function getEnumValue(identifier: string | number, enumMetadataId: string): number | undefined {
```

exyi (approved):

PR #1014, opened 2021-05-05
Added support for Dictionary<,> in viewModels

### LINKED ISSUE #1008: DotVVM 3 - Dictionary values not rendered
When I start new DotVVM project and change DotVVM version to 3, I cannot use dictionaries in value binding?

I have in **DefualtViewModel.cs** definition:
`public Dictionary<string, bool> Pravidla { get; set; } = new Dictionary<string, bool>() { {"key1", true }, { "key2", false } };`

In **Default.dothtml** I put:
`<div>{{value: _root.Pravidla["key1"]}} - {{value: _root.Pravidla["key2"]}}</div>`

After Rebuild and run the project, no dictionary values are rendered. 

In browser dev. tools I can see proper values in __dot_viewmodel_root element:
```
 "viewModel": {
    "$type": "14T6z33hWTxwPvI4Zryl151VX+s=",
    "Pravidla": [
      {
        "Key": "key1",
        "Value": true
      },
      {
        "Key": "key2",
        "Value": false
      }
    ],
    "Title": "Hello from DotVVM!", 
...
```
So why it is not rendered properly in DotVVM 3.02 while in DotVVM 2.5.2 it is working ok?
Am I doing something wrong?

acizmarik:
Resolves #1008. This PR adds support for dictionaries in viewModels. 

In order for this to work, it was necessary to change how we treat `IndexExpression`. Previously, accesses to `List<>` and `Dictionary<,>`  would be compiled as `{indexer}[index]`. This PR uses invocations of explicit get/set methods: `get_{indexer}(index)` or `set_{indexer}(index, value)`. This allows registering JS translations for these methods.


src/DotVVM.Framework/Resources/Scripts/collections/dictionaryHelper.ts
```diff
@@ -0,0 +1,26 @@
...
+        let keyValuePair = ko.unwrap(dictionary[index]);
+        if (ko.unwrap(keyValuePair.Key) == identifier) {
+            (keyValuePair.Value as any).setState(value);
+            return;
+        }
+    }
+
+    // Create new record if we did not find provided key
+    let patch = dictionary.map(function (e) { return {}; }).concat({ "Key": identifier, "Value": value });
+    observable.patchState(patch);
```
exyi, line 25: It would be better to work on the `observable.state`, and then call `.setState`. Also, the dictionary argument is then redundant.

```suggestion
    const dictionary = [... observable.state]
    for (let index = 0; index < dictionary.length; index++) {
        let keyValuePair = dictionary[index];
        if (keyValuePair.Key == identifier) {
            dictionary[index] = { Value: value, Key: keyValuePair.Key }
            observable.setState(dictionary)
            return;
        }
    }

    // Create new record if we did not find provided key
    dictionary.push({ "Key": identifier, "Value": value });
    observable.setState(dictionary);
```

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
```diff
@@ -116,11 +116,23 @@ public void AddDefaultMethodTranslators()
...
-            JsExpression indexer(JsExpression[] args, MethodInfo method) =>
+            JsExpression listIndexer(JsExpression[] args, MethodInfo method) =>
                 BuildIndexer(args[0], args[1], method.DeclaringType.GetProperty("Item"));
-            AddMethodTranslator(typeof(IList), "get_Item", new GenericMethodCompiler(indexer));
-            AddMethodTranslator(typeof(IList<>), "get_Item", new GenericMethodCompiler(indexer));
-            AddMethodTranslator(typeof(List<>), "get_Item", new GenericMethodCompiler(indexer));
+            JsExpression dictionaryGetIndexer(JsExpression[] args, MethodInfo method) =>
+                new JsIdentifierExpression("dotvvm").Member("dictionaryHelper").Member("getItem").Invoke(args[0], args[1]);
+            JsExpression dictionarySetIndexer(JsExpression[] args, MethodInfo method) =>
+                new JsIdentifierExpression("dotvvm").Member("dictionaryHelper").Member("setItem").Invoke(args[0].Clone(), args[1], args[2], args[0].EnsureObservableWrapped());
```
exyi, line 124: I'm not sure if this wouldn't better on cases like: `(C ? Dictionary1 : Dictionary2)["ss"] = X`

```suggestion
                new JsIdentifierExpression("dotvvm").Member("dictionaryHelper").Member("setItem").Invoke(args[0].Clone(), args[1], args[2], args[0].WithAnnotation<ResultShouldBeObservable>());
```

exyi (approved):

src/DotVVM.Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
acizmarik: I just tested original `setItem` implementation and the conditional expression works as expected in both cases (there is no `ResultShouldBeObservable` as suggested):

1. using `args[0].EnsureObservableWrapped()`
2. using `args[0].WithAnnotation(ShouldBeObservableAnnotation.Instance)`

quigamdev:
@acizmarik Could you please merge main into this branch and solve the merge conflicts before merging this PR?

PR #1016, opened 2021-05-06
Fixed RouteLinks with targets in SPA

### LINKED ISSUE #998: RouteLink don't work with target in Spa
`<dot:RouteLink RouteName="SecondPage" Text="Routelink with target=_blank" target="_blank" />`

I can't use dot:RouteLink with `target` in `SpaContentPlaceHolder`. Javascript blocks click on that link.

acizmarik:
Resolves #998


exyi (approved):
:+1:

PR #1018, opened 2021-05-11
Added support for flags enums which are serialized as a comma-separated list of strings

tomasherceg:
If an enum has the `Flags` attribute, it is serialized as a comma-separated list of entries by Newtonsoft JSON - e.g. `One, Two, Four`. 
This PR adds support for this enum format to the coercer.

Also, it means that we should probably introduce a client-side enum-comparison function and use it to compare enum values in the future releases.

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
@@ -74,17 +74,32 @@ function tryCoerceNullable(value: any, innerType: TypeDefinition, originalValue:
 function tryCoerceEnum(value: any, type: EnumTypeMetadata): CoerceResult {
     let wasCoerced = false;
     if (typeof value === "string") {
-        if (value in type.values) {
-          return { value };
-        } else if (value !== "") {
-            // number value as string
-            const numberValue = Number(value);
-            if (!isNaN(numberValue)) {
-                value = numberValue; 
-                wasCoerced = true;
-            }
+        if (value === "") {
+            return { value: 0, wasCoerced: true };
         }
-    } 
+
+        let parts = value.replace(/\s/g, "").split(',');
+        let coercedResult = 0;
+        for (let i = 0; i < parts.length; i++) {
+            if (parts[i] in type.values) {
+                coercedResult |= type.values[parts[i]];
+                continue;
+            } else if (parts[i] !== "") {
+                // number value as string
+                const numberValue = Number(parts[i]);
+                if (!isNaN(numberValue)) {
+                    coercedResult |= numberValue;
+                    wasCoerced = true;
+                    continue;
+                }
+            }
+            return new CoerceError(`Cannot cast '${parts[i]}' to type 'Enum(${keys(type.values).join(",")})'.`);
+        }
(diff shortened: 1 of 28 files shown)


src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
```diff
@@ -74,17 +74,32 @@ function tryCoerceNullable(value: any, innerType: TypeDefinition, originalValue:
...
     if (typeof value === "string") {
-        if (value in type.values) {
-          return { value };
-        } else if (value !== "") {
-            // number value as string
-            const numberValue = Number(value);
-            if (!isNaN(numberValue)) {
-                value = numberValue; 
-                wasCoerced = true;
+        if (value === "") {
```
exyi, line 77: This is quite a nasty edge case that could bite on non-flags enums. Could we support this only on flags enums that do not have a zero case?

src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
```diff
@@ -74,17 +74,32 @@ function tryCoerceNullable(value: any, innerType: TypeDefinition, originalValue:
...
+            return { value: 0, wasCoerced: true };
```
exyi, line 78: Also we should not produce `value: 0` in case there is a case like `None = 0`

src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
exyi: Actually, Newtonsoft.Json won't produce this empty string edge case when no flag is set: https://github.com/JamesNK/Newtonsoft.Json/blob/f7e7bd05d9280f17993500085202ff4ea150564a/Src/Newtonsoft.Json/Utilities/EnumUtils.cs#L203. It will probably produce value `0` or use a case like `None = 0`

tomasherceg:
I've re-implemented the matching for flags enums and tried to make sure that the coerced value is deterministic (items are ordered by underlying value). 
I hope that I haven't forgotten about any of the situations.

The tests are not passing because the CI was changed and doesn't work for 3.0 branch, but all tests passed locally.

acizmarik (approved):
Looks nice. Could not figure out any situation where this breaks

src/DotVVM.Framework/Resources/Scripts/metadata/coercer.ts
```diff
@@ -73,26 +73,106 @@ function tryCoerceNullable(value: any, innerType: TypeDefinition, originalValue:
...
+            if (reorderRequired) {
+                matched.sort((a, b) => type.values[a] - type.values[b]);
                 wasCoerced = true;
             }
+            if (wasCoerced) {
+                value = "";
+                for (let v of matched) {
+                    if (value !== "") value += ",";
+                    value += v;
+                }
```
exyi, line 129: Why not?

```suggestion
                value = matched.join(",")
```

exyi (approved):
It's a bit complicated, but probably necessary :/

Approved, just some nitpicks you can have a look at.

PR #1021, opened 2021-05-12
postbacks: add support for passing AbortSignal into the postback

exyi:


tomasherceg:
This looks nice, but we'll have to update the `fetch` polyfill for IE11 to support that. According to https://github.com/github/fetch it should be supported.

tomasherceg:
I've added missing polyfills for IE11 (there was a bunch of them needed) and now it should work.
The failed test on OWIN worked on the local machine, it's probably just a timing issue.

tomasherceg (approved):

tomasherceg:
I've reviewed all the tests and everything seems to work in IE (except for two randomly crashing tests but these are fine).

PR #1022, opened 2021-05-12
Warn when uppercase attribute is used as it was likely a non-existent property

exyi:


src/DotVVM.Framework/Controls/Infrastructure/BodyResourceLinks.cs
```diff
@@ -56,7 +56,7 @@ internal static string RenderWarnings(IDotvvmRequestContext context)
             foreach (var w in collector.GetWarnings())
             {
                 var msg = JsonConvert.ToString(w.ToString(), '"', StringEscapeHandling.EscapeHtml);
-                result += $"dotvvm.log.logWarning({msg});\n";
+                result += $"console.warn({msg});\n";
```
acizmarik, line 59: We should be using `dotvvm.log` wrappers

acizmarik (approved):

src/DotVVM.Framework/Controls/Infrastructure/BodyResourceLinks.cs
exyi: The problem is that this code runs before dotvvm is initialized.

PR #1023, opened 2021-05-13
Merge unit test projects together

exyi:
This will hopefully make it easier to run all the unit tests we have. And also cleans the solution a bit.


quigamdev (approved):
I have fixed just some small bugs. CI is green, so we can merge it.

PR #1031, opened 2021-05-26
Fixed issue with array and nullable type declarations in lambda parameters

### LINKED ISSUE #1024: Binding parser incorrectly interprets nullable types
DotVVM fails to parse nullable types in lambda parameters. 

**Example**: `MyNullableIntArray.Where((int? arg) => arg.Value > 0)`
**Bug**: `(int? arg)` gets interpreted as a malformed parenthesised conditional expression. Parser does not try to parse the expression as a lambda and generates an error

**Work-around**: `Nullable<int>` works

acizmarik:
Resolves #1024. Additionally, I found out that it was not possible to use array types too. This PR should fix these issues and makes it possible to use constructs such as `int?`, `int[]`, `int?[][]` as lambda parameters.


src/DotVVM.Framework/Compilation/Parser/Binding/Parser/TypeDeclarationBindingParserNode.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public enum TypeModifier
+    {
+        None,
+        Array,
+        Nullable,        
+    }
+
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class TypeDeclarationBindingParserNode : BindingParserNode
```
exyi, line 19: This types seems overloaded - it should be abstract without the Type and Next properties as these are only used in certain subtypes.

Also, I'd call this a type reference - definition would be a `class X { }` syntax.

I'd imagine hierarchy likes this:

```csharp

abstract class TypeReferenceBindingParserNode
sealed class ArrayTypeReferenceBindingParserNode(TypeReferenceBindingParserNode elementType) : TypeReferenceBindingParserNode
sealed class NullableTypeReferenceBindingParserNode(TypeReferenceBindingParserNode innerType) : TypeReferenceBindingParserNode
sealed class ActualTypeReferenceBindingParserNode(BindingParserNode type) : TypeReferenceBindingParserNode
```

Including generic types into this hierarchy would be nice.

acizmarik:
I implemented the suggested class hierarchy. I also reworked how we treat generics so that they can be handled in a similar way - there is now a `GenericTypeReferenceBindingParserNode`.

exyi (approved):

PR #1034, opened 2021-05-30
Updated JS translation registration for LINQ methods to work with .NET 6

tomasherceg:
.NET 6 added several new overloads to existing LINQ methods which break the registration of JS translators. 
I've added parameters filter overload which makes matching the methods easier.

This is working with .NET 6 Preview 4.


acizmarik (approved):
Nice. 
I also tested this together with the new list translations and everything seems to work on .NET 6 preview 4. Also once this is merged I can refactor some translation registrations to use this filtering

PR #1035, opened 2021-05-31
Fix control compiler deadlock

exyi:
I don't really know how to replicate it, @quigamdev could you please try if this fixes the issue?


quigamdev:
I will try it.

tomasherceg (approved):

PR #1037, opened 2021-06-03
Decorator support added for composite controls

tomasherceg:
When implementing wrappers for Fluent UI, I discovered many cases where it would be useful to be able to specify a "decorator" for the composite control.

The decorator can add additional properties, and wrap the generated control with some other code.
The first argument of the decorator is the result of the `GetContents` method, and the decorator must return the same type (single control or a collection).

```CSHARP
    public class EnabledCompositeControlDecorator
    {
        public static DotvvmControl DecorateControl(
            DotvvmControl control,

            [DefaultValue(null)] IValueBinding<bool> enabled
        )
        {
            if (enabled != null)
            {
                control.SetBinding(FormControls.EnabledProperty, enabled);
            }
            return control;
        }
    }
```

The decorators are added on the `CompositeControl` using an attribute - my plan is to be able to apply them on a base class.

```CSHARP
[CompositeControlDecorator(typeof(EnabledCompositeControlDecorator))]
```

### DIFF
--- a/src/DotVVM.Framework/Controls/CompositeControl.cs
+++ b/src/DotVVM.Framework/Controls/CompositeControl.cs
@@ -27,6 +27,12 @@ namespace DotVVM.Framework.Controls
         {
             public MethodInfo GetContentsMethod;
             public ImmutableArray<Func<IDotvvmRequestContext, CompositeControl, object>> Properties;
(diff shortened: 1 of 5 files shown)


tomasherceg:
We decided that we don't like this feature. :-(

PR #1038, opened 2021-06-04
TextBox doesn't show numeric value of 0

### LINKED ISSUE #1030: TextBox with Type Number and nullable integer bound does not allow zero
When binding `TextBox` with `Type="Number"` to a nullable integer and go to a zero through the up and down buttons it automatically sets the value to `null`.

ViewModel: 
```
public class DefaultViewModel : MasterPageViewModel
{
    public int? Value { get; set; }
}
```

View: 
```
<dot:TextBox Type="Number" Text="{value: Value}"/>
```

tomasherceg:
Resolves #1030


exyi (approved):
:+1:

PR #1009, opened 2021-04-27
Added JavaScript translations for List methods

acizmarik:
As suggested by #831, this PR adds various translations for methods on `List<T>`

### Regular `List<T>` methods

* `Add(element)`
* `AddRange(elements)`
* `Clear()`
* `Insert(index, element)`
* `InsertRange(index, elements)`
* `RemoveAt(index)`
* `RemoveAll(predicate)`
* `RemoveRange(index, count)`
* `Reverse()`

### Custom extensions
* `AddOrUpdate(element, matcher, updater)`
* `RemoveFirst(predicate)`
* `RemoveLast(predicate)`

PR #1039, opened 2021-06-04
Refactored JavaScript translations

acizmarik:
This PR introduces following changes:
- `arrayHelper`, `dictionaryHelper` and `stringHelper` were moved into `dotvvm.translations` and no longer pollute dotvvm root
- Registrations that previously required complex logic to select `MethodInfo` were rewritten using parameterFilter lambdas

#1009 should be merged first.

### DIFF
--- a/src/DotVVM.Framework/Resources/Scripts/dotvvm-root.ts
+++ b/src/DotVVM.Framework/Resources/Scripts/dotvvm-root.ts
@@ -124,9 +124,11 @@ const dotvvmExports = {
         logPostBackScriptError,
         level
     },
-    arrayHelper,
-    dictionaryHelper,
-    stringHelper
+    translations: {
+        arrayHelper,
+        dictionaryHelper,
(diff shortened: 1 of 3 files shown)


src/DotVVM.Framework/Resources/Scripts/dotvvm-root.ts
```diff
@@ -124,9 +124,11 @@ const dotvvmExports = {
         logPostBackScriptError,
         level
     },
-    arrayHelper,
-    dictionaryHelper,
-    stringHelper
+    translations: {
+        arrayHelper,
```
exyi, line 128: I think that when it's in the `translations` field, we could just call these `translations.array`, `translations.dictionary`, ...

Also, would be nice to add `as any` after this object, so it's not in the public typescript definitions, this is (mostly) not something users should use directly.

exyi (approved):

PR #1046, opened 2021-06-12
Configure Azure Pipelines (YAML) for Linux

cafour:
Adds and modifies scripts, configs, csprojs, and other files so that our C# unit tests, Jest tests, and AspNetCoreLatest UI tests can run on Linux in Azure Pipelines.

(The Windows version of the CI job will follow, once we (@MichalTichy and me) fix all the snags.)

### DIFF
--- a/ci/linux/run.sh
+++ b/ci/linux/run.sh
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/bin/bash
 
 # ================
 # argument parsing
@@ -10,7 +10,7 @@ LONGOPTS="help,\
     root:,config:,samples-profile:,samples-port:,samples-port-api:,\
     all,clean,npm-build,sln-restore,sln-build,unit-tests,js-tests,ui-tests,\
     no-all,no-clean,no-npm-build,no-sln-restore,no-sln-build,no-unit-tests,no-js-tests,no-ui-tests"
-TEMP=$(getopt -o "$SHORTOPS" -l "$LONGOPTS" -n "$PROGRAM" -- "$@")
+TEMP=$(getopt -o "$SHORTOPTS" -l "$LONGOPTS" -n "$PROGRAM" -- "$@")
 if [ $? -ne 0 ]; then
         exit 1
 fi
@@ -111,7 +111,7 @@ EOF
                 exit 1
             fi
 
-            eval $OPTION=$VALUE
+            eval "$OPTION=$VALUE"
             shift
             continue;
         ;;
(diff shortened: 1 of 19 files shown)


ci/linux/run.sh
```diff
@@ -0,0 +1,295 @@
...
+    echo "$(tput setaf 3)--------------------------------"
+    tput sgr0
+}
+
+function run_named_command {
+    NAME=$1
+    shift
+
+    print_header $NAME
+    tput setaf 4 && printf "running '$@'\n" && tput sgr0
```
acizmarik, line 172: Why use `printf` instead of `echo`? Also it would be safer to write something like `printf "running %s\n" "$@"`

ci/linux/run.sh
```diff
@@ -0,0 +1,295 @@
...
+    eval $@
```
acizmarik, line 173:
```suggestion
    eval "$@"
```

ci/linux/run.sh
```diff
@@ -0,0 +1,295 @@
...
+            OPTION=${OPTION^^}
+            OPTION=${OPTION//-/_}
+            eval test -n "\$$OPTION"
+            if [ $? -ne 0 ]; then
+                # this flag doesn't exit
+                echo >&2 "Option '$1' is not recognized."
+                exit 1
+            fi
+
+            eval $OPTION=$VALUE
```
acizmarik, line 114:
```suggestion
            eval "$OPTION"=$VALUE
```

ci/linux/run.sh
```diff
@@ -0,0 +1,295 @@
...
+# argument parsing
+# ================
+
+PROGRAM='run.sh'
+SHORTOPTS="h"
+LONGOPTS="help,\
+    root:,config:,samples-profile:,samples-port:,samples-port-api:,\
+    all,clean,npm-build,sln-restore,sln-build,unit-tests,js-tests,ui-tests,\
+    no-all,no-clean,no-npm-build,no-sln-restore,no-sln-build,no-unit-tests,no-js-tests,no-ui-tests"
+TEMP=$(getopt -o "$SHORTOPS" -l "$LONGOPTS" -n "$PROGRAM" -- "$@")
```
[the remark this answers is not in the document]
cafour: This was a typo. Nice catch. :)

ci/linux/run.sh
cafour: I don't recall why I decided to use `printf` but `echo` works just as nicely.

acizmarik (approved):
It seems a few tests are failing that we should probably look into. Other than that new pipelines look nice 👍

cafour:
Closed in favour of #1067.

PR #1050, opened 2021-06-16
Fixed typo in parameter name.

MichalTichy:
DotvvmBindableObject.GetAllAncestors parameter incudingThis renamed to includingThis.


exyi (approved):
Definitely makes sense. I'd leave for the 4.0 version, since this is a breaking change, although not very annoying to fix.

PR #1052, opened 2021-06-16
Fixed issue with assigning to arrays and lists

### LINKED ISSUE #977: Can not assign expression of type Index
It is impossible to assign to a list item in (static) command. Such code does not work:

```html
<dot:Button staticCommand: _parent.Tasks[_index] = FetchNewRowVersion()}" />

```

acizmarik:
Resolves #977. This PR fixes issues that occurred in staticCommands when assigning into lists and arrays.

**Changes**:
- **Lists**: Registered JS translation for set_Indexer on Lists and provided its implementation that updates observables
- **Arrays**: Assignments to array elements are compiled using an explicit call to `Array.SetValue(value, index)`. This step made it possible to register a JS translation for array element assignments. The typescript implementation is the shared with lists.

### DIFF
--- a/src/DotVVM.Framework.Tests/Binding/BindingCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests/Binding/BindingCompilationTests.cs
@@ -784,6 +784,22 @@ namespace DotVVM.Framework.Tests.Binding
         }
 
         [TestMethod]
+        public void BindingCompiler_ListIndexer_Get()
+        {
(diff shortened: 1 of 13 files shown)


acizmarik:
I just realized that it is broken also for arrays. Will leave this as a draft until resolved. 

Arrays are more complicated since assigning array elements is not compiled into method calls (but rather using ldelem / stelem instructions). This means I can not simply register JS translations for this.

exyi (approved):
:+1:

PR #1055, opened 2021-06-19
Always render everything in RouteLink

### LINKED ISSUE #781: RouteLink - Rendering based on RenderSettings.Mode
While investigating issue #779 I came across strange behavior of RoutLink control. If the ` RenderSettings.Mode` is set to `Server` it renders `a` tag with `href` attribute containing correctly formed url with query parameter values put in correct places.  Javascript logic needed to update the `href` attribute value on client side is also rendered:
```html
<a 
   href='/ControlSamples/RouteLink/RouteLinkQueryParameters?string=default&amp;int=5'
   onclick="event.stopPropagation();return !this.hasAttribute('disabled');" 
   class='link' 
   data-bind='attr: { &#39;href&#39;: &#39;/&#39; + "ControlSamples/RouteLink/RouteLinkQueryParameters" + dotvvm.buildUrlSuffix("", {"string": TestString,"int": TestInteger}) }, dotvvmEnable: true'>
   Test link
</a>
```

If the ` RenderSettings.Mode` is set to `Client` it renders the javascript logic, but there is no `href` attribute on the rendered `a` tag.
```html
<a 
   onclick="event.stopPropagation();return !this.hasAttribute('disabled');" 
   class='link' 
   data-bind='attr: { &#39;href&#39;: &#39;/&#39; + "ControlSamples/RouteLink/RouteLinkQueryParameters" + dotvvm.buildUrlSuffix("", {"string": TestString,"int": TestInteger}) }, dotvvmEnable: true'>
   Test link
</a>
```
If I understand philosophy of `RenderSettings.Mode` correctly it is so that `Client` means no server values are rendered in the original page source, just the javascript logic needed to update those values. Mode `Server` should render initial values from the viewmodel into the original page source as well as javascript logic for updating the values on the client side.

This is well and good in general but I find it strange to have `a` tag with no `href` attribute. Also not rendering links in the source is search engine unfriendly. And since `Client` mode is the default setting, this will result in poor discoverability of the web site and poor quality html for most of dotvvm users.

exyi:
Both the binding and server-side rendered value should be included in both
Server and Client Mode. Links are essential part of web and should be
server-side rendered whenever possible.

resolves  #781

### DIFF
--- a/src/DotVVM.Framework.Tests/ControlTests/testoutputs/SimpleControlTests.RouteLink.html
+++ b/src/DotVVM.Framework.Tests/ControlTests/testoutputs/SimpleControlTests.RouteLink.html
@@ -3,45 +3,45 @@
 	<body>
 		
 		<!-- client rendering, no params -->
-		<a href="/Simple" onclick="event.stopPropagation();return !this.hasAttribute('disabled');" data-bind="dotvvm-enable: true">Click me</a>
+		<a href="/Simple" onclick="event.stopPropagation();return !this.hasAttribute('disabled');">Click me</a>
 		
 		<!-- client rendering, no params, query and suffix -->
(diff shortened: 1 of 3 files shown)


acizmarik (commented):
LGTM. There just seem to be 2 relevant UI tests failing. Also we should probably hold with merging these changes or create a `v4-main` branch.

acizmarik (approved):

Issue #932, opened 2021-02-01
Custom Initializer Control Property


Mylan719:
I could use a feature that would let me register custom initializer for a control. My use case is as follows:

I would like to add a registered a resource for all my markup controls. I don't want to create code behind for every single markup control, just so that I could override `OnInit` and register my resource. I would like to use something like server side styles to register custom iaction to initialize a control for all controls that satisfy given filter condition.

This would no be hard to implement at all. Basicly we could have an internal properties like `OnInitAction`, `OnLoadAction`, `OnPreRenderAction`. And those would be run before or after `OnInit`, `OnLoad`, `OnPreRender`. I could then set those properties using Server Styles.

tomasherceg:
The concrete use case for requesting resources will be solved by #1056.

Issue #902, opened 2020-11-08
Add the ability to work with Control in server side style


martindybal:
The goal: Use server style to add PostbackHandler to all buttons which has data-ui attribute. Copy data-ui to postback handler property. 

Problem: there is no way How to get Control in SetControlProperty

Workaround: Use javascript to read data-ui from sender. 

```
.Register<DotVVM.BusinessPack.Controls.Button>()
.WithCondition(context => !string.IsNullOrWhiteSpace((context.Control.GetHtmlAttribute("data-ui") as ResolvedPropertyValue)?.Value as string))
.SetControlProperty<EventTrackPostBackHandler>(PostBack.HandlersProperty,
                    op => op.SetDotvvmProperty(EventTrackPostBackHandler.EventProperty, "VALUE OF DATA-UI"),
                    StyleOverrideOptions.Append);
```

exyi:
This is already implemented and merged into the v4-master.

Your case is written as follows:

```csharp
  config.Styles.Register<DotVVM.BusinessPack.Controls.Button>(c => c.HasHtmlAttribute("data-ui"))
       .AddPostbackHandler(c =>
              new EventTrackPostBackHandler { Event = c.GetHtmlAttribute("data-ui") }));
```

You can expect preview of v4 to be available reasonably soon.

Issue #895, opened 2020-10-28
Extensibility point for (not only) attached properties


tomasherceg:
Right now, we have `ActiveDotvvmProperty` that allows hooking up to `AddAttributesToRender` which is a great extensibility point - it lets anyone declare attached properties that can externally enhance any control or element. 

Unfortunately, `AddAttributesToRender` happens too late for some usages - for example, when the property needs to request an additional resource or adjust the control tree in some way.
I'd like to have some way of being able to hook a custom functionality earlier - maybe for `Init`, `Load`, and `PreRender` stages. 

Or do we have something like this? I know that a similar thing can be achieved using a decorator, but the advantage of the attached properties is that they can be applied using server-side styles.

exyi:
There is a few problems with adding Init, Load and PreRender to attached properties.

1. When the property is added later, do we have to invoke Init retrospectivelly? Probably yes, as we do for children, but that complicates (and slows down) working with properties significantly.
2. What about when the property value is changed, should we re-evaluate Init/Load/PreRender?
3. We optimize away Init, Load and PreRender invocations on controls that do not implement these method. It would complicate a lot of thing if any control could have a property that would have this method. Also, iterating all properties on all controls takes some time.

If you'd basically like to apply decorators using server-side styles, maybe it would be less problematic to add support for that. Alternativelly, you could add a child component that will modify its parent - it feels a lot like a hack, but it's not much different from the more powerful attached property you propose.

You are right, that requesting a resource from the property is quite a common use case... I don't really know how that could be solved. If it's only a script file, you can still request it in the render phase.

PR #1056, opened 2021-06-20
Refactor server-side styles to make them more type safe and powerful

exyi:
* Allows using lambas to query properties
* Allows setting properties based on existing properties
* Allows setting DotvvmControl instances directly (a ResolvedControl is created from that)
* Allows setting bindings in styles (in convenient way)

The API should be source compatible in common cases, but it's not at all binary compatible

Plus there is an extensibility point that should allow basically any modification of the control tree (the `StyleBuilder.AddApplicator` method)

Can resolve several open issues:

#902 can be now done as:

```csharp
  config.Styles.Register<Button>(c => c.HasHtmlAttribute("data-confirm-msg2"))
       .SetDotvvmProperty(PostBack.HandlersProperty, c =>
              new ConfirmPostBackHandler("Are you sure to do: " + c.Property(b => b.Text)));
```

or (less convenient, might be needed in some cases)

```csharp
config.Styles.Register<Button>(c => c.HasHtmlAttribute("data-confirm-msg"))
    .AppendControlProperty(PostBack.HandlersProperty, new ConfirmPostBackHandler(), s =>
        s.SetProperty(c => c.Message, c => c.Parent.GetHtmlAttribute("data-confirm-msg")));
```

https://github.com/riganti/dotvvm/issues/932

You can now register the resource (or a custom control that can do anything in OnInit) using Styles API:

```csharp
config.Styles.RegisterRoot()
    .AddCondition(...)
    .PrependContent(new RequiredResource("my-resource"));
```

In most cases,  #895 can also be emulated by this (an attached property which is converted to a control in children). Add support for decorating controls would be useful.


src/DotVVM.Framework/Compilation/Styles/StyleBuilderMethods.Helpers.cs
```diff
@@ -0,0 +1,23 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using DotVVM.Framework.Utils;
+using System.Reflection;
+using System.Linq;
+using DotVVM.Framework.Controls;
+using DotVVM.Framework.Compilation.Styles;
+
+public static partial class StyleBuilderExtensionMethods
```
tomasherceg, line 10: Is there a reason why these two methods are in a separate file?

tomasherceg (approved):

src/DotVVM.Framework/Compilation/Styles/StyleBuilderMethods.Helpers.cs
exyi: These methods are specific for some property or control, the other file is for general purpose functions. I think we will add more of them in this file.

PR #1057, opened 2021-06-22
Static command tests

fero02:


src/DotVVM.Framework/Resources/Scripts/utils/stringHelper.ts
```diff
@@ -6,3 +6,8 @@
 
     return tokens;
 }
+
+export function join<T>(elements: T[], delimiter: string): string {
+    let unwrappedElements = elements.map(function (element) { return ko.unwrap(element); });
```
exyi, line 11:
```suggestion
    let unwrappedElements = elements.map(ko.unwrap);
```

I'd prefer inlining it into the translation, but not that it would matter that much.

exyi (approved):

quigamdev (approved):

PR #1059, opened 2021-06-28
Added JavaScript translations for `DateTime` and `WebUtility`

### LINKED ISSUE #831: Suggestion for JS translations available in command expressions

acizmarik:
Resolves #831. This PR adds the following JS translations:

### `DateTime` property getters

* `Year`
* `Month`
* `Day`
* `Hour`
* `Minute`
* `Second`
* `Millisecond`

### `WebUtility` methods
Use these together with `@import System.Net`
* `UrlEncode(str)` mapped to `encodeURIComponent(str)`
* `UrlDecode(str)` mapped to `decodeURIComponent(str)`

### DIFF
--- a/src/DotVVM.Framework.Tests/Binding/BindingCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests/Binding/BindingCompilationTests.cs
@@ -912,6 +912,7 @@ namespace DotVVM.Framework.Tests.Binding
         public TestViewModel2 TestViewModel2B { get; set; }
         public TestEnum EnumProperty { get; set; }
         public string StringProp2 { get; set; }
+        public DateTime DateTime { get; set; }
         public DateTime? DateFrom { get; set; }
         public DateTime? DateTo { get; set; }
         public object Time { get; set; } = TimeSpan.FromSeconds(5);
--- a/src/DotVVM.Framework.Tests/Binding/JavascriptCompilationTests.cs
+++ b/src/DotVVM.Framework.Tests/Binding/JavascriptCompilationTests.cs
@@ -850,6 +850,20 @@ namespace DotVVM.Framework.Tests.Binding
         }
 
         [TestMethod]
+        [DataRow("DateTime.Year", "getFullYear", false)]
+        [DataRow("DateTime.Month", "getMonth", true)]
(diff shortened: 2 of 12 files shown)


exyi (approved):
:+1:

PR #1062, opened 2021-07-09
Added JS translations for string method overloads with `StringComparison`

acizmarik:
This PR adds another batch of JS translations for string methods:

- `string.Contains(string value, StringComparison options)`
- `string.EndsWith(string value, StringComparison options)`
- `string.IndexOf(string value, StringComparison options)`
- `string.IndexOf(string value, int startIndex, StringComparison options)`
- `string.LastIndexOf(string value, StringComparison options)`
- `string.LastIndexOf(string value, int startIndex, StringComparison options)`
- `string.StartsWith(string value, StringComparison options)`

The only truly recognized values from `StringComparison` enums are `InvariantCulture` and `InvariantCultureIgnoreCase` with the `InvariantCulture` being the default. Once this feature gets merged I will update relevant documentation pages accordingly.


exyi (approved):
I was thinking that we could implement this purely in the translator by looking at the constant and wrapping both arguments with `toLower`. This seems easier, though :)

PR #1067, opened 2021-07-15
Configure Azure Pipelines (YAML)

cafour:
Adds and modifies scripts, configs, csprojs, and other files so that our C# unit tests, Jest tests, and UI tests can run in Azure Pipelines.

(This is the complete implementation and thus closes #1046.)

### DIFF
--- a/ci/linux/test.ps1
+++ b/ci/linux/test.ps1
@@ -1,0 +1,1 @@
+docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
--- a/ci/linux/test.ps1
+++ b/ci/linux/test.ps1
@@ -1,1 +1,0 @@
-docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
--- a/ci/linux/test.sh
+++ b/ci/linux/test.sh
@@ -1,0 +1,1 @@
+docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
--- a/ci/linux/test.sh
+++ b/ci/linux/test.sh
@@ -1,1 +1,0 @@
-docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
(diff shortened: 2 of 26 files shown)


ci/linux/test.ps1
```diff
@@ -0,0 +1 @@
+docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
```
MichalTichy, line 1: This accesses internal docker registry.

ci/linux/test.sh
```diff
@@ -0,0 +1 @@
+docker run --rm -it --mount "type=bind,src=${PWD},dst=/s" docker-reg.riganti.cz/docker-agent:latest-linux bash
```
MichalTichy, line 1: This accesses internal docker registry.

MichalTichy: Not sure if we event need this script.

ci/linux/test.ps1
MichalTichy: Not sure if we event need this script.

ci/windows/test.ps1
```diff
@@ -0,0 +1 @@
+docker run --rm -it --mount "type=bind,src=${PWD},dst=C:\s" docker-reg.riganti.cz/docker-agent:latest-win powershell
```
MichalTichy, line 1: This accesses internal docker registry.

src/DotVVM.Framework/package-lock.json
```diff
@@ -0,0 +1,6429 @@
+{
```
MichalTichy, line 1: I dont think that this file should be part of repository

src/DotVVM.Samples.Tests/Feature/ApiTests.cs
```diff
@@ -202,7 +202,7 @@ public void Feature_Api_AzureFunctionsApiTable()
             });
         }
 
-        [Fact]
+        [Fact(Skip = "Doesn't work on OWIN because it relies on _apiCore.")]
```
MichalTichy, line 205: In the future it would be nice to have SkipOnOwin / SkipOnAspNetCore attributes.

MichalTichy (changes requested):

src/DotVVM.Framework/package-lock.json
cafour: Why? It guarantees that everyone uses the same dependencies, which seems especially desirable in a CI setting.
(See https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json.)

src/DotVVM.Samples.Tests/Feature/ApiTests.cs
cafour: Agreed. Please create a separate issue.

quigamdev (approved):

PR #1068, opened 2021-07-16
Fixed issues with string interpolation

acizmarik:
This PR fixes a few issues that were found during testing string interpolation feature.

1. `dotvvm.globalization.format` function was incorrectly called. This could have caused rendering of `undefined value`s when user specified multiple interpolated expressions
2. `dotvvm.globalization.format` incorrectly interpretted formatting options. This caused that many formatting options were unrecognized by the globalization library
3. `BindingTokenizer` could have read one character past the string boundary. This was only observable when a string ends with `{{` or `}}` sequence. Rendered strings had additional character (always the ending quotation mark).
4. `BindingParser` was not skipping whitespaces after parsing interpolated string. This caused parsing exceptions in some cases.

Together with @fero02 we also found some additional minor issues related to the `dotvvm.globalization` library that will be addressed in a separate PR.


exyi (approved):
:+1:

Issue #1064, opened 2021-07-11
Validator.InvalidCssClass and other properties not working when applied on GridViewColumn


tomasherceg:
Since the `GridViewColumn` itself is not a part of the control tree, the attached properties assigned to it are not inherited.
We should either "copy" them to the `td` element, or provide another way to apply these properties on the contents of the column (maybe add the `CellDecorators` property that will be similar to `RowDecorators`).

mrnustik:
👍 for `CellDecorators`, this would also add the possibility to for example change the colour of a cell based on its value.

tomasherceg:
Related to #153

exyi:
This is already done, right?

PR #1077, opened 2021-07-23
Cell decorators in GridViewColumn

tomasherceg:
This PR adds support for `CellDecorators` property requested in #1064. 
There will be a separate PR with support for `Validator` properties in concrete columns (`GridViewTextColumn` and others).


src/DotVVM.Framework/Controls/GridView.cs
```diff
@@ -94,6 +94,21 @@ public GridViewFilterPlacement FilterPlacement
         public static readonly DotvvmProperty RowDecoratorsProperty =
             DotvvmProperty.Register<List<Decorator>?, GridView>(c => c.RowDecorators);
 
+        /// <summary>
+        /// Gets or sets a list of decorators that will be applied on the header row.
+        /// </summary>
+        [MarkupOptions(AllowBinding = false, MappingMode = MappingMode.InnerElement)]
+        [ControlPropertyBindingDataContextChange("DataSource")]
+        [CollectionElementDataContextChange(1)]
```
exyi, line 102: Header rows don't have this data context.

src/DotVVM.Framework/Controls/GridViewColumn.cs
```diff
@@ -127,6 +128,47 @@ public bool Visible
         public static readonly DotvvmProperty VisibleProperty
             = DotvvmProperty.Register<bool, GridViewColumn>(c => c.Visible, true);
 
+        /// <summary>
+        /// Gets or sets a list of decorators that will be applied on each cell which is not in the edit mode.
+        /// </summary>
+        [PopDataContextManipulation]
```
exyi, line 134: Cell decorators shouldn't pop, they are in the row context.

exyi (commented):
Would you please add test for bindings in the CellDecorators? I'd like to be sure that we are right with the data context types. You can also use this unit testing approach, so we don't rely on UI tests for things that have nothing to do with JS: https://github.com/riganti/dotvvm/blob/main/src/DotVVM.Framework.Tests/ControlTests/SimpleControlTests.cs#L110

src/DotVVM.Framework/Controls/GridViewColumn.cs
tomasherceg: Good catch, thanks!

src/DotVVM.Framework/Controls/GridView.cs
exyi: I think this is still a problem and it's not covered by the test

exyi (approved):
:+1:

GrabzIt:
Great framework by the way. However is this feature working correctly?

When I use: 

            <dot:GridViewTemplateColumn>
                <CellDecorators>
                    <dot:Decorator class="col-1" />
                </CellDecorators>
                <ContentTemplate>
                    {{value: HTML}}
                </ContentTemplate>
            </dot:GridViewTemplateColumn>

I get:

<td>
    <celldecorators>
        
    </celldecorators>
    <contenttemplate class="col-1">
        Everything is up and running.
    </contenttemplate>
</td>

I am using version 3.2.1. It could be that I am misunderstanding how this should work I was expecting something like.

<td class="col-1">
        Everything is up and running.
</td>

quigamdev:
Hello @GrabzIt, this feature is in DotVVM 4.0. We have released a preview version of it. So you can check out the NuGet.

GrabzIt:
Thanks, that explains it!

PR #1078, opened 2021-07-24
Prefer to use .state in staticCommands

exyi:
This should fix the annoying synchronization issues when state is updated
in user code and staticCommand is invoked, which reads the old data :/

### DIFF
--- a/src/DotVVM.Framework.Hosting.AspNetCore/DotVVM.Framework.Hosting.AspNetCore.csproj
+++ b/src/DotVVM.Framework.Hosting.AspNetCore/DotVVM.Framework.Hosting.AspNetCore.csproj
@@ -49,6 +49,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Localization" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Localization" Version="2.1.0" />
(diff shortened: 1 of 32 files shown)


src/DotVVM.Framework/Compilation/Javascript/ParametrizedCode.cs
```diff
@@ -229,7 +244,7 @@ public void Add(CodeParameterInfo parameter)
                 parameters.Add(parameter);
             }
 
-            public void Add(ParametrizedCode code, byte operatorPrecedence = 20)
```
acizmarik, line 232: Can you check that this change is correct?

src/DotVVM.Framework.Hosting.AspNetCore/DotVVM.Framework.Hosting.AspNetCore.csproj
```diff
@@ -49,6 +49,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Localization" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Localization" Version="2.1.0" />
+    <PackageReference Include="Microsoft.NET.Sdk.Razor" Version="3.1.12" />
```
acizmarik, line 52: This should not be here

quigamdev (approved):

PR #1081, opened 2021-07-25
Added control usage validation to GridViewColumn

tomasherceg:
According to #1064, I've added validations for `GridView` columns so the user is not able to specify `DataContext` and any attached properties. 
Originally, I thought about allowing the concrete column types to indicate that they support specific attached properties, but then I noticed that the columns already work with `Validator.Value` (they can either apply the validator on the control itself, or add a stand-alone validation control), and the other validator properties can be set using `CellDecorators` (provided that every table column needs different settings which is not a frequent use case).


src/DotVVM.Framework/Controls/GridViewColumn.cs
```diff
@@ -205,6 +210,28 @@ private void SetSortedCssClass(HtmlGenericControl cell, ISortableGridViewDataSet
...
+            if (control.Properties.ContainsKey(DataContextProperty))
+            {
+                yield return new ControlUsageError("Changing the DataContext property on the GridViewColumn is not supported!", control.DothtmlNode);
+            }
+
+            // disallow attached properties on columns
+            foreach (var property in control.Properties)
+            {
+                // ignore attached properties that are set by runtime and not from markup
+                if (property.Value.DothtmlNode == null) continue;
```
exyi, line 226: This will also allow setting these properties by server-side styles, not sure we want that.

exyi: It will also disallow the Validator.Value that we support, right?

exyi (commented):
Could you please add unit test for this, we don't need to rely on UI tests to test the compiler. You can use this approach: https://github.com/riganti/dotvvm/blob/main/src/DotVVM.Framework.Tests/ControlTests/SimpleControlTests.cs#L110

Also, we should add test for the Validator.Value property if we support that.

src/DotVVM.Framework/Controls/GridViewColumn.cs
tomasherceg: @exyi `Validator.Value` is set automatically to the same expression as the `ValueBinding`, so there is no point in setting `Validator.Value` on the column. If someone wants to do it, we'll have `CellDecorators` for that.

src/DotVVM.Framework/Controls/Internal.cs
```diff
@@ -60,6 +61,11 @@ public class Internal
 
         public static DotvvmProperty ReferencedViewModuleInfoProperty =
             DotvvmProperty.Register<ViewModuleReferenceInfo, Internal>(() => ReferencedViewModuleInfoProperty, isValueInherited: true);
+
+        public static bool IsViewCompilerProperty(DotvvmProperty property)
+        {
+            return property == DataContextTypeProperty;
```
exyi, line 67:
```suggestion
            return property.DeclaringType == typeof(Internal);
```

There are other properties (like unique id, markup line, ..) which might be there.

exyi (approved):

PR #1082, opened 2021-07-26
fix OnPreInit invocation on dotvvm markup control

exyi:
This caused weird bugs when this constructor was used and
the markup control contained @js directive, since the resource was not loaded


acizmarik (approved):
👍

PR #1084, opened 2021-07-26
Add DotVVM.Framework.Testing project

exyi:
The project is meant to contain helper methods for testing dotvvm controls


src/Tools/CI/restore.cmd
```diff
@@ -21,6 +21,9 @@ dotnet restore
 cd ../DotVVM.Framework.PerfTests
 dotnet restore
 
+cd ../DotVVM.Framework.Testing
```
quigamdev, line 24: This file actually is not needed. We should remove it.

src/Tools/CI/release.cmd
```diff
@@ -11,6 +11,9 @@ dotnet build ../../DotVVM.Framework -c Release
 dotnet restore ../../DotVVM.Framework.Hosting.AspNetCore
 dotnet build ../../DotVVM.Framework.Hosting.AspNetCore -c Release
 
+dotnet restore ../../DotVVM.Framework.Testing
+dotnet build ../../DotVVM.Framework.Testing -c Release
```
quigamdev, line 15: This file actually is not needed. We should remove it.

src/Tools/CI/build.cmd
```diff
@@ -11,6 +11,9 @@ dotnet build ../../DotVVM.Framework
 dotnet restore ../../DotVVM.Framework.Hosting.AspNetCore
 dotnet build ../../DotVVM.Framework.Hosting.AspNetCore
 
+dotnet restore ../../DotVVM.Framework.Testing
```
quigamdev, line 14: This file actually is not needed. We should remove it.

quigamdev (approved):

src/Tools/CI/build.cmd
exyi: OK, won't do that in this PR though. Could you please do that when you have a bit of time?

PR #1087, opened 2021-07-29
Load even assemblies from DotvvmMarkupConfiguration.Controls

exyi:
The reason is that some users and components (like DotVVM.Contrib)
register the controls using c.Markup.Controls.Add. This
does not add the assembly to the Markup.Assemblies list,
which in turn causes issues with explicit assembly loading.

These issues are super mysterious and hard to debug, so we should release this in v3


quigamdev (approved):

quigamdev:
I had to add a condition to filter nulls and whitespace characters. Otherwise, unit tests fail on assembly loading that throws InvalidArgumentException, because one item is null.

PR #1089, opened 2021-07-31
Styles: Append, Prepend, Wrap, Replace, Tags

exyi:
Added capability of changing the the control tree using server side styles. Added these properties:

```csharp
        /// <summary> List of controls that will placed as wrappers to this component (the last will be top-most) </summary> 
        [MarkupOptions(MappingMode = MappingMode.InnerElement)]
        public static CompileTimeOnlyDotvvmProperty WrappersProperty =
            CompileTimeOnlyDotvvmProperty.Register<IEnumerable<DotvvmBindableObject>, Styles>("Wrappers");
        /// <summary> List of controls that will be placed after this control (if possible) </summary> 
        [MarkupOptions(MappingMode = MappingMode.InnerElement)]
        public static CompileTimeOnlyDotvvmProperty AppendProperty =
            CompileTimeOnlyDotvvmProperty.Register<IEnumerable<DotvvmBindableObject>, Styles>("Append");
        /// <summary> List of controls that will be placed before this control in reverse order (if possible) </summary> 
        [MarkupOptions(MappingMode = MappingMode.InnerElement)]
        public static CompileTimeOnlyDotvvmProperty PrependProperty =
            CompileTimeOnlyDotvvmProperty.Register<IEnumerable<DotvvmBindableObject>, Styles>("Prepend");
        /// <summary> A control which will be used instead of the specified control. Properties will be translated to the properties of the new control when possible or added as attached properties when not directly translatable. </summary> 
        [MarkupOptions(MappingMode = MappingMode.InnerElement)]
        public static CompileTimeOnlyDotvvmProperty ReplaceWithProperty =
            CompileTimeOnlyDotvvmProperty.Register<DotvvmBindableObject, Styles>("ReplaceWith");
        /// <summary> No Server Side Styles will be applied to this control. When set by a style, this will be the last style to be applied to this control (although when the style contains multiple applicators, all of them will be applied) </summary> 
        public static CompileTimeOnlyDotvvmProperty ExcludeProperty =
            CompileTimeOnlyDotvvmProperty.Register<bool, Styles>("Exclude");
        /// <summary> A tag which can be used to selectively style certain controls. </summary> 
        public static CompileTimeOnlyDotvvmProperty TagProperty =
            CompileTimeOnlyDotvvmProperty.Register<string[], Styles>("Tag");

        /// <summary> Append to this property to add a required resource somewhere to the page. </summary> 
        [MarkupOptions(MappingMode = MappingMode.Exclude)]
        public static CompileTimeOnlyDotvvmProperty RequiredResourcesProperty =
            CompileTimeOnlyDotvvmProperty.Register<string[], Styles>("RequiredResources");
```

`Styles.Tag` can be used like compile-time css class. `Style.Exclude` is for special circumstances when we don't want to apply styles to a specific control (note that it does apply for styles defined directly in the control, only for styles defined in DotvvmConfiguration)

`Styles.Append` appends compoents after the control where this property is applied. `Styles.Prepend` is the opposite. `Styles.Wrappers` are controls which are put above the control in the tree. Note that wrappers are applied after append and prepend.

`Styles.Replace` is a bit special - it completely removes the original control and replaces it with a different one. All properties are translated to the new control when possible. Otherwise the properties are left there as attached properties.


src/DotVVM.Framework/Compilation/Styles/StyleBuilderMethods.Helpers.cs
```diff
@@ -9,15 +9,117 @@
...
-        sb.SetControlProperty(PostBack.ConcurrencyQueueSettingsProperty, handler);
+        sb.SetControlProperty(PostBack.ConcurrencyQueueSettingsProperty, handler, options: StyleOverrideOptions.Append);
 
+    /// <summary> Adds a new postback handler to the PostBack.Handlers property </summary>
     public static IStyleBuilder<T> AddPostbackHandler<T>(
         this IStyleBuilder<T> sb,
         Func<IStyleMatchContext<T>, PostBackHandler> handler)
         where T: DotvvmBindableObject =>
-        sb.SetDotvvmProperty(PostBack.ConcurrencyQueueSettingsProperty, handler);
+        sb.SetDotvvmProperty(PostBack.ConcurrencyQueueSettingsProperty, handler, StyleOverrideOptions.Append);
```
exyi, line 24: This is wrong, let's add a test for it.

quigamdev (approved):
I have had a call with @exyi and we have gone through the review. So all my comments are already resolved. This PR is ready for merge.

src/DotVVM.Framework/Compilation/ViewCompilingVisitor.cs
```diff
@@ -245,8 +245,8 @@ protected string CreateControl(ResolvedControl control)
                 name = emitter.EmitInvokeControlBuilder(control.Metadata.Type, control.Metadata.VirtualPath);
             }
             emitter.RegisterDotvvmProperties(name);
-            // RawLiterals don't need these helper properties
-            if (control.Metadata.Type != typeof(RawLiteral))
+            // RawLiterals don't need these helper properties unless in root
+            if (control.Metadata.Type != typeof(RawLiteral) || control.Parent is ResolvedTreeRoot)
```
quigamdev, line 249: Nice 👍

quigamdev:
Tests are green. I think we can merge it.

PR #1095, opened 2021-08-03
Enable access to parent extension parameters

### LINKED ISSUE #392: Allow Access to Parent Extension Parameters
Currently, it is not possible to access `_index` in two nested repeaters like this:

```html
<dot:Repeater DataSource="{value: Collection1}">
   <dot:Repeater DataSource="{value: Collection2}">
           index in the second collection is {{value: _index}}
           index in the first collection is {{value: how can I access the parent's _index}}
    </dot:Repeater>
</dot:Repeater>
```

There is several possibilities how to handle it:
* `_parent_index` identifier - easy to implement, but I don't like the syntax and it will create a lot of (mostly redundant) identifiers, that will spam IntelliSense
* `_parent._index` - virtual property on the `_parent` identifier, it is not hard to implement in the framework, but I'm not that sure about VS extension where the expressions are translated to Roslyn syntax tree which of course does not support that. cc @Mylan719

exyi:
The syntax is _parent._index, _root._index and so on. _this._index can also
be used, athough it's of little practical use. Only difference between
_index and _this._index is that the _this will disallow access to
parameters which are inherited from above. For example, _this._page will only
work in the root data context while _page works essentialy everywhere.

resolves #392

### DIFF
--- a/src/DotVVM.Framework.Testing/BindingTestHelper.cs
+++ b/src/DotVVM.Framework.Testing/BindingTestHelper.cs
@@ -60,22 +60,25 @@ namespace DotVVM.Framework.Testing
                 TypeConversion.ImplicitConversion(parsedExpression, expectedType, true, true);
         }
 
-        public ParametrizedCode ValueBindingToParametrizedCode(string expression, Type[] contexts, Type? expectedType = null, NamespaceImport[]? imports = null, bool nullChecks = false, bool niceMode = true)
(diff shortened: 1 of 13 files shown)


quigamdev (changes requested):
I have added sample ControlSamples/Repeater/IndexInNestedRepeater that represents a bug in server rendering that I have found. 

Client rendering and server rendering should produce the same result, but server rendering does seem not to respect context or something like that.

quigamdev (approved):

PR #1098, opened 2021-08-09
Fixed REST API bindings generation

### LINKED ISSUE #889: Default Newtonsoft.Json settings in generated REST API Bindings

### LINKED ISSUE #911: Dotvvm Swashbuckle 2.5.0 does not generate responses in client
I am using REST Api bindings for the first time (.NET Core 3.1 + all the latest Dotvvm packages). When I generate client by running dotvvm console commands "dotnet dotvvm api ...", all public methods of created client return void (or Task). Some of them should return response object.

API example:
```
        [HttpGet]
        public IEnumerable<Airport> Airports()
        {
            return SampleData.Airports;
        }
```
Should generate:
```
       [DotVVM.Framework.Configuration.RestApiRegistrationHelpers.HttpMethod("Get")]
        public System.Collections.ObjectModel.ObservableCollection<FlightFinder.Shared.Airport> Airports()
        {
            return System.Threading.Tasks.Task.Run(async () => await AirportsAsync(System.Threading.CancellationToken.None)).GetAwaiter().GetResult();
        }
```
But generates:
```
        [DotVVM.Framework.Configuration.RestApiRegistrationHelpers.HttpMethod("Get")]
        public void GetAirports()
        {
            System.Threading.Tasks.Task.Run(async () => await GetAirportsAsync(System.Threading.CancellationToken.None)).GetAwaiter().GetResult();
        }
```

Swagger JSON looks OK and I see response objects there. NSwag tool also generates correct clients.

If I use DotVVM.Api.Swashbuckle.AspNetCore 2.4.0.1 everything is fine. I made simple demo. I targeted .NET Core 2.0, because DotVVM.Api.Swashbuckle.AspNetCore 2.4.0.1 has troubles with .NET Core 3+ (see #842).

Demo - Not working 2.5.0 - https://github.com/mirecad/dotvvm-samples-flight-finder (latest master)
Demo - Working 2.4.0.1 - https://github.com/mirecad/dotvvm-samples-flight-finder/commit/fb1f252ef0e7fe858e08265ce3845271da831ab7

Steps to reproduce:
- Run FlightFinder.Api project
- run command `dotnet dotvvm api regen` in src/FlightFinder.Web folder
- check generated ApiClient.cs

acizmarik:
Resolves #911. Generated C# clients should now contain correct code. This also resolves #889 as all generated bindings now seem to create own instance of `JsonSerializerSettings` that gets passed around.


acizmarik:
I was not able to fix the github api sample as the original swagger.json seems to be lost... However, other samples should be working right now. We should decide whether to keep the github sample or remove it.

exyi (approved):
So, I don't really understand what was the problem, but this won't affect anything else in the framework and API binding can't be broken even more :sweat_smile:

PR #1100, opened 2021-08-10
Added more JS translations for string methods

acizmarik:
This PR introduces support for the following .NET methods:

- `String.IsNullOrWhiteSpace(string value)`
- `string.PadLeft(int length)`
- `string.PadLeft(int length, char c)`
- `string.PadRight(int length)`
- `string.PadRight(int length, char c)`
- `string.Trim()`
- `string.Trim(char c)`
- `string.TrimStart()`
- `string.TrimStart(char c)`
- `string.TrimEnd()`
- `string.TrimEnd(char c)`


src/DotVVM.Framework/Resources/Scripts/utils/stringHelper.ts
```diff
@@ -49,3 +49,43 @@ export function join<T>(elements: T[], delimiter: string): string {
...
+        char = char.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    }
+    else {
+        char = "\\s";
+    }
+    return text.replace(new RegExp(char + "*$"), "");
+}
+
+export function padStart(text: string, length: number, char: string) {
+    if (length <= text.length) {
```
exyi, line 74: I think you can use this function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart. We don't support IE anyway

quigamdev: We cannot drop IE support at main. This will be possible for v4.

quigamdev (approved):
I have fixed a small issue with registration. We should merge this into main and after that, we need to remove IE support to reduce boilerplate code.

PR #1101, opened 2021-08-11
Changed default value for Repeater.RenderAsNamedTemplate

quigamdev:
A repeater should render templates by default as named templates. It causes js issues in controls like business-pack TreeView etc..


exyi:
> It causes js issues in controls like business-pack TreeView

That means that the TreeView is buggy, it should work even in the inline-template.

quigamdev:
I thought that hierarchical controls have to use the script template. Still, this seems to me like a better default behaviour. In case the application of ViewModel is slow browser renders invalid HTML for a split second. Unfortunately, this incorrectly rendered HTML is noticeable. Rendering of a repeater as a named template does not solve this issue entirely but it behaves better than the current default rendering.

exyi (approved):
:+1:

PR #1102, opened 2021-08-11
Upgrade framework targets 

quigamdev:


src/DotVVM.Core/DotVVM.Core.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <VersionPrefix>4.0.0-preview01-116912</VersionPrefix>
-    <TargetFrameworks>netstandard1.6;net451</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
```
exyi, line 4: net472 should be compatible with netstandard2.0, right? So could we get rid of explicit target for net472?

quigamdev: I tried to remove net472 but unfortunately, there is some problem with the build. I'm not completely sure whether it is possible to simply remove the net4xx completely.

quigamdev: Ok, I have fixed the issue with the build but still, there is a problem with loading/referencing assemblies. It seems like we have to leave there the net472 and most probably we have to discourage people from using [net461-net471].

src/DotVVM.Compiler/DotVVM.Compiler.csproj
```diff
@@ -10,7 +10,7 @@ for some reason. What fun. -->
 
   <!-- Related to compilation -->
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net461;net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net472;net461;net5.0</TargetFrameworks>
```
exyi, line 13: Did you forgot to remove net461? If not, you'll have to include it in the condition bellow

src/DotVVM.Core/DotVVM.Core.csproj
exyi: The problem is still that you'll be able to install dotvvm on net461 since it support netstandard2.0...

quigamdev:
@exyi What if we just remove netstandard2.0 and leave there just net472 and netstandard2.1 ?

exyi:
Leaving only net472 and netstandard2.1 seems quite reasonable, IMHO.

quigamdev:
@cafour Thanks for incorporating the last changes (net472 and netstandard2.1)

tomasherceg (approved):

PR #1106, opened 2021-08-17
Fix redirection in staticCommand & SPA

exyi:
Fixes problems with redirecting staticCommand in SPA
- the redirect worked, but since no exception was thrown, rest of the command continued to run, which caused problems on the new page
- now it throws an exception, which cancels rest of the binding
- also, the exception may be similarly thrown by a user function if it would want to make a redirect


quigamdev (approved):
Seems ok to me

PR #1107, opened 2021-08-17
Do not UpdateSource automatically in SetValue

exyi:
Instead, SetValueToSource was added, if the user really wants to update the
value in viewModel (which might be useful in control commands, for example)

Please have a look, but don't merge, we need to discuss it first.


src/DotVVM.Framework/Controls/DotvvmBindableObject.cs
```diff
@@ -132,17 +132,16 @@ public void MagicSetValue(DotvvmProperty[] keys, object[] values, int hashSeed)
...
+            SetValueRaw(property, value);
+        }
+
+        /// <summary> Sets the value of specified property by updating the view model this property is bound to. Throws if the property does not contain binding </summary>
+        public void SetValueToSource(DotvvmProperty property, object? value)
+        {
+            var binding = GetBinding(property);
+            if (binding is null)
+                throw new DotvvmControlException(this, $"Property {property} does not contain binding, so it's source can not be updated.");
+            ((IUpdatableValueBinding)binding).UpdateSource(value, this);
```
quigamdev, line 144: Does the GetBinding always return IUpdatableValueBinding? Otherwise, this can be a source of InvalidCastException

PR #1109, opened 2021-08-21
TemplateHost control

tomasherceg:
When working on #535, we found that it will be useful to have a `TemplateHost` control. 

I've put it in a separate PR and added some UI tests - the control can be used to pass templates in markup controls.


src/DotVVM.Framework/Controls/TemplateHost.cs
```diff
@@ -0,0 +1,38 @@
...
+            set { SetValue(TemplateProperty, value); }
+        }
+        public static readonly DotvvmProperty TemplateProperty
+            = DotvvmProperty.Register<ITemplate, TemplateHost>(c => c.Template, null);
+
+
+
+        protected internal override void OnLoad(IDotvvmRequestContext context)
+        {
+            Template.BuildContent(context, this);
```
exyi, line 34: Let's check here that data context of the template is compatible with the data context of this control. Otherwise, it will run into strange errors in the nearest binding...

src/DotVVM.Framework/Controls/TemplateHost.cs
```diff
@@ -18,13 +20,38 @@ public class TemplateHost : DotvvmControl
...
+        public static readonly DotvvmProperty TemplateProperty
+            = DotvvmProperty.Register<ITemplate, TemplateHost>(c => c.Template, null);
+
 
 
         protected internal override void OnLoad(IDotvvmRequestContext context)
         {
-            ContentTemplate?.BuildContent(context, this);
+            var placeHolder = new PlaceHolder();
+            Template.BuildContent(context, placeHolder);
```
exyi, line 36: Nullable types should complain somewhere here, please add a simple nullcheck to avoid NRE when users forgets to set the property.
```suggestion
            Template.NotNull("TemplateHost.Template is required").BuildContent(context, placeHolder);
```


Also, I think it would be convenient to add a constructor overload with template as parameter. Maybe we could also add overload to the AppendChildren(ITemplate) helper method that would automatically create this control.

src/DotVVM.Samples.Common/Views/ControlSamples/TemplateHost/CompositeControlWithTemplate.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public static DotvvmControl GetContents(
+            ValueOrBinding<string> headerText,
+            ITemplate contentTemplate
+        )
+        {
+            return new HtmlGenericControl("fieldset")
+                .AppendChildren(
+                    new HtmlGenericControl("legend", new TextOrContentCapability() { Text = headerText }),
+                    new Framework.Controls.TemplateHost() { Template = contentTemplate }
```
exyi, line 19: Is there some naming conflict?

```suggestion
                    new TemplateHost() { Template = contentTemplate }
```

exyi (approved):

src/DotVVM.Samples.Common/Views/ControlSamples/TemplateHost/CompositeControlWithTemplate.cs
tomasherceg: Yes, it's correct. 
The folders in control samples have the same names as the controls being tested, so the class is in the `TemplateHost` namespace. :-(

PR #1110, opened 2021-08-22
Add a MSBuild task for running `yarn build`

exyi:
In case yarn is not installed, npm is used.
Yarn is preferred, because it takes less time to run it, so it's
less annoying.
The Target only runs when any typescript file is changed,
otherwise it's skipped by MSBuild

In case there are some JS error, the build should continue normally,
so hopefully it won't cause some build issues

This should solve the problem that we forget to build the JS and
then something does not work. And when someone clones the repo,
it just won't compile because the JS files are missing


cafour:
This seems to work alright. Would you mind removing the `npm build` step from `ci/linux/run.sh` and `ci/windows/run.ps` or should I do it?

quigamdev (approved):

PR #1111, opened 2021-08-24
Remove support of virtual properties

### LINKED ISSUE #802: What to do with virtual dotvvm properties?
We have a concept of virtual dotvvm properties that create `DotvvmProperty` in the background from a standard C# property when it is annotated with `MarkupOptionsAttribute`. It has some obvious limitations (the property can not contain bindings, can not be inherited) and some less obvious limitations (can not contain even `resource` binding, can not be set from styles).

Basically, the same applies to property groups, which have two modes - is set directly to a property (for example [`HtmlGenericControl`](https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Controls/HtmlGenericControl.cs#L56)) or it's stored in `control.properties` and accessed through `VirtualPropertyGroupDictionary` (for example [`RouteLink.Params-...`](https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Controls/RouteLink.cs)) (in this case, virtual word is very overloaded and it means exactly the opposite to the paragraph above :weary:)

Since this feature showed itself as not very useful (AFAIK it's actually not used), we could maybe drop it for the sake of consistency in the following major release.

Or, we could at least fix the inconsistencies - allow setting these properties though compile-time style. We could also allow them to hold bindings by a `ValueOrBinding<T>` type (see #801). The advantage of these properties is that they are declared with a lot less boilerplate, and maybe this could make writing DotVVM controls less scary.

exyi:
resolves #802

They can be removed since capabilities now do a better job without
all the glitches.

Main reason for removal is that it leads to very unintuitive behavior when
the property is not registered for some reason (usually explicit assembly loading)

It also created redundant edge cases on some places and it's not really used in practise.
The error message hopefully explains what is going on.

I have left support for virtual properties in property groups, since we still use this in HtmlGenericControl.Attributes.
We'll discuss that and maybe also remove that one.


src/DotVVM.Framework/Binding/DotvvmProperty.cs
```diff
@@ -238,29 +238,24 @@ public static void InitializeProperty(DotvvmProperty property, ICustomAttributeP
...
+                select p).ToArray();
+
+            if (properties.Any())
+            {
+                var controlHasOtherProperties = registeredProperties.Values.Any(p => p.DeclaringType == controlType);
+                var explicitLoadingHelp =
+                    !controlHasOtherProperties ? $" The control does not have any other properties registered, it could indicate that DotVVM did not register properties from this assembly - it is common with ExplicitAssemblyLoading enabled, try to register {controlType.Assembly} into config.Markup.Assemblies." : "";
+                var deprecationHelp = " DotVVM version <= 3.x did support this, but this feature was removed as it lead to many issues. Please register the property using DotvvmProperty.Register. If you find this annoyingly verbose, you could use control capabilities instead (using DotvvmCapabilityProperty.Register).";
+                throw new NotSupportedException($"Control {controlType} has properties which are not registered as DotvvmProperty but have a MarkupOptionsAttribute: {string.Join(", ", properties.Select(p => p.Name))}." + explicitLoadingHelp + deprecationHelp);
```
cafour, line 257: Since this exception message is rather long, I have a suggestions to improve its typography:

* Use quotation marks around arguments to be consistent with most of our other exception messages (`'{controlType}'`).
* Either put a comma before "which", or use "that".
* "a DotvvmProperty".
* "...any other properties registered. It could indicate..."
* I think we can omit "- it is common with ExplicitAssemblyLoading enabled" for brevity.
* "DotVVM version <= 3.x" -> "DotVVM versions prior to 3.0"
* "...using DotvvmProperty.Register or use control capabilities instead."

cafour (changes requested):

src/DotVVM.Framework/Binding/DotvvmProperty.cs
exyi: "DotVVM versions prior to 3.0" - that would be <

> I think we can omit "- it is common with ExplicitAssemblyLoading enabled" for brevity.

I think it's useful information, this error is quite unlikely to occur without this option and adding to Assemblies will not help if ExplicitAssemblyLoading is disabled.

PR #1114, opened 2021-08-24
Use template element instead of script tag

exyi:
This should make it more reliable, since templates can be nested


quigamdev (approved):
Nice 👍

Issue #1075, opened 2021-07-20
Add Status Page with DotvvmConfiguration


exyi:
To simplify configuration debugging, we should add a page that will dump the contents of DotVVM Configuration in a readable way. Most importantly, it should contain the items which are usually autoconfigured (like registered controls, routes, resources, ...).

With improvements like #535, it would be also helpful to list DotvvmProperties and capabilities of each control.

quigamdev:
This can be part of the diagnostics status page (https://github.com/riganti/dotvvm-samples-compilation-status-page)

quigamdev:
It looks like this issue can be closed. 
#1119 added a status page to the framework.

PR #1119, opened 2021-08-26
Add status page into Framework

cafour:
As part of #1075, I merged the riganti/dotvvm-samples-compilation-status-page repo into this one, integrated it into Framework, and simplified it a bit.


src/DotVVM.Framework.Hosting.Owin/AppBuilderExtensions.cs
```diff
@@ -125,6 +124,7 @@ private static DotvvmConfiguration UseDotVVM(this IAppBuilder app, string applic
             startupTracer.TraceEvent(StartupTracingConstants.DotvvmConfigurationUserConfigureFinished);
 
             modifyConfiguration?.Invoke(config);
+            config.Development.Apply(config);
```
quigamdev, line 127: Please change Development to Diagnostics

src/DotVVM.Framework/Configuration/DotvvmCompilationPageConfiguration.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Threading.Tasks;
+using DotVVM.Framework.Diagnostics;
+using DotVVM.Framework.Hosting;
+using Newtonsoft.Json;
+
+namespace DotVVM.Framework.Configuration
+{
+    public class DotvvmCompilationPageConfiguration
+    {
+        public const string DefaultUrl = "_dotvvm/compilation";
```
quigamdev, line 16: "_dotvvm/diagnostics/compilation"

src/DotVVM.Framework/Configuration/DotvvmCompilationPageConfiguration.cs
```diff
@@ -0,0 +1,139 @@
...
+        public const string DefaultRouteName = "_CompilationPage";
```
quigamdev, line 17: Please change to "_dotvvm_diagnostics_compilation"

src/DotVVM.Framework/Configuration/DotvvmConfiguration.cs
```diff
@@ -141,6 +141,17 @@ public bool Debug
         }
         private bool _debug;
 
+        /// <summary>
+        /// Gets or sets the configuration for features meant for use during the development of a DotVVM application.
+        /// </summary>
+        [JsonProperty("development")]
```
quigamdev, line 147: please change to diagnostics

quigamdev (changes requested):

src/DotVVM.Framework.Hosting.Owin/AppBuilderExtensions.cs
cafour: Done. However rather unexpectedly, the `DotvvmDiagnosticsConfiguration` already exists (for 6 days already). It is in a different namespace but it might still cause confusion.

@exyi Can those two be merged somehow?

exyi: I guess you mean this: https://github.com/riganti/dotvvm/blob/main/src/DotVVM.Framework/Diagnostics/DotvvmDiagnosticsConfiguration.cs

It exists for more than 6 days, from 2017 actually. Anyway, it is public, but seems to be only for internal use, so feel free to rename it (and maybe make it internal)

src/DotVVM.Framework/Configuration/DotvvmCompilationPageConfiguration.cs
```diff
@@ -0,0 +1,139 @@
+﻿#nullable enable
```
exyi, line 1: This is already redundant

src/DotVVM.Framework/Configuration/DotvvmDiagnosticsConfiguration.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        private void ThrowIfFrozen()
+        {
+            if (isFrozen)
+                throw FreezableUtils.Error(nameof(DotvvmDiagnosticsConfiguration));
+        }
+
+        public void Freeze()
+        {
+            isFrozen = true;
```
exyi, line 35: Add `CompilationPage.Freeze()` Also, I think this method is not invoked from anywhere.

src/DotVVM.Framework/Diagnostics/CompilationPage.dothtml
```diff
@@ -0,0 +1,164 @@
...
+                    class="nav"
+                    Class-active="{value: ActiveTab == 0}"/>
+        <dot:Button Click="{staticCommand: ActiveTab = 1}"
+                    Text="Controls"
+                    class="nav"
+                    Class-active="{value: ActiveTab == 1}"/>
+        <dot:Button Click="{staticCommand: ActiveTab = 2}"
+                    Text="Master pages"
+                    class="nav"
+                    Class-active="{value: ActiveTab == 2}"/>
```
exyi, line 33:
```suggestion
                    Class-active={value: ActiveTab == 2}/>
```

src/DotVVM.Framework/Diagnostics/CompilationPage.dothtml
```diff
@@ -0,0 +1,164 @@
...
+            </dot:GridView>
+        </section>
+
+        <section Visible="{value: ActiveTab == 2}">
+            <h2>Master pages</h2>
+            <dot:GridView DataSource="{value: MasterPages}" class="nowrap">
+
+                <RowDecorators>
+                    <dot:Decorator Class-failure="{value: Status == CompilationState.CompilationFailed}" />
+                    <dot:Decorator Class-success="{value: Status == CompilationState.CompletedSuccessfully}" />
```
exyi, line 109:
```suggestion
                    <dot:Decorator Class-failure={value: Status == 'CompilationFailed'}
                                   Class-success={value: Status == 'CompletedSuccessfully'} />
```

Don't take this too seriously, just some life simplifying tips

src/DotVVM.Framework/Diagnostics/CompilationPageApiPresenter.cs
```diff
@@ -0,0 +1,40 @@
...
+            this.compilationService = compilationService;
+        }
+
+        public async Task ProcessRequest(IDotvvmRequestContext context)
+        {
+            var response = context.HttpContext.Response;
+            var isAuthorized = await context.Configuration.Diagnostics.CompilationPage.AuthorizationPredicate(context);
+            if (!isAuthorized)
+            {
+                response.StatusCode = 403;
```
exyi, line 24:
```suggestion
                response.StatusCode = 401;
```

Not very sure, but 401 means "not logged in" while 403 is "logged in, but without rights". In this case it's most likely going to be the first case.

src/DotVVM.Framework/Diagnostics/CompilationPageApiPresenter.cs
```diff
@@ -0,0 +1,40 @@
...
+                return;
+            }
+
+            var result = await compilationService.CompileAll(false);
```
exyi, line 28:
```suggestion
            var result = await compilationService.CompileAll(buildInParallel: false);
```

And actually, why not build in parallel by default?

src/DotVVM.Framework/Diagnostics/CompilationPageViewModel.cs
```diff
@@ -0,0 +1,47 @@
...
+                Context.InterruptRequest();
+                return;
+            }
+
+            await base.Init();
+        }
+
+        public async Task CompileAll()
+        {
+            await viewCompilationService.CompileAll(true, true);
```
exyi, line 39:
```suggestion
            await viewCompilationService.CompileAll(buildInParallel: true, forceRecompile: true);
```

And I don't think forcing recompilation is good default. Could we have a switch or second button for this in the UI if it's needed in some cases.

src/DotVVM.Framework/Compilation/DotHtmlFileInfo.cs
```diff
@@ -9,7 +9,9 @@ public sealed class DotHtmlFileInfo
         public CompilationState Status { get; internal set; }
         public string? Exception { get; internal set; }
 
-        /// <summary>Gets or sets the virtual path to the view.</summary>
+        /// <summary>
+        /// Gets or sets the virtual path to the view.
+        /// </summary>
```
exyi, line 14: ?

src/DotVVM.Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
```diff
@@ -1,705 +1,324 @@
...
+        
+        public string TransformText()
+        {
+            // head
+            Write(
+$@"<!DOCTYPE html>
+<html>
+    <head>
+        <title>Server Error in Application</title>
+        <meta charset=""UTF-8"" />
```
exyi, line 59:
```suggestion
        <meta charset=UTF-8 />
```

You can easily save yourself this annoying quoting quotes

src/DotVVM.Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
```diff
@@ -1,705 +1,324 @@
...
+        <div>
+");
+            foreach (var f in Formatters)
             {
-                this.WriteText(line);
-                this.Write("<br />");
+                var checkedStr = f == Formatters[0] ? "checked='checked'" : "";
+                WriteUnencoded(
+$@"
+            <input type=""radio"" id=""menu_radio_{f.Id}"" class=""collapse"" name=""menu"" {checkedStr} />
```
exyi, line 94:
```suggestion
            <input type=radio id='menu_radio_{f.Id}' class=collapse name=menu {checkedStr} />
```

And you can use single quotes when it's necessary (or if you hate unquoted attributes)

src/DotVVM.Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
```diff
@@ -1,705 +1,324 @@
...
+
+        private class SourceLine
         {
-            get
-            {
-                return this.toStringHelperField;
-            }
+            public int? LineNumber { get; set; }
+
+            public string Text { get; set; }
```
exyi, line 321: Thanks for migrating this, you are right that the T4 template did not help readability at all

exyi (approved):
I like this

src/DotVVM.Framework/Diagnostics/CompilationPageApiPresenter.cs
cafour: Yep, but 401 requires the following:

> The server generating a 401 response MUST send a WWW-Authenticate header field1 containing at least one challenge applicable to the target resource. [1]

We cannot enforce this with the authorization predicate. A 403 should mean that the server understands the requests but refuses to authorize it, which seems applicable in this situation.

\[1]: https://httpstatuses.com/401

src/DotVVM.Framework/Diagnostics/CompilationPageApiPresenter.cs
cafour: Not sure. I don't see a reason why not. This was taken directly from the status page repo.

src/DotVVM.Framework/Diagnostics/CompilationPageApiPresenter.cs
quigamdev: I think 403 is correct in this case. If the compilation page is configured to be available only on localhost without any authentication, then I believe the correct response is 403 because the server just refuses to authorize the request, not authenticate it.

quigamdev (approved):

PR #1124, opened 2021-08-28
Remove support of virtual property groups

exyi:
migrate HtmlGenericControl.Attributes to our property group dictionary

Add support for merging values in the property group dictionary, so
Attributes.Add("class", "abc") will append to existing class
attribute if it already exists.


quigamdev (approved):
Reviewed on call

quigamdev:
@exyi Just please add a test for moving context on static commands (when coping binding from one static command to another).

Issue #1079, opened 2021-07-24
New structure of the repo


tomasherceg:
There was an idea to merge some of DotVVM-related repos into the main repo (so we can simplify the process of publishing all packages together). This should also include changing the structure of the repo as it gets messier and messier.

1. Currently, we have the `/src/Tools` folder which contains build and publish scripts. We should move it to `/build`, and also clean it up - many of the scripts are not used anymore.

2. The projects in the `/src` should be reorganized to separate the following areas:
* `framework`
* `samples` - these are used internally by Selenium UI tests
* `extensions` - this folder should also include the projects from related repos: [Dynamic Data](https://github.com/riganti/dotvvm-dynamic-data), [Tracing](https://github.com/riganti/dotvvm-tracing), [Status Page](https://github.com/riganti/dotvvm-samples-compilation-status-page), [Hot Reload](https://github.com/riganti/dotvvm-diagnostics-hot-reload), and [Server Side Cache](https://github.com/riganti/dotvvm-diagnostics-server-side-cache)
* `tools` - contains various CLI tools and utilities
* `templates` - contains a `dotnet new` template

Each folder can have a subfolder called `tests` with tests for the particular area.

Here is the **proposed structure of the `src` folder** (without renaming the projects discussed in a separate issue #723): 

```POWERSHELL
framework
  tests      # do we want it here, or in the root? 
    DotVVM.Framework.PerfTests # superseded by dotvvm-benchmarks, remove
    DotVVM.Framework.Tests
    fallback-test-project   # unused -> remove
  DotVVM.Core
  DotVVM.Framework
  DotVVM.Framework.Hosting.AspNetCore
  DotVVM.Framework.Hosting.Owin
tools
  DotVVM.CommandLine
  DotVVM.Compiler
  DotVVM.Compiler.Light  # can we delete it? I think it is not used.
  DotVVM.Tools.StartupPerfTester

extensions
  api
    tests
      DotVVM.Framework.Api.Swashbuckle.AspNetCore.Tests
      DotVVM.Framework.Api.Swashbuckle.Owin.Tests
    DotVVM.Framework.Api.Swashbuckle.AspNetCore
    DotVVM.Framework.Api.Swashbuckle.Owin
    DotVVM.Framework.Api.Swashbuckle.Owin.old     # can we delete it?
  selenium
    tests
      DotVVM.Framework.Tools.SeleniumGenerator
    DotVVM.Framework.Testing.SeleniumHelpers
[...]

PR #1125, opened 2021-08-28
Changed structure of repository based on #1079

quigamdev:


cafour (changes requested):
The new structure is IMHO much better organized than the old one. A few remarks:

* I still (as in #1079) think it would be simpler to use PascalCase for all folder names with the exception of `/src/`, `/ci/`, and `/.azure/`, since the rule for using a different case is not obvious (and I for one still don't know what exactly the rule is).
* I wouldn't be afraid to collapse the `/src/framework/` group since it contains our most important projects and should thus be the easiest to access.
* Delete `/src/.vscode/`.
* Shouldn't the Framework's unit tests (`/src/tests`) be next to Framework?

quigamdev:
@cafour 
* We have been discussing the usage of pascal case. The principle that we used is only folders that contain projects using pascal case.
* In case that we collapse `src/framework` then there is necessary to differentiate project folders and just structure folders. 
* `/src/.vscode/` deleted

> Shouldn't the Framework's unit tests (/src/tests) be next to Framework?
* That is difficult to say. At this moment we have only one project that is in `/src/tests`. It seems like a good idea to place it next to framework projects. On the other hand, `/src/framework` contains only projects that are NuGet packages. So from this point of view, it does not make sense to place it into the same folder.

exyi:
>I still (as in New structure of the repo #1079) think it would be simpler to use PascalCase for all folder names with the exception of /src/, /ci/, and /.azure/, since the rule for using a different case is not obvious (and I for one still don't know what exactly the rule is).

As @quigamdev said, we'll use capital letter only for project and inside project, little letter for top-level folders. Similar to what dotnet/runtime seems to be doing: https://github.com/dotnet/runtime/tree/main/src

>Shouldn't the Framework's unit tests (/src/tests) be next to Framework?

We discussed this issue and I think it will be easier to have one unit test project for all nuget package we will put into this repo. It will make it easier to run tests and share code between them.

cafour:
@quigamdev, @exyi

>  The principle that we used is only folders that contain projects using pascal case.

That's what I though, but then again `/src/Templates` does not contain a project, but `/src/Templates/content` does. Are you sure you're not imposing a constraint that might turn out being needlessly strong?

> Similar to what dotnet/runtime seems to be doing: https://github.com/dotnet/runtime/tree/main/src

The `dotnet/runtime` repo is as far as I can see centered around its C/C++ side with majority of C# code in a flat structure under `/src/libraries/`. I'd say that the structure that `dotnet/roslyn`, `dotnet/aspnetcore`, and the other dotnet C# repos follow is more applicable in our case.

> We discussed this issue and I think it will be easier to have one unit test project for all nuget package we will put into this repo. It will make it easier to run tests and share code between them.

Fair. Is it then necessary to have the `/src/tests/` folder if it's only going to contain a single test project?

---

Overall, I'd say that the rules that you applied to folder naming are a little too strict, inflexible, and even needlessly complex. I think it's perfectly fine to have a structure folder next to a project one and not be able to differentiate between the two kinds.

quigamdev:
* `/src/Templates/content` is mistake, it should be `/src/templates/Content`
I do not mind whether the naming is pascal case or not. But the roslyn repository is actually lower case, isn't it?

quigamdev:
After discussion, we will change the folders to use PascalCase

cafour:
I renamed structure folders to PascalCase and moved `/src/Tests/UnitTests` to `/src/Tests`. I left `/src/Template/content` as is since lowercase seems to be the convention among the dotnet templates (see `/src/ProjectTemplates` in `dotnet/aspnetcore` for reference).

exyi:
@cafour Can we merge it now?

cafour:
@exyi One last thing that we discussed with @quigamdev (and that I haven't gotten around to actually do yet) is that we might want to move src/Tests to src/Framework/Tests and use that as a test project for everything in src/Framework.

exyi:
@cafour 

> One last thing that we discussed with @quigamdev (and that I haven't gotten around to actually do yet) is that we might want to move src/Tests to src/Framework/Tests and use that as a test project for everything in src/Framework.

Wait a minute, that does not look like a great idea to me, I'd like to have one main unit test project for all packages we are going to put into this repository. I like it as it's now in the in src/Tests. This way we are not going to forget to run some other unit test project (and yes, we definitely want unit tests for dynamic data, ...)

cafour (approved):

cafour:
@exyi Fair enough. Feel free to merge it. :)

PR #1127, opened 2021-08-29
Reduce redundant noise in generated HTML

exyi:
Implemented custom HtmlEncoding that takes into account if
we are writing attribute and if quotes or apostrophes are used.
We can also assume that we use UTF-8 encoding.
That allows us to encode less characters which improves readability
of the generated code and performance (I will prepare results from benchmarking before we merge this)


quigamdev (approved):
We have reviewed this PR on a call with @exyi

PR #1128, opened 2021-09-02
Disable colored output in JS builds (through MSBuild)

cafour:
I propose we ban all colored JS-build-related output from appearing in our color-disabled Visual Studio output panels.


cafour:
Unfortunately, I have not found an environment variable that would prevent rollup from printing unicode arrows, which needless to say the output panel doesn't support as well.

exyi (approved):
Sure.

tomasherceg (approved):

PR #1129, opened 2021-09-02
Capabilities registered by interfaces

exyi:
This PR adds a `IObjectWithCapability<T>` interface which can be used to register capabilities very simply. It will also mark controls implementing specific capability, which could be used for helper functions (to set html attributes, for example). The interface is only a marker.

Plus, there are few general improvements which I needed while testing this:

* Properties respect attributes on the attributeProvider AND on the declaring property. This means you can choose to put the attribute on the property or the backing field, whichever you like.
  - while it will prevent strange behavior "the attribute is here and does nothing", this is a breaking change. If you have put your attribute on the field and relied on the fact of being ignored, you'll get burnt.
  - I need this because the attributes should be respected through the capability hierarchy (most importantly data context changes)
* private GetContents method is allowed
* added ToString to ValueOrBinding. Debugging is pain without it.
* few checks


src/DotVVM.Framework/Controls/IDotvvmControl.cs
```diff
@@ -6,6 +6,13 @@
...
 namespace DotVVM.Framework.Controls
 {
+    public interface IDotvvmObjectLike
+    {
+        DotvvmBindableObject Self { get; }
+    }
+    public interface IObjectWithCapability<TCapability>: IDotvvmObjectLike
+    {
+    }
     public interface IDotvvmControl: IRenderable
```
exyi, line 16:
```suggestion
    public interface IDotvvmControl: IRenderable, IDotvvmObjectLike
```

exyi: Doccoment here

quigamdev (approved):
We have reviewed this PR on a call with @exyi. 
Please merge it after you solve the notes.

PR #1130, opened 2021-09-02
Capabilities may use existing properties

exyi:
Before, all properties were declared new, and a conflict exception
was raised. Now it's allowed if the type matches, which allow
adding capabilities to existing controls.

This PR depends on #1129

### DIFF
--- a/src/Framework/Framework/Binding/DotvvmCapabilityProperty.Helpers.cs
+++ b/src/Framework/Framework/Binding/DotvvmCapabilityProperty.Helpers.cs
@@ -1,0 +1,135 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using DotVVM.Framework.Binding.Expressions;
+using DotVVM.Framework.Controls;
+using DotVVM.Framework.Utils;
+
+namespace DotVVM.Framework.Binding
+{
+    public partial class DotvvmCapabilityProperty
+    {
+        internal static class Helpers
+        {
+            public static ValueOrBinding<T>? GetOptionalValueOrBinding<T>(DotvvmBindableObject c, DotvvmProperty p)
+            {
+                if (c.properties.TryGet(p, out var x))
+                    return ValueOrBinding<T>.FromBoxedValue(x);
+                else return null;
+            }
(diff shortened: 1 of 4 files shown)


src/DotVVM.Framework.Tests/Runtime/CapabilityPropertyTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        public sealed class TestCapability
+        {
+            public string Something { get; set; } = "kokosovina";
+            public string SomethingElse { get; set; } = "baf";
+        }
+
+        [DotvvmControlCapability]
+        public sealed class TestNestedCapability
+        {
+            public string Something { get; set; } = "abc";
```
exyi, line 116: TODO: test for case insensitivity

src/DotVVM.Framework/Utils/ReflectionUtils.cs
```diff
@@ -382,6 +383,17 @@ public static Type UnwrapTaskType(this Type type)
                 return type;
         }
 
+        public static Type UnwrapValueOrBinding(this Type type)
```
tomasherceg, line 386: Shall we also support `IValueBinding<T>` and others here?

tomasherceg (approved):

src/DotVVM.Framework/Utils/ReflectionUtils.cs
exyi: IValueOrBinding<T> does not exist, only the struct ValueOrBinding is generic

src/DotVVM.Framework.Tests/Runtime/CapabilityPropertyTests.cs
exyi: Turns out, we actually allow two DotVVM properties to have the same name with different casing... It only crashes the compiler later :facepalm: I will not change it in this PR, so we can review it separately. Also, I'm not sure what is the best solution:

* we can make the registration case insensitive, so it will throw an error when it's registered
* we can just adjust the error message compiler throws


Not sure whether the DotvvmProperty.Resolve method should case sensitive or not.

PR #1135, opened 2021-09-03
Remove some junk from generated Javascript

exyi:
* Added JsPrettificationVisitor which reduces usage of the infamous ,
   operators, removes redundant string concatenations, and merges
   variable declaration and assignment when it's easily possible
* Removed nested lambdas in static commands, one function
  is used for declaring local variables and for the options argument


quigamdev (approved):
We have reviewed this PR on call with @exyi

PR #1137, opened 2021-09-04
 Add github action for unit tests 

exyi:
When build or unit tests fail, it's quite annying to go though the Azure CI,
so I have added a simple github action that will say that.

It also checks that we can build DotVVM.Framework without warnings

### DIFF
--- a/src/Framework/Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
+++ b/src/Framework/Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
@@ -37,12 +37,16 @@ namespace DotVVM.Framework.Hosting.ErrorPages
         public string Summary { get; }
         public IErrorSectionFormatter[] Formatters { get; }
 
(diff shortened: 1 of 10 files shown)


src/Framework/Framework/Hosting/ErrorPages/ErrorPageTemplate.cs
```diff
@@ -37,12 +37,16 @@ public class ErrorPageTemplate : IErrorWriter
         public string Summary { get; }
         public IErrorSectionFormatter[] Formatters { get; }
 
-        public void WriteText(string str)
+        public void WriteText(string? str)
```
quigamdev, line 40: This seems like an unrelated change. Why these changes are here?

exyi: This only fixes the warnings that we now check in CI

quigamdev (approved):

PR #1140, opened 2021-09-04
Remove JS unit tests from Azure CI

exyi:
It only induces instability into the pipeline and it's handled by the
Github Action


quigamdev (approved):
Thanks.

PR #1141, opened 2021-09-05
Fix Framework's CompileJS BeforeTargets

cafour:
See the comment above the target for explanation.

### DIFF
--- a/src/Framework/Framework/DotVVM.Framework.csproj
+++ b/src/Framework/Framework/DotVVM.Framework.csproj
@@ -125,7 +125,7 @@
   </ItemGroup>
 
   <!-- BeforeBuild is ran for every target framework. However, unless its input files change, this target is skipped. -->
-  <Target Name="CompileJS" Inputs="@(TypescriptFile)" Outputs="obj/javascript/root-only/dotvvm-root.js;obj/javascript/root-only-debug/dotvvm-root.js;obj/javascript/root-spa/dotvvm-root.js;obj/javascript/root-spa-debug/dotvvm-root.js" BeforeTargets="BeforeBuild">
+  <Target Name="CompileJS" Inputs="@(TypescriptFile)" Outputs="obj/javascript/root-only/dotvvm-root.js;obj/javascript/root-only-debug/dotvvm-root.js;obj/javascript/root-spa/dotvvm-root.js;obj/javascript/root-spa-debug/dotvvm-root.js" BeforeTargets="DispatchToInnerBuilds;BeforeBuild">
 
     <PropertyGroup>
       <EnvOverrides>NO_COLOR=1;FORCE_COLOR=0;TERM=dumb</EnvOverrides>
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/DotVVM.Framework.csproj
```diff
@@ -124,8 +124,8 @@
     <TypescriptFile Include="Resources/Scripts/**/*.ts" />
   </ItemGroup>
 
-  <!-- DispatchToInnerBuilds is there to run it just once for all TargetFrameworks -->
-  <Target Name="CompileJS" Inputs="@(TypescriptFile)" Outputs="obj/javascript/root-only/dotvvm-root.js;obj/javascript/root-only-debug/dotvvm-root.js;obj/javascript/root-spa/dotvvm-root.js;obj/javascript/root-spa-debug/dotvvm-root.js" BeforeTargets="DispatchToInnerBuilds">
+  <!-- BeforeBuild is ran for every target framework. However, unless its input files change, this target is skipped. -->
+  <Target Name="CompileJS" Inputs="@(TypescriptFile)" Outputs="obj/javascript/root-only/dotvvm-root.js;obj/javascript/root-only-debug/dotvvm-root.js;obj/javascript/root-spa/dotvvm-root.js;obj/javascript/root-spa-debug/dotvvm-root.js" BeforeTargets="BeforeBuild">
```
exyi, line 128:
```suggestion
  <Target Name="CompileJS" Inputs="@(TypescriptFile)" Outputs="obj/javascript/root-only/dotvvm-root.js;obj/javascript/root-only-debug/dotvvm-root.js;obj/javascript/root-spa/dotvvm-root.js;obj/javascript/root-spa-debug/dotvvm-root.js" BeforeTargets="DispatchToInnerBuilds;BeforeBuild">
```


Actually, I just figured out there can be both of them and it works :D

* when building for multiple targets, it is executed only once
* when building only for one target, it also gets executed

cafour: Seems a little redundant to me, since when building with `DispatchToInnerBuilds;BeforeBuild` and both framework targets then `CompileJS` is hit 3 times (skipped twice), while with just `BeforeBuild` it is hit just twice (and skipped once).

Either way we have to depend on MSBuild to make the right decision and run `CompileJS` just the first time it is hit if ever.

exyi: For me, this executes the JS build twice in parallel... with both targets, it only executes it once (as it's skipped twice)

exyi: When you look at the CI, it also runs twice in parallel. The "Github Actions / dotnet unit tests (windows-latest)" seems to fail for this reason (apparently yarn install is not thread safe or something :sweat_smile: )

cafour: My bad. Forgot to clean the repo before trying this out. Sorry.

exyi: No problem, we can just MSBuild for causing all these problems :D

tomasherceg (approved):

PR #1142, opened 2021-09-05
Remove support for Internet Explorer

exyi:


acizmarik (approved):
I suppose the file `mono_crash.21acb2be5c.0.json` was added by mistake - it should be deleted before merging. Other than that looks good!

PR #1143, opened 2021-09-06
Changed visibility of `TypeInferer` to public

acizmarik:
As we are adding support for lambda functions to VS Extension, it would be nice to have also type inference capabilities for the Intellisense. Using this we could provide type information for individual parameters as the user is typing a lambda function.

### DIFF
--- a/src/Framework/Framework/Compilation/Inference/ITypeInferer.cs
+++ b/src/Framework/Framework/Compilation/Inference/ITypeInferer.cs
@@ -9,7 +9,7 @@ using DotVVM.Framework.Compilation.Inference.Results;
 
 namespace DotVVM.Framework.Compilation.Inference
 {
-    internal interface ITypeInferer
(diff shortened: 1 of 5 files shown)


exyi:
Ok, we can make this class public, although it's not really a public API.

Would you please add InternalsVisibleTo attribute for VS extension? There is going to be more things which could be useful for VS Extension but are not really part of our API.

acizmarik:
IMO almost everything regarding compilation should be internal. That being said the `InternalsVisibleTo` should be definitely added for VSExtension and is a better solution of this.

I am just wondering if we want to refactor visibility of some classes that are now public only due to being used by VSExtension. Since I am almost certain that nobody outside of DotVVM team use those. Also we can make some breaking changes before the upcomming major release 😄.

exyi:
I think we decided that we should rather use the internal visible to attribute, so I'll close this PR.

PR #1145, opened 2021-09-09
Validation - absolute paths

MichalTichy:
Reworked validation paths to be use better syntax for validation paths.

### New features

* New format of the validation property paths:
   * **/** - target is root
   * **/exampleProp/A** - returns prop A located withing property exampleProp located at provided context
   * **/exampleProp/B/1** - returns second item from collection B located within property exampleProp located at provided context.
* It is now possible to set validation errors that are outside of validation target
* New extension methods on `ValidationErrorFactory` to support all identified use-cases

### Breaking changes
* Validation target paths are now specified using new format as showcase above
* `ViewModelValidationError` no longer has public ctor and its property setters are internal
   * Users should use `ValidationErrorFactory` and its various extension methods for adding validation errors
* Removed static method `BuildPath<TValidationTarget>` on class `PropertyPath` that was used in past to generate knockout property path expressions. This method no longer works and has been obsolete for quite some time.
* An exception is now thrown whenever validation target is bound to a property that gets evaluated to `null` during postback

### Future plans (will be addressed in separate PRs)
* Validation for static commands
* JS API for adding validation errors
* Possibility to return partially changed viewmodel together with some validation errors

This PR supersedes #979.


acizmarik:
PR review comments from #979 should be already addressed. 

There are still some features to be implemented (see Future plans in description). However, we decided to resolve these within a separate PR. There should be no more breaking changes though.

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -1,18 +1,51 @@
...
+/**
+ * Traverses provided context according to given path.
+ * @example / - returns context
+ * @example "" or null - returns context
+ * @example /exampleProp/A - returns prop A located withing property exampleProp located at provided context
+ * @example /exampleProp/B/1 - returns second item from collection B located within property exampleProp located at provided context
+ * @returns found property as unwrapped object
+ */
+export function traverseContext(context: any, path: string): any 
+{
```
quigamdev, line 23: If I understand this function correctly, path = `/` will return the current view model and not the root. Am I wrong?

MichalTichy: `/` returns provided context.
Based on validation target this may or may not be root.
Root VM is used for application of validation errors from server.

DataContext of origin element (sender) is used for client side validation.

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -44,26 +44,27 @@ const createValidationHandler = (path: string) => ({
...
-        return callback()
+        return callback();
     }
 })
 
 export function init() {
     postbackHandlers["validate"] = (opt) => createValidationHandler(opt.path);
-    postbackHandlers["validate-root"] = () => createValidationHandler("dotvvm.viewModelObservables['root']");
-    postbackHandlers["validate-this"] = () => createValidationHandler("$data");
+    postbackHandlers["validate-root"] = () => createValidationHandler("/");
```
MichalTichy, line 66: Those handlers are not necessary.
We can remove them and keep only the first one but this may hinder readability.

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
```diff
@@ -90,8 +111,10 @@ public static ValidationResult CreateValidationResult<T>(ValidationContext valid
                 var dataContext = DataContextStack.Create(e.expression.Parameters.Single().Type);
                 var expression = ExpressionUtils.Replace(e.expression, BindingExpressionBuilder.GetParameters(dataContext).First(p => p.Name == "_this"));
                 var jsast = translator.CompileToJavascript(expression, dataContext);
```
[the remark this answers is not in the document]
acizmarik: I tried this. However, I found out that property renaming was not resolved (if I recall correctly). Once it was a JS AST it was correct

src/Framework/Framework/ViewModel/Validation/ViewModelValidationError.cs
```diff
@@ -1,21 +1,34 @@
...
+        /// <summary>
+        /// Gets or sets the error message.
+        /// </summary>
+        [JsonProperty("errorMessage")]
+        public string ErrorMessage { get; internal set; }
+
         /// <summary>
         /// Contains path that can be evaluated on the client side.
-        /// E.g.: Product().Suppliers()[2].Name
+        /// E.g.: Product/Suppliers/2/Name
```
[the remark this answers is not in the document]
acizmarik: Yes, it should. I am committing the suggestion.

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
```diff
@@ -24,18 +25,41 @@ public static class ValidationErrorFactory
             where T : class, IDotvvmViewModel =>
             AddModelError(vm, expr, message, vm.Context);
 
-        public static ViewModelValidationError AddModelError<T, TProp>(T vm, Expression<Func<T, TProp>> expr, string message, IDotvvmRequestContext context)
+        public static ViewModelValidationError AddModelError<T, TProp>(this T vm, Expression<Func<T, TProp>> expr, string message, IDotvvmRequestContext context)
```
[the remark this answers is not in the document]
acizmarik: I wanted to create a somewhat unified API. However, now that I think about it - would it be a bad idea to have all these extensions on the `IDotvvmRequestContext`? We already have there for example the `FailOnInvalidModelState`.

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -1,18 +1,51 @@
...
+    if (path == null)
+        path = "";
+
+    var parts = path.split(/[/[\.\]]+/);
```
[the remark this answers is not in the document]
MichalTichy: Supporting all the old paths was not a goal (but we can definitely at least try to).
The idea was to handle common mistakes like using indexers or navigation using `.` .

@exyi
Do you think that we should also support paths containing `()`?

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -1,18 +1,51 @@
...
+    var currentLevel = context;
+    var currentPath = "";
+    for (var i = 0; i < parts.length; i++) {
+        if (shouldBeConvertedFromDataContext(currentLevel, parts.slice(i))) {
```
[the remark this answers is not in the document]
MichalTichy: There are some use cases when we get $data and $rawData in path.
- `$data`
   - Is received when we are validating current context (`validate-this`). This could be replaced by root path ('/').
- `$rawData`
   -   If I'm correct this comes as part of validation path with changed data context (@acizmarik knows more about the server part). We were not able to get of this unfortunately.

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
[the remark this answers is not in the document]
acizmarik: Mentioned methods were moved to `DotvvmRequestContext` others remained in the factory

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -44,26 +44,27 @@ const createValidationHandler = (path: string) => ({
...
+    postbackHandlers["validate-this"] = () => createValidationHandler("$data"); //TODO
```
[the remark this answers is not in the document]
MichalTichy:
```suggestion
    postbackHandlers["validate-this"] = () => createValidationHandler("$data"); 
```
I've decided to keep it as is.

acizmarik:
Since there was a lot of confusion on how this actually works, I will try to describe the server part in this comment. Right now there are three ways how to create validation errors.

### **Validation attributes**
   * **Applied**: only if affected object is within the specified `validationTarget`
   * **Checked by**: `ViewModelValidator`
      * If violated => create new unresolved `ViewModelValidationError` with information about `TargetObject` and `PropertyPath`
         * `TargetObject` is the owner object of the member (property) affected by this attribute
         * `PropertyPath` is relative path from `TargetObject` to the violating member (property)
      * If not violated => skip
   * **Calculate absolute path using**: `ValidationErrorPathExpander`
      * Traverse whole viewmodel from root and try to find `TargetObject`
      * If `TargetObject` was found => concatenate current path (path from root to `TargetObject`) with the relative `PropertyPath` and mark the corresponding error as resolved
      * Once whole viewmodel was traversed, remove all unresolved errors (they are not reachable from root)
   * **Result**: Server sends always absolute paths


### **Manually added errors using expressions**
   * **Applied**: always (does not respect `validationTarget`)
       * create new unresolved `ViewModelValidationError` with information about `TargetObject` and `PropertyPath`, both of which is extracted from the provided lambda function (expression)
   * **Calculate absolute path using**: `ValidationErrorPathExpander` (same approach as with attributes)
   * **Result**: Server sends always absolute paths

### **Manually added errors using hand-written paths**
   * **Applied**: always (does not respect `validationTarget`)
       * yields new resolved `ViewModelValidationError` with absolute `PropertyPath`
   * **Note**: Provided path is not fully checked on server side due to various complications, such as property renaming and possible viewmodel modifications in javascript
       * We just check it "looks" like an absolute path
   * **Result**: Server sends always absolute paths (but they might be malformed - i.e. need to be validated on client's side)

src/Framework/Framework/Compilation/Javascript/PropertyPathExtractingVisitor.cs
```diff
@@ -1,32 +1,37 @@
...
+            var sb = new StringBuilder();
+            while (stack.Count > 0)
+            {
+                if (sb.Length > 0)
+                    sb.Append($"/{stack.Pop()}");
+                else
+                    sb.Append(stack.Pop());
+            }
+
+            return (sb.Length > 0) ? sb.ToString() : "/";
```
exyi, line 34: I guess the path should start with a `/`, right?

src/Framework/Framework/Controls/KnockoutHelper.cs
```diff
@@ -131,16 +128,11 @@ string getHandlerScript()
                 var validationPath = expression is IStaticCommandBinding ? null : GetValidationTargetExpression(control);
                 return GetPostBackHandlersScript(control, propertyName,
                     // validation handler
-                    validationPath == null ? null :
-                    validationPath == RootValidationTargetExpression ? "\"validate-root\"" :
-                    validationPath == "$data" ? "\"validate-this\"" :
-                    $"[\"validate\", {{path:{JsonConvert.ToString(validationPath)}}}]",
+                    validationPath == null ? null : $"[\"validate\", {{path:{JsonConvert.ToString(validationPath)}}}]",
```
exyi, line 131: I'd vote for keeping the validate-root shortcut - it reduces the number of quotes in the expression

src/Framework/Framework/Controls/KnockoutHelper.cs
```diff
@@ -351,7 +343,7 @@ private static JsExpression TransformOptionValueToExpression(DotvvmBindableObjec
             }
         }
 
-        private const string RootValidationTargetExpression = "dotvvm.viewModelObservables['root']";
+        private const string RootValidationTargetExpression = "/";
```
exyi, line 346: At this point, I think we can just inline this constant.

src/Framework/Framework/Resources/Scripts/postback/postbackCore.ts
```diff
@@ -202,16 +202,6 @@ function updateDynamicPathFragments(context: any, path: string[]): void {
     }
 }
 
-function processPassedId(id: any, context: any): string {
```
exyi, line 205: :thinking: Wondering how postbacks in markup control in client-rendered repeater work without this

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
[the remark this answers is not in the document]
exyi: @MichalTichy so, can we remove this?

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -1,18 +1,51 @@
...
+            currentLevel = context["$data"];
+        }
+        let expressionPart = parts[i];
+        if (expressionPart === "")
+            continue;
+
+        var currentLevelExpanded = currentLevel instanceof Function ? ko.unwrap(currentLevel) : currentLevel;
+
+        var nextNode = currentLevelExpanded[expressionPart];
```
exyi, line 40:
```suggestion
        var nextNode = ko.unwrap(currentLevel)[expressionPart];
```

src/Framework/Framework/Resources/Scripts/validation/error.ts
```diff
@@ -11,11 +11,11 @@ export function detachAllErrors() {
 }
 
 export function getErrors<T>(o: KnockoutObservable<T> | null): ValidationError[] {
-    o = unwrapComputedProperty(o);
+    let unwrapped = unwrapComputedProperty(o);
```
exyi, line 14:
```suggestion
    const unwrapped = unwrapComputedProperty(o);
```

src/Framework/Framework/Resources/Scripts/validation/error.ts
```diff
@@ -31,18 +31,18 @@ export class ValidationError {
             throw new Error(`ValidationError cannot be attached to "${observable}".`);
         }
 
-        observable = unwrapComputedProperty(observable);
-        if (!observable.hasOwnProperty(ErrorsPropertyName)) {
-            observable[ErrorsPropertyName] = [];
+        let unwrapped = unwrapComputedProperty(observable);
```
exyi, line 34:
```suggestion
        const unwrapped = unwrapComputedProperty(observable);
```

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -44,26 +44,27 @@ const createValidationHandler = (path: string) => ({
...
+                    validateViewModel(validationTarget);
+                });
 
             if (allErrors.length > 0) {
                 logError("validation", "Validation failed: postback aborted; errors: ", allErrors);
                 return Promise.reject(new DotvvmPostbackError({ type: "handler", handlerName: "validation", message: "Validation failed" }))
             }
         }
-        return callback()
+        return callback();
```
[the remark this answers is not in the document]
exyi: ?

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -44,26 +44,48 @@ const createValidationHandler = (path: string) => ({
...
+const runClientSideValidation = (validationTarget:any,options:PostbackOptions) => {
+
+    watchAndTriggerValidationErrorChanged(options,
+        () => {
+            detachAllErrors();
+            validateViewModel(validationTarget);
+        });
+}
+
+const createRootValidationHandler = () => ({
```
exyi, line 69: Is this used anywhere?

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
```diff
@@ -20,22 +21,25 @@ namespace DotVVM.Framework.ViewModel.Validation
 {
     public static class ValidationErrorFactory
     {
-        public static ViewModelValidationError AddModelError<T, TProp>(this T vm, Expression<Func<T, TProp>> expr, string message)
-            where T : class, IDotvvmViewModel =>
-            AddModelError(vm, expr, message, vm.Context);
+        public static ViewModelValidationError AddModelError<T>(this T vm, string message)
+            where T : class, IDotvvmViewModel
+            => vm.Context.AddModelError(message);
```
exyi, line 26: In this case, the error should be bound to the passed object
```suggestion
            => vm.Context.AddModelError(vm, a => a, message);
```

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
```diff
@@ -20,22 +21,25 @@ namespace DotVVM.Framework.ViewModel.Validation
...
+
+        public static ViewModelValidationError AddModelError<T>(this T vm, string propertyPath, string message)
+            where T : class, IDotvvmViewModel
+            => vm.Context.AddModelError(propertyPath, message);
```
exyi, line 30: And this too.

src/Tests/ControlTests/testoutputs/ServerSideStyleTests.PostbackHandlers.html
```diff
@@ -2,16 +2,16 @@
 	<head></head>
 	<body>
 		One handler
-		<input onclick="dotvvm.postBack(this,[],&quot;8UxwOb8axY9NSuJp&quot;,&quot;&quot;,null,[[&quot;confirm&quot;,{message:&quot;a&quot;}],&quot;validate-root&quot;],[],undefined).catch(dotvvm.log.logPostBackScriptError);event.stopPropagation();return false;" type="button" value="">
+		<input onclick="dotvvm.postBack(this,[],&quot;8UxwOb8axY9NSuJp&quot;,&quot;&quot;,null,[[&quot;confirm&quot;,{message:&quot;a&quot;}],[&quot;validate&quot;, {path:&quot;/&quot;}]],[],undefined).catch(dotvvm.log.logPostBackScriptError);event.stopPropagation();return false;" type="button" value="">
```
exyi, line 5: This is exactly why I'd like to keep the validate-root shortcut...

src/Tests/ViewModel/PropertyPathTests.cs
```diff
@@ -27,9 +27,9 @@ public void PropertyPath_GetPropertyPath()
 
             viewModel.Context.ModelState.ValidationTarget = viewModel;
 
-            var error = viewModel.AddModelError(vm => vm.Users[0].Post.PostText, "Testing validation error.");
+            var error = ValidationErrorFactory.AddModelError(viewModel, vm => vm.Users[0].Post.PostText, "Testing validation error.");
 
-            Assert.AreEqual("Users()[0]().Post().PostText", error.PropertyPath);
+            Assert.AreEqual("Users/0/Post/PostText", error.PropertyPath);
```
exyi, line 32: Aren't the path supposed to start with /?

src/Framework/Framework/Compilation/Javascript/PropertyPathExtractingVisitor.cs
acizmarik: This is used only internally for extracting paths from provided lambda functions (`TargetObject` -> Property). Path prefix (root -> `TargetObject`) is calculated later. So it should not be absolute at this point. However, the slash is not necessary - I am removing it. Since we are just concatening paths it does not really matter it is an empty string or an slash. If somebody provides an identity `vm => vm`, the result path should be just the prefix.

I am also adding validation that lambda expressions do not contain unsupported expression, such as conditional expressions.

src/Framework/Framework/Controls/KnockoutHelper.cs
acizmarik: `validate-root` is returned

src/Framework/Framework/Controls/KnockoutHelper.cs
acizmarik: Agree. Inlined.

src/Tests/ViewModel/PropertyPathTests.cs
acizmarik: This test does not run the `ValidationErrorPathExpander` thats why it is not absolute. This is basically testing internal API - user would not see this. It just tests extracting paths from lambda functions

src/Tests/ControlTests/testoutputs/ServerSideStyleTests.PostbackHandlers.html
acizmarik: I added the shortcut in `b812885`. Although it seems like the file was changed in `main` and no longer contains `validate-root`... Regardless, it is back so I am marking as resolved.

src/Framework/Framework/Resources/Scripts/validation/validation.ts
acizmarik: It is used to create a handler for `validate-root` that was just resurrected 😄.

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
acizmarik: I don't see why it would be helpful here. We do not work much with errors that have hand-written property paths. The only thing we could check is whether `vm` is reachable from root viewmodel.

But then again, I remember us having discussion that this API is only "useful" when user is doing some weird stuff with viewmodels in javascript. Due to this reason I don't feel like we want to do any checks here. What do you think?

src/Framework/Framework/Resources/Scripts/validation/validation.ts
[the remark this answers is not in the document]
acizmarik: Removed

src/Framework/Framework/ViewModel/Validation/ValidationErrorFactory.cs
acizmarik: After internal discussion we decided that this overload will be removed.

src/Framework/Framework/Resources/Scripts/postback/postbackCore.ts
acizmarik: This is being tracked in a separate issue #1218.

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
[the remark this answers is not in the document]
acizmarik: Removed

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
acizmarik: This is already resolved. `/` always means root viewmodel. Same as `validate-root`. Moreover, `validate-this` is removed

exyi (approved):

cafour (approved):
I haven't followed this PR very closely but at the very least I went through the error messages and found no typos. :D

quigamdev:
Just, we need to fix the tests before merging it.

PR #1146, opened 2021-09-09
Serialize all properties into the dotvvm_serialized_config.json.tmp

exyi:
This should help the VS extension with all the new ways to register
dotvvm properties. Could also help humans to debug why some property
does not exist.

### DIFF
--- a/src/Framework/Framework/Binding/DotvvmProperty.cs
+++ b/src/Framework/Framework/Binding/DotvvmProperty.cs
@@ -450,6 +450,7 @@ namespace DotVVM.Framework.Binding
         {
             return registeredAliases.Values;
         }
+        public static IEnumerable<DotvvmProperty> AllProperties => registeredProperties.Values;
 
         public override string ToString()
         {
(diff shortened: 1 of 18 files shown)


src/Framework/Framework/Binding/DotvvmProperty.cs
```diff
@@ -345,6 +345,8 @@ public static DotvvmProperty[] ResolveProperties(Type type)
             return registeredProperties.Values.Where(p => types.Contains(p.DeclaringType)).ToArray();
         }
 
+        public static IEnumerable<DotvvmProperty> AllProperties => registeredProperties.Values;
+
```
cafour, line 349: This is a duplicate of `GetRegisteredProperties` introduced by property aliasing. I don't care which version survives (`AllProperties` is shorter, so I'd actually prefer it), just please don't forget to add `AllAliases` when you merge `main` into this branch. Thanks. :)

cafour (changes requested):

src/Framework/Framework/Binding/DotvvmProperty.cs
exyi: Done, would please have another look?

cafour (approved):

PR #1148, opened 2021-09-11
Simplify generated JS: use ?. and ?? operators

exyi:


exyi:
The core of this in the JavascriptNullCheckAdder.cs  which was in UTF16 encoding for some reason and github refuses to show the diff :facepalm:

acizmarik (approved):
Nice 👍. I checked `JavascriptNullCheckAdder.cs` locally. It's good to see that generated code will be simplified.

PR #1149, opened 2021-09-12
Capability properties will implicitly use capability name as property name

exyi:
More specifically `prefix + capabilityType.Name` is used as a default name for
these DotvvmProperties. When declared using RegisterCapability method,
name other can be specified. However, properties declared implicitly
through GetContents arguments or through other capabilities
will ignore the argument/property name and use the
capability name instead


tomasherceg (approved):
Nice

acizmarik (approved):
Nice! I like the way it simplified usage of the API

PR #1156, opened 2021-09-21
Added infrastructure for Roslyn Analyzers

acizmarik:
This PR introduces initial support for adopting Roslyn Analyzers and implements a couple of analyzers as a proof-of-concept.

### Added analyzers
1. [**DotVVM02**; **Serializability**; *Use only serializable properties in viewmodels*]: the purpose of this analyzer is to check that for all viewmodels its specified properties are serializable. User can use only DotVVM-friendly types.
2. [**DotVVM03**; **Serializability**; *Do not use public fields in viewmodels*]: the purpose of this analyzer is to notify user to use only properties for storing state of viewmodels.
3. [**DotVVM04**; **Serializability**; *Do not use abstract types for properties in viewmodels*]: the purpose of this analyzer is to notify user that only concrete types should be used in viewmodels. This is recommended to ensure that all properties are correctly serialized.

### Testing
There are two new projects that can be utilized when testing analyzers.

1. `DotVVM.Analyzers.Tests` is a unit testing project.
2. `DotVVM.Analyzers.Vsix` is a simple extension for Visual Studio. It can be marked as a starting project and once run, analyzers can be tested in a separate experimental VS instance.


src/Analyzers/Analyzers.CodeFixes/AnalyzersCodeFixProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+//using System.Threading.Tasks;
+//using Microsoft.CodeAnalysis;
+//using Microsoft.CodeAnalysis.CodeFixes;
+//using Microsoft.CodeAnalysis.CodeActions;
+//using Microsoft.CodeAnalysis.CSharp;
+//using Microsoft.CodeAnalysis.CSharp.Syntax;
+//using Microsoft.CodeAnalysis.Rename;
+//using Microsoft.CodeAnalysis.Text;
+
+//namespace DotVVM.Analyzers
```
quigamdev, line 16: Why is this commented?

src/Analyzers/Analyzers.CodeFixes/DotVVM.Analyzers.CodeFixes.csproj
```diff
@@ -0,0 +1,51 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <VersionPrefix>4.0.0-preview01-116912</VersionPrefix>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
quigamdev, line 5: Could we use netstandard2.1? Is there any restriction?

src/Analyzers/Analyzers.Tests/DotVVM.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
quigamdev, line 4: Could we use net50?

src/Analyzers/Analyzers.Tests/Verifiers/VisualBasicAnalyzerVerifier`1+Test.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Microsoft.CodeAnalysis.VisualBasic.Testing;
+
+namespace DotVVM.Analyzers.Tests
+{
+    public static partial class VisualBasicAnalyzerVerifier<TAnalyzer>
+        where TAnalyzer : DiagnosticAnalyzer, new()
+    {
+        public class Test : VisualBasicAnalyzerTest<TAnalyzer, MSTestVerifier>
```
quigamdev, line 10: Can you remove all VB parts? It does not make much sense to keep it in repo.

src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        private static void AnalyzeSyntaxTree(SemanticModelAnalysisContext context)
+        {
+            var semanticModel = context.SemanticModel;
+            var syntaxTree = context.SemanticModel.SyntaxTree;
+            var viewModelInterface = semanticModel.Compilation.GetTypeByMetadataName(dotvvmViewModelInterfaceMetadataName);
+            var bindAttribute = semanticModel.Compilation.GetTypeByMetadataName(dotvvmBindAttributeMetadataName);
+
+            // Check all classes
+            foreach (var classDeclaration in syntaxTree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>())
```
quigamdev, line 77: Could we verify whether there is some API that we just add a listener and catch all events for changes of Class Declaration?

src/Analyzers/Analyzers/Serializability/TypeSymbolExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+                    return ((INamedTypeSymbol)typeSymbol).EnumUnderlyingType.IsSerializationSupportedImpl();
+                case TypeKind.Class:
+                case TypeKind.Struct:
+
+                    var namedTypeSymbol = typeSymbol as INamedTypeSymbol;
+                    var arity = namedTypeSymbol?.Arity;
+                    var args = namedTypeSymbol?.TypeArguments ?? ImmutableArray.Create<ITypeSymbol>();
+                    var symbol = (arity.HasValue && arity.Value > 0) ? typeSymbol.OriginalDefinition : typeSymbol;
+
+                    if (symbol.IsPrimitive() || supportedTypesForSerialization.Contains(symbol))
```
quigamdev, line 60: Do you check for self-referencing types? Otherwise, StackOverflow can happen.

src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+                    continue;
+
+                if (property.Type.Kind() == SyntaxKind.NullableType)
+                {
+                    // Serialization of nullable type
+                    var nullable = property.Type as NullableTypeSyntax;
+                    if (semanticModel.GetSymbolInfo(nullable!.ElementType).Symbol is not ITypeSymbol elementInfo)
+                        continue;
+
+                    if (elementInfo.IsPrimitive())
```
quigamdev, line 115: This most probably will add a diag message for DateTime? as unserializable.
How does this take in count reference nullable types?

src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+            var semanticModel = context.SemanticModel;
+            foreach (var property in viewModel.DescendantNodes().OfType<PropertyDeclarationSyntax>()
+                .Where(f => f.Modifiers.Any(m => m.Kind() == SyntaxKind.PublicKeyword)))
+            {
+                if (semanticModel.GetDeclaredSymbol(property) is not IPropertySymbol propertySymbol)
+                    continue;
+                if (semanticModel.GetSymbolInfo(property.Type).Symbol is not ITypeSymbol propertyTypeSymbol)
+                    continue;
+
+                if (IsSerializationIgnored(propertySymbol, bindAttribute))
```
quigamdev, line 105: This only verifies that you used the bind attribute. But we should verify whether it is used with bind.none or verify that the type is serializable. Also, we should check JsonIgnoreAttributes

PR #1157, opened 2021-09-23
Working with bindings in server-side styles and CompositeControls

exyi:
* Added x.Select(x => ...) method to ValueOrBinding and bindings
   - this method will still be quite slow, but it should not be terrible, so it should be safe to use even in runtime controls
   - obviously, usage in server-side style is preferred
* Added .And and .Or extension methods to ValueOrBinding.
   - useful for setting Visible when it's already set to something - in that case we can use the And method to have both conditions
* Added support for ValueOrBinding to styles
* Asking for c.Property(...) in styles will return the default value, if the property does not exist on this element.
* Deduplicate css `class`es in attributes when set using styles.
* Added helpers like IsNull, IsNullOrEmpty, ... to ValueOrBinding. Includes helpers NotNull, NotNullOrEmpty, since negating a ValueOrBinding is bit more annoying
   - also, the IsNullOrEmpty JS translation is now more concise (most importantly does not duplicate the expression)

### DIFF
--- a/src/Framework/Framework/Binding/BindingCombinator.cs
+++ b/src/Framework/Framework/Binding/BindingCombinator.cs
@@ -32,6 +32,16 @@ namespace DotVVM.Framework.Binding
                     )
                 )
             );
+        static IBinding CreateOrElseCombination(IBinding a, IBinding b) =>
+            a.DeriveBinding(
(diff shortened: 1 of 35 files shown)


quigamdev (approved):
We have reviewed this PR on call with @exyi

PR #1166, opened 2021-10-09
Fixed an issue with ambiguous serialization of `StaticCommandInvocationPlan`

acizmarik:
This PR fixes an issue with serialization and deserialization of `StaticCommandInvocationPlan`. Previously only method name and somewhat limited information about arguments was passed. This caused issues on deserialization as in some cases the method could not be matched (for example similar method overloads).

cc @martindybal


acizmarik:
I just realized that this change breaks tests `MarkupControlTests.MarkupControl_PassingStaticCommand` as the metadata token changes across different builds. Do we need it or can I remove it? Right now it fails because the encrypted part changes

acizmarik:
Fixed. Right now full type name gets passed. This should be more consistent compared to metadata tokens.

PR #1180, opened 2021-10-20
Refactor capabilty registration: split type scanning from code generation

exyi:
This allows us to store the scanned metadata in the capabilty property
(in PropertyMapping and PropertyGroupMapping), which will allow us to
enable usage of capabilities in server-side styles.

I also changed the capabilities to be records with init-only setters.
This will prevent patterns like control.GetCapability<X>().Prop = "a",
which does not work. Also, it provides a ToString implementation,
which is nice for debugging


quigamdev (approved):
We are merging this PR and a grammar check will be done as a separate PR.

PR #1187, opened 2021-10-23
ToBrowserLocalTime implemented

tomasherceg:
I've implemented a prototype of ToBrowserLocalTime extension method on DateTime that could be used to convert UTC dates into client local times.

There are some test cases missing:
- nullability support
- server rendering (the method doesn't know the client's timezone on the server, so it does no conversion)
- I am using TimeShift JS library to simulate fixed offset on the client-side, but I haven't tested it with different timezones than I have set on my machine so I don't know if it works properly

I was also wondering about adding a method `InExplicitTimeZone(tz)`, but there is a problem with representing timezones on the client. All JS libraries are using IANA timezones, but .NET has its own timezone representation. 
IANA timezone IDs will be supported from .NET 6 so maybe we can introduce this feature to .NET 6 version of DotVVM. Or maybe it can be solved by NodaTime.

### DIFF
--- a/src/Framework/Framework/Binding/HelperNamespace/DateTimeExtensions.cs
+++ b/src/Framework/Framework/Binding/HelperNamespace/DateTimeExtensions.cs
@@ -9,12 +9,14 @@ namespace DotVVM.Framework.Binding.HelperNamespace
         /// Converts the date (assuming it is in UTC) to browser's local time.
         /// CAUTION: When evaluated on the server, no conversion is made as we don't know the browser timezone.
         /// </summary>
+        [Obsolete("When evaluated on the server, no conversion is made as we don't know the browser timezone.")]
         public static DateTime ToBrowserLocalTime(this DateTime value) => value;
 
         /// <summary>
         /// Converts the date (assuming it is in UTC) to browser's local time.
         /// CAUTION: When evaluated on the server, no conversion is made as we don't know the browser timezone.
         /// </summary>
+        [Obsolete("When evaluated on the server, no conversion is made as we don't know the browser timezone.")]
         public static DateTime? ToBrowserLocalTime(this DateTime? value) => value;
 
     }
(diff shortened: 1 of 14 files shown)


exyi:
Maybe we could add these attributes to warn people that this method will now work server-side: https://devblogs.microsoft.com/dotnet/announcing-net-6/#enable-custom-guards-for-platform-compatibility-analyzer

src/Framework/Framework/Binding/HelperNamespace/DateTimeExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace DotVVM.Framework.Binding.HelperNamespace
+{
+    public static class DateTimeExtensions
+    {
+
+        /// <summary>
+        /// Converts the date (assuming it is in UTC) to browser's local time.
+        /// CAUTION: When evaluated on the server, no conversion is made as we don't know the browser timezone.
+        /// </summary>
+        public static DateTime ToBrowserLocalTime(this DateTime value) => value;
```
quigamdev, line 12: This is not nice! If you have BusinessPack export to excel etc... this will export the wrong data.

quigamdev (changes requested):

src/Framework/Framework/Binding/HelperNamespace/DateTimeExtensions.cs
tomasherceg: What is your suggestion then?

If someone wants to generate an Excel file on the server, they will need to know the user's timezone. There is no standard way how to send it to the server, and we decided recently that we don't want to introduce such mechanism in the framework.

We can leave this to the BP users - maybe they will figure the timezone out somehow, or maybe exporting the dates in UTC makes sense in their case.

I just don't see why we should block this framework feature because of this. 
We cannot throw the exception here because many controls evaluate getters, and we want the server rendering to render the UTC date.

quigamdev: After an internal discussion with @tomasherceg,  we could not find any suitable solution. So, it can go as it is.

quigamdev (approved):

PR #1195, opened 2021-10-27
Fixed sorting for configuration serialization

MichalTichy:


src/Tests/Runtime/config-tests/ConfigurationSerializationTests.AuxOptions.json
```diff
@@ -1,28 +1,28 @@
...
     "applicationPhysicalPath": "/opt/myApp",
+    "clientSideValidation": false,
+    "debug": true,
+    "defaultCulture": "cs-CZ",
+    "diagnostics": {
+      "compilationPage": {}
+    },
+    "experimentalFeatures": {},
     "markup": {
+      "ViewCompilation": {
```
tomasherceg, line 12: We should change this parameter to be serialized as `viewCompilation`.

tomasherceg (commented):
I've found more places where the property names are pascal case instead of camel case. 
I think we should consider reformatting this before we start to use this intensively in VS extension and other tools. What do you think?

exyi:
done more simply in #1200

PR #1196, opened 2021-10-31
Add translations for Convert.ToX for numeric types

exyi:
This is useful for invoking a specific overload of a method (for example Math.Round),
only conversions which do basically nothing are implemented.

We don't support casting syntax, and this is easier to implement :sweat_smile:


quigamdev (approved):
Thanks 👍

acizmarik (approved):
👍

PR #1201, opened 2021-11-03
Replace 'can not' with 'cannot'

cafour:
"Cannot" is far more common than "can not". [[1]] "Can not" is possible though, just not in these cases. [[2]] :D

[1]: https://www.merriam-webster.com/words-at-play/cannot-vs-can-not-is-there-a-difference
[2]: https://www.grammarly.com/blog/cannot-or-can-not/

(There are also some changes to trailing whitespace. I turned off all auto trimming features in vscode but it still created them. Sorry about that.)


exyi (approved):
Thanks :sweat_smile: 

> (There are also some changes to trailing whitespace. I turned off all auto trimming features in vscode but it still created them. Sorry about that.)

no big deal

PR #1202, opened 2021-11-03
Fixed null propagation for extension methods

### LINKED ISSUE #1185: JS Linq methods do not work in nested repeater
I receive **ArgumentNullException**, when I try to use `Where()` function in nested repeater Is this a bug or it is not supported?

Tested on: 3.2.1 and 4.0.0-preview01-final

```
        <dot:Repeater DataSource="{value: Parents}">
            <ItemTemaplete>
                Parent: 
                {{value: Id}}
                Children: 
                <dot:Repeater DataSource="{value: Children.Where(x => x.Id > 5)}">
                    <ItemTemplate>
                        {{value: Id}}
                    </ItemTemplate>
                </dot:Repeater>
            </ItemTemaplete>
        </dot:Repeater>
```

exyi:
We don't do null propagation on method arguments, but extension method's
first argument is essentially a target, so we'll do null propagation on that

resolves  #1185

### DIFF
--- a/src/Framework/Framework/Compilation/Binding/ExpressionNullPropagationVisitor.cs
+++ b/src/Framework/Framework/Compilation/Binding/ExpressionNullPropagationVisitor.cs
@@ -5,6 +5,7 @@ using System.Reflection;
 using DotVVM.Framework.Utils;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


quigamdev (approved):
👍

PR #1204, opened 2021-11-03
Utilize RecordExceptions

exyi:
Add common interface IDotvvmException for exception with some
DotVVM context information.
\+ a record base class implementing information

Most custom exceptions are now switched to using this base class,
which eliminates needless exception nesting
(for example missing resource in DotvvmControlException),
since the contextual information about current control will be placed into
 the existing exception.

Record exception, in general, make it easy to store the information from
the error message in a program-readable format - this is useful
for catching those exceptions and also for debugging, since you can
explore the exception properties in debugger or the error page.

The goal isn't to migrate everything to this, but it would be nice
to gradualy migrate the cases when it makes sense:
* exceptions that occur at runtime and users might want to handle them
* exceptions which should contain some dotvvm context (position in dothtml, current binding)

For example, here you can expand the exception and see which resource is causing a problem:

image


quigamdev (approved):
Very nice.

PR #1206, opened 2021-11-05
Add HierarchyRepeater

cafour:
There is currently no control in the framework able to render recursive hierarchies, so I backported `HierarchyRepeater` from business pack over here.

Implementing any `ItemsControl`, let alone a repeater, is a mine field, so any feedback is very, very welcome.


quigamdev (changes requested):
We need to rewrite this code to use the standard Item Template mechanism as Repeater uses. 
Tbh we should completely rewrite this control.

cafour:
@quigamdev

> We need to rewrite this code to use the standard Item Template mechanism as Repeater uses.

What do you mean by standard item template mechanism? This control already uses the same mechanism as Repeater with the exception of the `ResourceManager.AddTemplateResource` method.

This method cannot be used since it generates the template id based from the hash of template's contents... which needs to contain the template id. This can be overcome in two ways: either by rendering the template twice -- once to obtain the id and once to put the id back in, or by generating the template id in some other way (as I did in this PR).

> Tbh we should completely rewrite this control.

This is already pretty much a complete rewrite of `bp:HierarchyRepeater`. I'd rather fix this implementation than start all over again.

exyi (approved):

src/Samples/Common/Views/ControlSamples/HierarchyRepeater/Basic.dothtml
```diff
@@ -0,0 +1,58 @@
...
+
+    <h2>No Item and Level Tags</h2>
+    <dot:HierarchyRepeater DataSource="{value: Roots}"
+                           ItemChildrenBinding="{value: Children}"
+                           RenderSettings.Mode="Client">
+        <p title={value: _root.GlobalLabel}>{{value: Name}}</p>
+    </dot:HierarchyRepeater>
+
+    <h2>Empty</h2>
+    <dot:HierarchyRepeater DataSource="{value: Empty}"
```
quigamdev, line 40: It would be nice to have a test without ItemTagName and LevelTagName to test that it also works with the placeholder

quigamdev (approved):
It looks good to me. I have found just minor issues. 
It would be nice to have also some perf test.

quigamdev (changes requested):
I have just noticed, that unit tests are failing.

src/Samples/Common/Views/ControlSamples/HierarchyRepeater/Basic.dothtml
quigamdev: Oh, yeah, sorry, my bad.

src/Framework/Framework/Controls/HierarchyRepeater.cs
```diff
@@ -0,0 +1,443 @@
...
+            ImmutableArray<int> parentPath,
+            int index)
+        {
+            DotvvmControl itemWrapper = string.IsNullOrEmpty(ItemTagName)
+                ? new PlaceHolder()
+                : new HtmlGenericControl(ItemTagName);
+            var dataItem = new DataItemContainer { DataItemIndex = index };
+            itemWrapper.Children.Add(dataItem);
+            dataItem.SetDataContextTypeFromDataSource(GetDataSourceBinding());
+            {
```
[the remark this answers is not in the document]
cafour: All of them should be removed now.

cafour:
Limitation: `ItemWrapperCapability` cannot have bindings with a data context set to the item. This limitation is cause by the need to have consistent date context nesting both in client-rendering and server-rendering.

quigamdev (approved):

PR #1207, opened 2021-11-05
Use http security headers by default

exyi:
Switches were added to DotvvmSecurityConfiguration, so the behavior
can be configured.

* X-Frame-Options: DENY is used by default. It should pose no problem,
 since DotVVM did not work with iframes anyway due to cookie
  SameSite policy. When CrossSite frames are enabled, the cookie will
  have SameSite=None
* we also check Sec-Fetch-Dest - if it's an iframe, we validate on server that iframes are allowed.
   This mostly allows us to print a nicer error message, but may also in theory prevent some iframe timing attacks
* X-XSS-Protection: 1; mode=block
   probably not super useful, but also not harmful
* X-Content-Type-Options: nosniff
   - disable inference of content-type based on content, this could prevent some XSS attacks.
   probably also not super useful, but very unlikely to cause any problems.

* When on https, we set the session cookie with __Host- prefix. This prevents
 it being used by subdomains. Can help only in obscure cases -
 but for example when a system on the subdomain is compromised,
 the attacked can not pivot to the parent domain so easily.

We also check Sec-Fetch-* headers - that tells us what the browsers intends
to do with the page and if it's cross-origin or same-origin request.
Basically, we don't allow cross-origin POST and SPA requests. We also
don't allow JS initiated GET requests to dotvvm pages, but this may
be disabled, as someone could rely on it.

If the browser does not send these Sec-Fetch-* headers, we don't check
anything. You can enable strict checking by Security.RequireSecFetchHeaders option.
By default it's enabled on compilation page to prevent SSRF
and it does not matter too much if it does not work for someone.

Importantly, we don't set Strict-Transport-Security, since it's quite hard
to undo if it's unwanted and should be set on all requests coming from the server,
not only from requests handled by dotvvm.

I basically went through this guide and implemented what makes sense in our context:
https://infosec.mozilla.org/guidelines/web_security#web-security-cheat-sheet

### DIFF
--- a/src/Framework/Framework/Hosting/DotvvmPresenter.cs
+++ b/src/Framework/Framework/Hosting/DotvvmPresenter.cs
@@ -52,6 +52,7 @@ namespace DotVVM.Framework.Hosting
             StaticCommandServiceLoader = staticCommandServiceLoader;
 #pragma warning restore CS0618
             ApplicationPath = configuration.ApplicationPhysicalPath;
+            SecurityConfiguration = configuration.Security;
         }
 
         public IDotvvmViewBuilder DotvvmViewBuilder { get; }
@@ -65,6 +66,8 @@ namespace DotVVM.Framework.Hosting
         public ICsrfProtector CsrfProtector { get; }
 
         public IViewModelParameterBinder ViewModelParameterBinder { get; }
+
+        public DotvvmSecurityConfiguration SecurityConfiguration { get; }
 
 #pragma warning disable CS0618
         [Obsolete(DefaultStaticCommandServiceLoader.DeprecationNotice)]
@@ -116,6 +119,8 @@ namespace DotVVM.Framework.Hosting
             {
                 await context.InterruptRequestAsMethodNotAllowedAsync();
             }
+
+            await ValidateSecFetchHeaders(context);
 
             var requestTracer = context.Services.GetRequiredService<AggregateRequestTracer>();
             if (context.HttpContext.Request.Headers["X-PostbackType"] == "StaticCommand")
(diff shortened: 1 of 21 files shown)


src/Framework/Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
```diff
@@ -105,5 +107,27 @@ public async Task<bool> Handle(IDotvvmRequestContext context)
 
             return true;
         }
+
+        public static void WriteSecurityHeaders(IDotvvmRequestContext context)
+        {
+            var config = context.Configuration.Security;
+            var route = context.Route?.RouteName;
+            if (config.ContentTypeOptionsHeader.IsEnabledForRoute(route))
+                context.HttpContext.Response.Headers.Add("X-Content-Type-Options", new [] { "nosniff" });
```
quigamdev, line 116: We should add these headers only when they are not already set.

src/Framework/Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -442,6 +447,72 @@ public async Task ProcessStaticCommandRequest(IDotvvmRequestContext context)
...
+                else
+                {
+                    if (site == "same-origin")
+                        await context.RejectRequest($"Same site iframe are disabled in this application. If you are the developer, you can enable iframes by setting DotvvmConfiguration.Security.FrameOptionsSameOrigin.EnableForRoute(\"{route}\")");
+                    else
+                        await context.RejectRequest($"Cross site iframe are disabled in this application. If you are the developer, you can enable cross-site iframes by setting DotvvmConfiguration.Security.FrameOptionsCrossOrigin.EnableForRoute(\"{route}\"). Note that it's not recommended to enable cross-site iframes for sites / pages where security is important (due to Clickjacking)");
+                }
+            }
+
+            if (!checksAllowed || string.IsNullOrEmpty(dest))
```
quigamdev, line 485: This should check also for 'site'.

quigamdev (changes requested):

quigamdev (approved):

PR #1215, opened 2021-11-10
Fix a bug in finding ElementAt method on net6

exyi:
probably, I'm not able to replicate it atm


quigamdev:
I am gonna merge the main into this PR. The PR seems ok, just do not want to approve it before tests go green.

quigamdev (approved):
I have checked the tests and all passed. There is a problem with old unstable selenium drivers. But that is an unrelated issue.

PR #1217, opened 2021-11-10
markup controls: Don't render unused properties

exyi:
Another crack at reducing the amount of bullshit HTML we produce...
Just added a visitor which goes through all client-side bindings
and checks which properties are actually used.
Then, when rendering a markup control, only those will be included.

It also reduces the possibility of invoking arbitrary commands bound to
the markup control.
And it will stop exposing objects which may not be intended for the user.

It is technically a breaking change for people accessing the $control field
manually, but I hope there is not that many of them.


quigamdev (approved):
Hopefully, no one uses ko.contextFor on markup control and accesses some control's property directly from JS. 
If someone needs all props, we can add explicit option for rendering all props.

PR #1220, opened 2021-11-11
HtmlWriter will not throw exception about attributes when only whitespace is written

exyi:


src/Tests/ControlTests/testoutputs/SimpleControlTests.Decorator.html
```diff
@@ -0,0 +1,10 @@
+<html>
+	<head></head>
+	<body>
+		<div class="c1"></div>
+		<div class="c2"></div>
+
```
cafour, line 6: Everything seems to be fine, but just to make sure I understand what's going on here: the empty line between `.c2` and the html comment is caused by a whitespace-only RawLiteral generated from the newline after the dothtml comment? This didn't work before?

cafour (approved):

src/Tests/ControlTests/testoutputs/SimpleControlTests.Decorator.html
exyi: This only didn't work before in v4, after we added this validation to HtmlWriter. #1189

PR #1223, opened 2021-11-12
Add net6.0 target and hot reload handler that flushes various caches

exyi:
It's by no means bullet-proof, but it's most likely going to fail in .NET
sooner than in DotVVM, so we are fine.

It enables adding properties to viewmodel and then using them in
a dothtml page. Everything else I tried lead to .NET saying "rude edit"

### DIFF
--- a/src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMapper.cs
+++ b/src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMapper.cs
@@ -7,6 +7,7 @@ using Newtonsoft.Json;
 using System.Collections.Concurrent;
 using DotVVM.Framework.Configuration;
 using DotVVM.Framework.Utils;
+using DotVVM.Framework.Runtime;
 
 namespace DotVVM.Framework.ViewModel.Serialization
 {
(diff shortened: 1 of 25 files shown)


src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMapper.cs
```diff
@@ -120,5 +123,33 @@ private static bool IsSetterSupported(PropertyInfo property)
...
+
+            var currentProperties = cachedItem.Properties.Select(p => (p.Name, p.Type, p.BindDirection, p.ViewModelProtection)).ToHashSet();
+            if (currentProperties.SetEquals(cachedItem.OriginalProperties))
+            {
+                this.serializationMapCache.TryRemove(hash, out _);
+                return true;
+            }
+
+            Console.WriteLine(string.Join(" ", freshProperties));
+            Console.WriteLine(string.Join(" ", currentProperties));
```
exyi, line 151: this is not exactly necessary

src/Samples/Common/ViewModels/DefaultViewModel.cs
```diff
@@ -13,6 +13,7 @@ public class DefaultViewModel : SamplesViewModel
     {
         public string Title { get; set; }
         public List<RouteData> Routes { get; set; }
+        // public string Name { get; set; }
```
exyi, line 16: remove

tomasherceg (approved):

PR #1224, opened 2021-11-14
Improve startup and compilation performance slightly

exyi:
Added reference to FastExpressionCompiler - we'll use that for serialization, capabilities. Unfortunately, we can't use it for bindings at the moment, because there are some bugs.

We used to look for extension method even without `_this` qualifier, so this `{value: Where(x => ...)}` probably worked and now it won't (because looking for extension methods is pretty expensive). In C# this is also disabled, so I think it makes sense to have this limitation.

Other things are just optimizations of existing methods that showed up as hot during profiling startup and compilation.


quigamdev (approved):
Very nice 👍

PR #1225, opened 2021-11-16
Rewriting the view compiller to use System.Linq.Expressions

Mylan719:
Presently we use Roslyn in the DotVVM view compiler to use generate control tree from the markup:

1. Runtime types and references are translated into a Roslyn C# syntax tree based on view markup
2. Roslyn compilation is created from the C# syntax tree
3. All required assebmlies are loaded and referenced into the compilation.
4. The Roslyn compilation is compiled and loaded into the runtime
5. Generated view builder is found in the generated assembly 
6. The method that builds the control tree is finaly executed

This way of doing things has several disadvantages:

- It's unnecesarily complex
- Has performance penalties
- Limits implementation of new features, since it is hard to extend or modify
- No simple way to use runtime generated types

We decided to rewrite the View compilation to using the `System.Linq.Expressions`. This makes a much cleaner straitforward implementation of the control tree building. Each view is constusted using expressions and runtime metadata, with no need for additional conversions. Blocks of expresions that build the view are encapsulated into lambdas that compile into delegates that can be passed directly to template builder, or executed to create the root control. We don't need to specificaly load any assemblies, until they are needed. Runtime generated types are no problem.


src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
```diff
@@ -699,7 +695,12 @@ protected virtual void ResolveControlContentImmediately(IAbstractControl control
             foreach (var node in content)
             {
                 var child = ProcessNode(control, node, control.Metadata, dataContext);
-                if (child != null) treeBuilder.AddChildControl(control, child);
+                if (child != null)
+                {
+                    treeBuilder.AddChildControl(control, child);
+                    if (!child.Metadata.Type.IsAssignableTo(ResolvedTypeDescriptor.Create(typeof(DotvvmControl))))
+                        node.AddError($"Content control must inherit from DotvvmControl, but {child.Metadata.Type} doesn't.");
+                }
```
exyi, line 703: This has to be here, because this case was uncovered and one test tested that the C# compilation failed :facepalm:  So this fixes the underlying problem

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        private int CurrentControlIndex;
+
+        public const string ControlBuilderFactoryParameterName = "controlBuilderFactory";
+        public const string ServiceProviderParameterName = "services";
+        public const string BuildTemplateFunctionName = "BuildTemplate";
+
+        private Dictionary<GroupedDotvvmProperty, string> cachedGroupedDotvvmProperties = new Dictionary<GroupedDotvvmProperty, string>();
+        private ConcurrentDictionary<(Type obj, string argTypes), string> injectionFactoryCache = new ConcurrentDictionary<(Type obj, string argTypes), string>();
+        private readonly Stack<BlockInfo> blockStack = new();
```
exyi, line 24:
```suggestion
        private readonly Stack<BlockInfo> blockStack = new();
        private currentBlock => blockStack.Peek()
```

I think this could improve the code a bit.

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+namespace DotVVM.Framework.Compilation
+{
+    public class DefaultViewCompilerCodeEmitter
+    {
+        private static Type[] emptyTypeArguments = new Type[] { };
+
+        private int CurrentControlIndex;
+
+        public const string ControlBuilderFactoryParameterName = "controlBuilderFactory";
+        public const string ServiceProviderParameterName = "services";
```
exyi, line 19:
```suggestion
        public readonly ParameterExpression ServiceProviderParameter = Expression.Parameter("services", typeof(IServiceProvider));
```

With Linq Expressions, you can do this ;) (and the same for the ControlBuilderFactoryParameterName). The same expression can be reused in multiple runs without problems and no need to lookup it by string in some dictionary.

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+            return Expression.NewArrayInit(elementType, convertedValues);
+        }
+
+        /// <summary>
+        /// Emits the control builder invocation.
+        /// </summary>
+        public ParameterExpression EmitInvokeControlBuilder(Type controlType, string virtualPath)
+        {
+            var builderName = "c" + CurrentControlIndex + "_builder";
+            var untypedName = "c" + CurrentControlIndex + "_untyped";
```
exyi, line 104:
```suggestion
```

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+        public void EmitSetProperty(string controlName, string propertyName, Expression valueExpression)
+        {
+            //[controlName].[propertyName] = [value] 
+            var controlParameter = GetParameterOrVariable(controlName);
+            var assigment = Expression.Assign(Expression.PropertyOrField(controlParameter, propertyName), valueExpression);
+
+            blockStack.Peek().Expressions.Add(assigment);
+        }
+
+        public Expression CreateDotvvmPropertyIdentifier(DotvvmProperty property)
```
exyi, line 141: I think we can inline this method (and other very simple ones)

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+        /// </summary>
+        public void PushNewMethod(string name, Type returnType, params ParameterExpression[] parameters)
+        {
+            blockStack.Push(new BlockInfo(name, returnType, parameters));
+        }
+
+        /// <summary>
+        /// Pops the method.
+        /// </summary>
+        public Delegate PopMethod()
```
exyi, line 330:
```suggestion
        public T PopMethod<T>() where T: Delegate
```

The type of the lambda that we want is know at both places where this is invoked, so I think we could make it like this

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
```diff
@@ -0,0 +1,360 @@
...
+        {
+            var blockInfo = blockStack.Pop();
+            var block = Expression.Block(blockInfo.Variables.Values, blockInfo.Expressions);
+
+            var lambda = Expression.Lambda(block, blockInfo.Parameters.Values);
```
exyi, line 335:
```suggestion
            var lambda = Expression.Lambda<T>(block, blockInfo.Parameters.Values);
```

src/Framework/Framework/Compilation/ViewCompiler/ViewCompilingVisitor.cs
```diff
@@ -19,48 +16,44 @@ namespace DotVVM.Framework.Compilation
...
-            var createsCustomDerivedType = view.Metadata.Type == typeof(DotvvmView);
+            var isPageView = view.Metadata.Type == typeof(DotvvmView);
 
-            if (createsCustomDerivedType)
+            if (isPageView)
             {
-                var resultControlType = className + "Control";
-                emitter.ResultControlTypeSyntax = SyntaxFactory.ParseTypeName(resultControlType);
-                emitter.EmitControlClass(view.Metadata.Type, resultControlType);
+                //TODO: Possibly emit custom generated type here
```
exyi, line 43: nah, I don't think that's good for anything

src/Framework/Framework/Compilation/ViewCompiler/ViewCompilingVisitor.cs
```diff
@@ -264,20 +256,19 @@ protected string CreateControl(ResolvedControl control)
...
         /// Processes the template.
         /// </summary>
-        protected string CreateTemplate(string builderMethodName)
+        protected string CreateTemplate(string builderMethodName, Delegate compiledDelegate)
         {
-            var templateName = emitter.EmitCreateObject(typeof(DelegateTemplate));
-            emitter.EmitSetProperty(templateName,
-                nameof(DelegateTemplate.BuildContentBody),
-                emitter.EmitIdentifier(builderMethodName));
+            var templateName = emitter.EmitCreateObject(typeof(DelegateTemplate)).Name;
```
exyi, line 269:
```suggestion
            var templateName = emitter.EmitValue(new DelegateTemplate { BuildContentBody = compiledDelegate });
```

I think we can even do something like this. No need to compile it and then execute it all over again.

quigamdev:
I love the idea. I still need to check the PR, but the idea is brilliant. We definitely gonna include this in 4.0 release!

src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
```diff
@@ -105,14 +104,6 @@ void editCompilationException(DotvvmCompilationException ex)
...
 
-        /// <summary>
-        /// Gets the name of the class from the file name.
-        /// </summary>
-        public static string GetClassFromFileName(string fileName)
-        {
-            return GetValidIdentifier(Path.GetFileNameWithoutExtension(fileName));
-        }
-
         protected static string GetValidIdentifier(string identifier)
```
quigamdev, line 107: Is this method still in use?

quigamdev (approved):
image

src/Framework/Framework/Compilation/ViewCompiler/ViewCompilingVisitor.cs
Mylan719: Solved, compressed the method too

src/Framework/Framework/Compilation/ViewCompiler/ViewCompilingVisitor.cs
Mylan719: Removed

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
Mylan719: Solved

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
Mylan719: Solved

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
Mylan719: Removed

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
Mylan719: I dont like this solution very much as "services" it is passed as an argument to the block, and therefore is added on block creation. This provides certain degree compile time checking. Also I find it more systematic.

src/Framework/Framework/Compilation/ViewCompiler/DefaultViewCompilerCodeEmitter.cs
Mylan719: Removed, replaced by `EmitValue`. The methot also no longer creates any identifier so it was at best unhelpful. ;)

src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
Mylan719: Nope, removed

exyi (approved):
:+1:

PR #1226, opened 2021-11-16
Make it possible to access entire capability in server-side styles

exyi:
For example, this makes it possible to write a common function for a style
and for a control.

Potentially, this infrastucture could be used to execute
entire composite controls at compile-time. This makes is a bit more
limiting, since it's only executed once, but if you wish
to create data bindings in the control, it's the only option
that will not make the page super slow just by putting
the control in a Repeater

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedPropertyCapability.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedPropertyCapability.cs
@@ -91,14 +91,14 @@ namespace DotVVM.Framework.Compilation.ControlTree.Resolved
                 var properties = propertyGroupLookup[pgroup].ToArray();
 
                 var propertyOriginalValue = prop.GetValue(obj);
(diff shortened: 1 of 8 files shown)


src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedPropertyCapability.cs
```diff
@@ -0,0 +1,121 @@
...
+                    .ToLookup(gp => gp.PropertyGroup);
+
+            foreach (var (prop, pgroup) in groupMappingList)
+            {
+                var properties = propertyGroupLookup[pgroup].ToArray();
+
+                var propertyOriginalValue = prop.GetValue(obj);
+                var dictionary = (System.Collections.IDictionary)(
+                    propertyOriginalValue ??
+                    Activator.CreateInstance(prop.PropertyType)
```
exyi, line 96: This will break if propertytype is IDictionary

tomasherceg (approved):

PR #1233, opened 2021-12-01
ValidationSummary.IncludeErrorsFromTarget set to true by default

exyi:


tomasherceg:
It should be `IncludeValidationErrorsFromTarget`, not children.

tomasherceg (approved):

PR #1234, opened 2021-12-03
Fix continuous integration (GitHub Actions)

cafour:
Fixes and extends our CI pipeline on GitHub.

**Jobs:**

_Build all projects without errors:_
Builds truly **all** projects and fails if compilation fails. This should prevent us pro merging commits that mess up compilation of our more peripheral projects (or those that only run on Windows). My reasoning behind this addition is that we don't want people to come to this repo and clone it only to find out that the solution cannot be compiled.

_Build published projects without warnings:_
Same job as what _build without warnings_ used to be, but builds all projects that we publish on NuGet. They are listed individually in the action, so that it's obvious which failed the job.

_.NET unit tests:_
Runs `src/Tests` and `src/Analyzers/Analyzers.Tests` and creates reports if there are any failed tests.

_JS unit tests:_
No changes were made to this job. Still runs our JS tests.

_UI tests:_
Runs UI tests on both Chrome and Firefox and both ASP.NET Core and OWIN. Creates reports if there are any failed tests.

**Composite actions**

* `.github/setup` -- sets up .NET and Node, restores and caches packages, and adds a `Build-Dotvvm` script that serves as shortcut for
  ```
  "msbuild.exe" -t:Clean,Build -v:m -warnAsError -p:Configuration=$configuration -p:TargetFramework=$targetFramework "$project"
  ```
* `.github/uitest` -- runs our UI tests on a given browser and publishes the results. If the runner's OS is Windows, runs the OWIN version.
* `.github/unittest` -- runs a given test project and publishes the results.
* `test-report` -- reports **failed** tests found in a `.trx` file using GitHub's [checks API](https://docs.github.com/en/rest/reference/checks). Reports nothing if all tests passed or skipped.

**Other changes**

* Replaces npm with yarn during CI.
* Converts old-style project to the new csproj style with [MSBuild.SDK.SystemWeb](https://github.com/CZEMacLeod/MSBuild.SDK.SystemWeb).
* Regenerates assembly redirects automatically (hence the changes to `web.config`s).
* Removes configuration for our **public** Azure Pipelines CI.

**Outstanding issues**

* Tests running in Firefox on Windows are insanely slow. This seems to be an issue with SeleniumUtils.
* There still isn't a deployment pipeline on GA
* Currently both .NET 5 and .NET 6 need to be set up. #1272 seems to fix this.

(These issues are IMHO beyond the scope of this PR.)

Please squash when merging.


exyi:
> Builds truly all projects during Build without warnings.

I'm quite not sure we want this, I only included project which we care about (those where we produce libraries), building all projects seems unnecessarily annoying to me.
Also we pretty much have to choose one target framework against which we want to build a specific package, your solution will IMHO build them for all targets. This is a problem, because in some cases it seems impossible to get it build without warnings on both netstandard2.1 and net6.0 (the IReadOnlyDictionary thing)

Moreover, this step used to indicate clearly which project failed to build, now it just says "MSBuild build" and you have to go through the logs... Could you please revert this part roughly to the previous version? I understand it's needlesly noisy to list all projects in the yaml, but it has some advantages... (and sln files are arguably even noisier :D)

exyi:
Some samples seem to fail because rollup is not invoked in the Samples/Common project. I guess we should add it to the csproj too, could you do that too?

.github/unittest/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    required: true
+  configuration:
+    default: Debug
+    description: The build configuration
+    required: false
+
+runs:
+  using: composite
+  steps:
+  - run: dotnet msbuild "${{ inputs.project }}"
```
exyi, line 17: actually what's the difference between dotnet build and dotnet msbuild in this case?

.github/unittest/action.yml
```diff
@@ -0,0 +1,26 @@
...
+      -t:Rebuild
+      -p:WarningLevel=0
+      -p:Configuration=${{ inputs.configuration }}
+    shell: bash
+  - run: dotnet test "${{ inputs.project }}"
+      --no-build
+      --logger "GitHubActions;report-warnings=true"
+      --configuration "${{ inputs.configuration }}"
+    shell: bash
```
exyi, line 26: and honesly not sure if we need a special action.yml for unit tests when it's just a matter of running `dotnet test`...

ci/windows/run.ps1
```diff
@@ -32,6 +32,9 @@ $testResultsDir = "$Root\artifacts\test\"
 $samplesDir = "$Root\src\Samples\Tests\Tests\"
 $ciDir = "$Root\ci\windows\"
 
+# set the codepage to UTF-8
+chcp 65001
```
exyi, line 36: actually, is this file used now? it seems to use uitest.ps1

exyi (approved):

ci/windows/run.ps1
cafour: Nope. It has been deleted.

.github/uitest/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      -TrxName "ui-test-results-${{ inputs.browser }}-${{ runner.os }}.trx"
+    shell: pwsh
+
+  # publish the result to github
+  - name: Parse test results
+    uses: NasAmin/trx-parser@v0.2.0
+    id: trx-parser
+    with:
+      TRX_PATH: ${{ github.workspace }}/artifacts/test
+      REPO_TOKEN: ${{ inputs.token }}
```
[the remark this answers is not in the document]
cafour: This little maneuver cost this PR another ~100 `CI` commits but it works now. :D

.github/unittest/action.yml
cafour: Now that we create a report for unit tests as well and also run Analyzer tests, I think it's worth separating into its own action.

.github/unittest/action.yml
cafour: None. Fixed. :D

.github/setup/action.yml
```diff
@@ -0,0 +1,53 @@
...
+  - run: |
+      corepack enable
+      yarn set version stable
+    shell: bash
+
+  # .NET-related
+  - uses: actions/setup-dotnet@v1
+    with:
+      dotnet-version: |
+        5.0.403
```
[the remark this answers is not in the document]
cafour: Jup, at least until #1272 gets merged.

cafour:
@exyi I tried to fix everything you mentioned (see updated PR description above). Some samples still fail, despite `rollup` being invoked via `package.json`, but this is IMHO out of scope for this PR (and I just really, really, really want to merge it now :D).

I'm squashing and merging this, since it's already approved.

PR #1237, opened 2021-12-12
Remove redundant dependencies on DotNet.Metaprogramming and CodeDom

exyi:
CodeDom was needed because of one last T4 template, so I rewrote the template.

We wanted Metaprogramming for async/await in expressions, but
that old version does not support that and
never versions don't run on .NET framework :/


src/Samples/Common/Views/ControlSamples/TemplateHost/CompositeControlWithTemplate.cs
```diff
@@ -13,6 +13,7 @@ public class CompositeControlWithTemplate : CompositeControl
             ITemplate contentTemplate
         )
         {
+            throw null;
```
tomasherceg, line 16:
```suggestion

```

tomasherceg (approved):

PR #1242, opened 2021-12-16
Put entire control hierarchy into the error page

exyi:
This comes in quite useful when debugging complex control
hierarchies created in code. This implementation
looks a bit crude, but it works...

image


tomasherceg:
I've tweaked CSS a bit so the section looks nicer, and added UI tests (they are UI tests because I wanted to see how the page will look like).

image

tomasherceg (approved):

exyi:
Nice, this looks significantly nicer :)

PR #1244, opened 2021-12-20
Fixed and improved serializability analyzers

acizmarik:
This PR introduces a couple of fixes and improvements to serializability analyzers.

- Fixed **bug**: nullables were sometimes incorrectly marked as not serializable.
- Fixed **bug**: certain supported types were incorrectly marked as not serializable (`GridViewDataSet<T>`, `UploadedFilesCollection` and `BusinessPackDataSet<T>`). They do not follow our general rules, however their serialization is well-defined and should not issue warnings.
- Fixed **bug**: that caused that some user-defined generic reference types were incorrectly marked as not serializable
- New **improvement**: symbols obtained from a specific `Compilation` within a specific `SemanticModel` are no longer cached indefinitely. Instead, we now construct cache for each `Compilation`. These caches are held using weak references.
- New **improvement**: sanity check. Serializability analyzers perform a simple sanity check to ensure that everything works before issuing warnings. (This might not be necessary anymore but I decided to leave it there).
- New **improvement**: serializability warnings now issue property paths. This is more straight-forward when looking for the root of the issue - especially when working with multiple levels of nested classes.

Additionally, `DateTimeOffset` was added to supported types. Analyzers no longer generate serializability warnings for properties that implement `IEnumerable` (however element type needs to be still serializable).

### DIFF
--- a/src/Analyzers/Analyzers/Serializability/TypeSymbolExtensions.cs
+++ b/src/Analyzers/Analyzers/Serializability/TypeSymbolExtensions.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Text;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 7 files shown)


acizmarik:
It seems like Analyzer tests are not run on CI currently. I will write some more tests and fix pipelines as well before merging.

src/Analyzers/Analyzers/Serializability/TypeSymbolExtensions.cs
```diff
@@ -10,143 +11,154 @@ namespace DotVVM.Analyzers.Serializability
...
 
             // Supported reference types (excluding user-defined types)
-            referenceTypeBuilder.Add(compilation.GetSpecialType(SpecialType.System_Object));
-            referenceTypeBuilder.Add(compilation.GetSpecialType(SpecialType.System_String));
-            referenceTypeBuilder.Add(compilation.GetTypeByMetadataName("System.Collections.Generic.List`1")!);
-            referenceTypeBuilder.Add(compilation.GetTypeByMetadataName("System.Collections.Generic.Dictionary`2")!);
+            cacheBuilder.Add(compilation.GetSpecialType(SpecialType.System_Object));
+            cacheBuilder.Add(compilation.GetSpecialType(SpecialType.System_String));
+            cacheBuilder.Add(compilation.GetTypeByMetadataName("System.Collections.Generic.List`1")!);
+            cacheBuilder.Add(compilation.GetTypeByMetadataName("System.Collections.Generic.Dictionary`2")!);
```
exyi, line 132: We should allow anything implementing IEnumerable

src/Analyzers/Analyzers/Serializability/TypeSymbolExtensions.cs
```diff
@@ -10,143 +11,154 @@ namespace DotVVM.Analyzers.Serializability
...
+            cacheBuilder.Add(compilation.GetSpecialType(SpecialType.System_Decimal));
+            cacheBuilder.Add(compilation.GetSpecialType(SpecialType.System_Char));
 
             // Special (whitelisted) value types
-            valueTypeBuilder.Add(compilation.GetSpecialType(SpecialType.System_DateTime));
-            valueTypeBuilder.Add(compilation.GetTypeByMetadataName("System.Guid")!);
-            valueTypeBuilder.Add(compilation.GetTypeByMetadataName("System.TimeSpan")!);
+            cacheBuilder.Add(compilation.GetSpecialType(SpecialType.System_DateTime));
+            cacheBuilder.Add(compilation.GetTypeByMetadataName("System.Guid")!);
+            cacheBuilder.Add(compilation.GetTypeByMetadataName("System.TimeSpan")!);
```
exyi, line 159: DateTimeOffset should also be (mostly) fine

exyi (approved):

PR #1252, opened 2022-01-03
Fixed issue with incorrect token `StartPosition` for interpolated expressions

acizmarik:
This PR fixes an issue that breaks the WIP string interpolation feature in VS Extension. 

Whenever an interpolated expression is parsed, we create a new instance of `BindingTokenizer` and `BindingParser`. However, we should also set an appropriate start indices for these expressions.

This is the behaviour I want to achieve:
image

However, right now it can desynchronize into something like this:
image
(Note that all interpolations are starting on the first index - which is the data currently generated by framework)


exyi:
Are you sure this sets the StartIndex even in child expressions? Could you please also test this `$"{ "x" + "y" + "z" }"`?

acizmarik:
Thanks for pointing that out! At first I though that having them relative to each expression is fine but it would probably make a mess somewhere else.

Anyway even child expressions should now have correct `StartPosition`s - that is absolute from the binding beginning.

acizmarik:
I am thinking that we could also remove the `Tokens` property from the `ParserBase` and just pass the `BindingTokenizer` to the parser directly. Since this is going to 4.0 we could make another small breaking change right? What do you think @exyi?

Also to me it does not sound like a bad idea from the design perspective. As it is quite common for parsers to directly interact with tokenizers (operations such as `GetNextToken` and so on).

exyi (approved):
Nice :)

PR #1260, opened 2022-01-07
Prototype of JsComponent-Svelte integration

exyi:


exyi:
It works well, we just need to make package out of it. Compared to the react bridge, it does not support templates (yet), but it does support 2-way binding (see the incrementer sample)

quigamdev (approved):
It would be nice if it supports templates. Otherwise, I haven't found any issues. Nice job. 👍

PR #1264, opened 2022-01-16
Serialize records correctly

exyi:
the serializer now supports injecting parameters into the constructor.

Instead of writing data directly into the properties, we
now first deserialize all properties into local variables
and then either call setters or use them as constructor parameters.

### DIFF
--- a/src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMap.cs
+++ b/src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMap.cs
@@ -9,6 +9,8 @@ using DotVVM.Framework.Configuration;
 using FastExpressionCompiler;
 using static System.Linq.Expressions.Expression;
 using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace DotVVM.Framework.ViewModel.Serialization
 {
@@ -70,6 +72,7 @@ namespace DotVVM.Framework.ViewModel.Serialization
         /// </summary>
         private Expression CallConstructor(Expression services, Expression[] properties)
         {
+            Debug.Assert(properties.Select(p => p.Type).SequenceEqual(Properties.Select(p => p.Type)));
             if (constructorFactory != null)
                 return Convert(Invoke(Constant(constructorFactory), services), Type);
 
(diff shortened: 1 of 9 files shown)


src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMap.cs
```diff
@@ -59,20 +64,45 @@ public void ResetFunctions()
...
+                throw new Exception($"Can not deserialize {Type.FullName} because it's abstact. Please avoid using abstract types in view model. If you really mean it, you can add a static factory method and mark it with [JsonConstructor] attribute.");
+
+            if (Constructor is null)
+                throw new Exception($"Can not deserialize {Type.FullName}, no constructor or multiple constructors found. Use the [JsonConstructor] attribute to specify the constructor used for deserialization.");
+
+            var parameters = Constructor.GetParameters().Select(p => {
+                var pIndex = Properties.FindIndex(pp => pp.ConstructorParameter == p);
+
+                if (pIndex < 0)
+                    return Call(typeof(ServiceProviderServiceExtensions), "GetRequiredService", new [] { p.ParameterType }, services);
```
exyi, line 95: Wrap the exception into something more specific that "service X does not exist"

src/Framework/Framework/ViewModel/Serialization/ViewModelSerializationMap.cs
```diff
@@ -246,16 +301,30 @@ public ReaderDelegate CreateReaderFactory()
...
             // build the lambda expression
             var ex = Expression.Lambda<ReaderDelegate>(
-                Expression.Convert(
-                    Expression.Block(Type, new[] { value, currentProperty, readerTmp }, block),
-                    typeof(object)).OptimizeConstants(),
-                reader, serializer, valueParam, encryptedValuesReader);
-            return ex.CompileFast(flags: CompilerFlags.ThrowOnNotSupportedExpression);
+                Expression.Block(typeof(object), new[] { value, currentProperty, readerTmp }.Concat(propertyVars), block).OptimizeConstants(),
+                reader, serializer, valueParam, encryptedValuesReader, servicesParameter);
+            return ex.CompileFast(flags: CompilerFlags.ThrowOnNotSupportedExpression | CompilerFlags.EnableDelegateDebugInfo);
```
quigamdev, line 327: Please remove `CompilerFlags.EnableDelegateDebugInfo` or leave it only for debug compilation.

quigamdev (approved):
Looks good to me👍

PR #1265, opened 2022-01-17
Added loaded assemblies tab into the Error Page

acizmarik:
This PR adds information about loaded assemblies participating in the execution at the time when an Error Page is generated. I believe that this information can be helpful when diagnosing exceptions from users, such as the following:

* **Missing package**: i.e. trying to use `HotReload` without referencing the proper packages and using framework < 4.0
* **Binary incompatibility**: i.e. `MissingMethodException` when user has incompatible framework and business pack versions
* **Culture issues**: i.e. using a concrete language (culture) that can potentially break some stuff when invoking culture-sensitive operations

image


exyi (approved):
Good idea.

PR #1267, opened 2022-01-20
Added property paths to JS validation errors

acizmarik:
This PR adds property paths into `ValidationError` type. It can be easily inspected in console to quickly determine which part of viewModel is being violated. Property path format is consistent with validation changes introduced in 4.0 (absolute paths from root vm and uses slashes instead of knockout expression syntax). See an example:

image

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
+++ b/src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
@@ -40,20 +40,19 @@ export function findPathToChildObject(vm: any, child: any, path: string): string
         return path;
     }
 
-    vm = ko.unwrap(vm);
     if (typeof vm !== "object" || vm == null) {
         return null;
     }
 
     if (Array.isArray(vm)) {
         // Iterate over its elements
-        vm.forEach(function (value, index) {
-            let result = findPathToChildObject(value, child, path + "/" + `[${index}]`)
+        let index = 0;
+        for (const value of vm) {
+            let result = findPathToChildObject(value, child, path + `/[${index}]`)
             if (result != null)
                 return result;
-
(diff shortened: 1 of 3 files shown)


src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -33,6 +34,44 @@ export function traverseContext(context: any, path: string): any {
...
+    if (typeof vm !== "object" || vm == null) {
+        return null;
+    }
+
+    if (Array.isArray(vm)) {
+        // Iterate over its elements
+        vm.forEach(function (value, index) {
+            let result = findPathToChildObject(value, child, path + "/" + `[${index}]`)
+            if (result != null)
+                return result;
```
exyi, line 53: This will not work, it returns from the inner anonymous function, not from the `findPathToChildObject` function. You need to use `for(const value of vm) {... }` and track the index manually

src/Framework/Framework/Resources/Scripts/utils/evaluator.ts
```diff
@@ -33,6 +34,44 @@ export function traverseContext(context: any, path: string): any {
     return currentLevel
 }
 
+export function findPathToChildObject(vm: any, child: any, path: string): string | null {
```
exyi, line 37: Also... I quite don't like that this function operates on knockout observables instead of the state objects. The obserables are created lazily, because they are quite costly, so this will essentially populate the entire viewmodel

exyi (changes requested):

acizmarik:
Thanks for the feedback. All mentioned issues should be fixed now.

exyi (approved):

PR #1271, opened 2022-01-31
Fixed serializability analysis issues

acizmarik:
This PR fixes a couple of issues that were found while testing analyzers on some internal projects. Most of the issues were connected with analysis of complex nested types.

* Fixed an unhandled exception that sometimes occurred when analyzing nullable reference types
* Only public and non-static members are considered when analyzing viewmodels
* Property attributes are always checked  (looking for `BindAttribute` and `JsonIgnoreAttribute`)
* Inner classes are not checked solely because they are defined within a viewmodel


acizmarik:
Tests pass locally. I also tested this on a couple of internal projects. Unit tests will be added in #1272

src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -74,65 +79,169 @@ private static void AnalyzeSyntaxTree(SemanticModelAnalysisContext context)
...
+
+        private static void AnalyzePropertyType(ITypeSymbol propertyType, Location location, string path, SerializabilityAnalysisContext context)
+        {
+            // Check if we previously resolved this type
+            if (context.IsVisited(propertyType))
+            {
+                ReportIfNotSerializable(propertyType, location, path, context);
+                return;
+            }
+            context.MarkAsVisited(propertyType);
```
exyi, line 134: Would be nice ignore these checks if the property or class has [JsonConverter] attribute.

src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -74,65 +79,169 @@ private static void AnalyzeSyntaxTree(SemanticModelAnalysisContext context)
...
+                        if (!context.IsVisited(underlyingTypeSymbol))
+                        {
+                            AnalyzePropertyType(underlyingTypeSymbol, location, path, context);
+                        }
+                        else
+                        {
+                            ReportIfNotSerializable(underlyingTypeSymbol, location, path, context);
+                        }
+                    }
+                    break;
```
exyi, line 192: I don't think we need this, the underlying type is always serializable anyway.

exyi (approved):

PR #1274, opened 2022-02-04
Add AddCssClass and AddCssStyle extension methods for ValueOrBinding

exyi:
Some not very nice code has to be used in Fluent UI and BS5 for this,
so it would be nice to have this helper.

In order for AddCssClass to work with any html control, not only
HtmlGenericControl, generic mechanism for handling property
groups from capabilies was introduced
 - see DotvvmCapabilityProperty.FindProperty
 - and see DotvvmBindableObjectHelper.CapabilityCache.cs

the FindProperty can lookup any property from a specified capability.
The CapabilityCache makes it reasonably fast specifically for
HtmlCapability

### DIFF
--- a/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
+++ b/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
@@ -130,6 +130,16 @@ namespace DotVVM.Framework.Binding
                 yield return x;
         }
 
+        /// <summary> Return a <see cref="DotvvmProperty" /> defined by this capability with the specified name. `null` is returned if there is no property with the name or if this capability does not have property mapping.
+        /// <code> HtmlGenericControl.HtmlCapabilityProperty.FindProperty("Visible") </code> </summary>
+        public DotvvmProperty? FindProperty(string name) =>
+            PropertyMapping?.FirstOrDefault(p => p.prop.Name == name).dotvvmProperty;
+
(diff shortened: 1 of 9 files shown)


src/Framework/Framework/Compilation/Styles/StyleMatchContextMethods.cs
```diff
@@ -344,6 +345,71 @@ public static IStyleMatchContext<T>[] ControlProperty<T>(this IStyleMatchContext
...
+        {
+            var value = s.GetValue();
+            if (value is IBinding binding)
+                return binding.GetProperty<ParsedExpressionBindingProperty>()?.Expression;
+        }
+        return null;
+    }
+
+    /// <summary> If the property has a binding and it's bound to view model property, returns its <see cref="PropertyInfo" />. Otherwise, returns null. </summary>
+    public static PropertyInfo? BindingProperty<TControl, TProp>(this IStyleMatchContext<TControl> c, Expression<Func<TControl, TProp>> pp)
```
tomasherceg, line 375: Rewrite to @acizmarik implementation in #1385

tomasherceg (commented):
Maybe we should dump the public API and make sure we have all overloads for all methods - hard-coded value, `ValueOrBinding<T>`, `ValueOrBinding<T>?`, `IValueBinding<T>` and `IValueBinding<T>?`.

acizmarik (approved):
Nice

PR #1277, opened 2022-02-04
Remove breaking change in HtmlLiteral

exyi:
We added too strict control validator to ConfigurableHtmlControl,
this commit only allows parts of it if it's certain
that the markup will produce a runtime error.


acizmarik (approved):

acizmarik:
Please fix the failing test

PR #1278, opened 2022-02-06
Fix markup control with properties and js directive

exyi:
allBindings is not a dictionary-object, but something with get/has methods


quigamdev (approved):
Thnaks 👍

PR #1281, opened 2022-02-06
Fixed DotvvmViewCompilationService when control has scoped dependency

quigamdev:
This PR fixes the resolution of scoped services in controls that has @baseType directive.


exyi (approved):
:+1:

PR #1286, opened 2022-02-08
Improved error reporting for compiled `BindingDelegate`

acizmarik:
This PR adds two checks for compiled `BindingDelegate`s. 

Previously, when data contexts were confused, weird exceptions were being thrown (mainly `IndexOutOfRangeException` and `InvalidCastException`). This fix changes the exceptions to be more descriptive.


acizmarik:
Since this generated code gets further processed by the NullPropagatingVisitor, in the end the code looks bad. We should probably redo this PR by injecting similar code at a later stage (after the null propagation is applied).

PR #1287, opened 2022-02-09
Adjusted validation API and added documentation

acizmarik:
This PR makes one more change to the validation API and introduces documentation comments for the API.

Right now users can easily end up in a weird situation. Previously, some documentation pages suggested adding validation errors similarly to this: `Context.AddModelError(nameof(MyProperty), "Something wrong with MyProperty")`. This however, uses the old way of providing validation paths. Therefore, right now such usage throws an exception that you are probably attempting to provide a knockout expression which is no longer supported.

In order to make it easier for users (i.e. discover this change during compilation rather than during runtime), I renamed the method to `AddRawModelError(string, string)`. Furthermore, I added documentation comments to better capture in the code what methods should be used under what circumstances.

### DIFF
--- a/src/Framework/Framework/ViewModel/Validation/DotvvmRequestContextValidationExtensions.cs
+++ b/src/Framework/Framework/ViewModel/Validation/DotvvmRequestContextValidationExtensions.cs
@@ -8,6 +8,10 @@ namespace DotVVM.Framework.ViewModel.Validation
 {
     public static class DotvvmRequestContextValidationExtensions
     {
+        /// <summary>
+        /// Adds a new validation error with the given message and attaches it to the root viewmodel.
+        /// </summary>
+        /// <param name="message">Validation error message</param>
         public static ViewModelValidationError AddModelError(this IDotvvmRequestContext context, string message)
         {
             var error = new ViewModelValidationError(message)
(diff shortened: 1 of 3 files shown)


src/Framework/Framework/ViewModel/Validation/DotvvmRequestContextValidationExtensions.cs
```diff
@@ -19,18 +23,32 @@ public static ViewModelValidationError AddModelError(this IDotvvmRequestContext
...
 
-        public static ViewModelValidationError AddModelError(this IDotvvmRequestContext context, string propertyPath, string message)
+        /// <summary>
+        /// Adds a new raw validation error. This method is intended only for advanced use-cases and it partially bypasses the validation framework. 
+        /// Users of this method must provide an absolute validation path from the root viewmodel. Individual path segments need to be delimited using the '/' character. 
+        /// Example 1) /Customer/Id. Example 2) /Items/0/Price.
+        /// </summary>
+        /// <param name="absolutePath">Absolute validation path from the root viewmodel</param>
+        /// <param name="message">Validation error message</param>
+        public static ViewModelValidationError AddRawModelError(this IDotvvmRequestContext context, string absolutePath, string message)
```
exyi, line 33: I think we should keep this overload, but make it obsolete. In many cases, it will still work, because most people were having most of the validated properties in the root view model. Moreover, we can explain in the obsolete comment what should they use now, this way they will just get an "overload not found" error.

PR #1288, opened 2022-02-10
Throw an error when an orphaned property group is present

exyi:
It's an indication that the deprecated method
of declaring property groups has been used,
so it's better to throw an error informing the user what to do
than just silently ignore the property.


acizmarik (approved):
👍

PR #1291, opened 2022-02-10
Added pdb files to nuget packages

acizmarik:
This PR changes the way we bundle nuget packages by including pdb files. This should enhance debugging experience and also show some more information on error pages.


exyi (approved):
Yes :)

PR #1231, opened 2021-11-27
Markup declared properties

exyi:

PR #1294, opened 2022-02-13
Better error message in unknown type in service directive

tomasherceg:


Mylan719:
This logic has been completely rewritten in #1231 I would not recomend merging this now.
See https://github.com/riganti/dotvvm/blob/caeeaad63c5555da6356b4d762918e5e4bed8025/src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedTreeBuilder.cs#L68
and
https://github.com/riganti/dotvvm/blob/markup-declared-properties/src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedServiceInjectDirective.cs

exyi (approved):

PR #1299, opened 2022-02-17
TextOrContentCapability exception added

Duzij:
Either Text property or Content must be set exception added


exyi (approved):
:+1: 

I will maybe add an `optional` argument which will make it return an empty enumerable, it might be nice in some cases.

PR #1300, opened 2022-02-17
AddCssClass ValueOrBinding overload added

Duzij:


src/Framework/Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -188,6 +188,13 @@ public static TControl AddAttribute<TControl>(this TControl control, string attr
...
 
+        /// <summary> Appends a css class to this control. Note that it is currently not supported if multiple bindings would have to be joined together. Returns <paramref name="control"/> for fluent API usage. </summary>
+        public static TControl AddCssClass<TControl, TValue>(this TControl control, ValueOrBinding<TValue>? value)
+            where TControl : IControlWithHtmlAttributes
+        {
+            return AddAttribute(control, "class", value?.UnwrapToObject());
+        }
+
         /// <summary> Appends a css class to this control. Note that it is currently not supported if multiple bindings would have to be joined together. Returns <paramref name="control"/> for fluent API usage. </summary>
         public static TControl AddCssClass<TControl>(this TControl control, ValueOrBinding<string> className)
```
exyi, line 199: I think that the overload already exists. However, it's not nullable, you can add the nullability to the existing overload.  It only accepts `string`, but I don't think we should accept anything else as a css class

Duzij: I created this overload to be able to do something like this in my composite controls:
`
var foo = ValueOrBinding<MyEnum>(MyEnum.Default)....
AddCssClass(foo )
`
This enum has `[EnumMember]` annotation, can I maybe convert `foo` variable to `ValueOrBinding<string>`?

exyi: I think that in that case you should prefer to call `foo.AsString()` to convert it to `ValueOrBinding<string>`

Duzij: Continue in https://github.com/riganti/dotvvm/pull/1307

PR #1307, opened 2022-02-22
AsString() Enums support added

Duzij:
After a short coding session with @tomasherceg, enums support was added to .AsString() method.
Solution is not ideal and need further discussion.


src/Framework/Framework/Compilation/Javascript/JavascriptTranslatableMethodCollection.cs
```diff
@@ -203,6 +203,9 @@ private void AddDefaultToStringTranslations()
         {
             AddMethodTranslator(typeof(object), "ToString", new PrimitiveToStringTranslator(), 0);
             AddMethodTranslator(typeof(Convert), "ToString", new PrimitiveToStringTranslator(), 1, true);
+            AddMethodTranslator(typeof(ReflectionUtils), "ToEnumString", parameterCount: 1, translator: new GenericMethodCompiler(
+                args => args[1].Clone()
```
exyi, line 207:
```suggestion
                args => args[1]
```

quigamdev (approved):
Looks good to me.

PR #1308, opened 2022-02-22
ActiveDotvvmProperty support for null command binding added

Duzij:
In case click was null, `control.SetProperty(Events.ClickProperty, click);` resulted in Exception, but rather property should not be set.

### DIFF
--- a/src/Framework/Framework/Binding/ActiveDotvvmProperty.cs
+++ b/src/Framework/Framework/Binding/ActiveDotvvmProperty.cs
@@ -10,20 +10,27 @@ using DotVVM.Framework.Hosting;
 
 namespace DotVVM.Framework.Binding
 {
-    public abstract class ActiveDotvvmProperty: DotvvmProperty
+    public abstract class ActiveDotvvmProperty : DotvvmProperty
     {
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Binding/ActiveDotvvmProperty.cs
```diff
@@ -10,20 +10,20 @@
...
         {
-            return DelegateActionProperty<ICommandBinding>.Register<TDeclaringType>(name, (writer, context, property, control) =>
-            {
-                var binding = control.GetCommandBinding(property) ?? throw new DotvvmControlException(control, $"Command binding expression was expected in {property}.") {
-                    RelatedProperty = property
-                };
-                var script = KnockoutHelper.GenerateClientPostBackScript(name, binding, control);
-                writer.AddAttribute(attributeName, script);
+            return DelegateActionProperty<ICommandBinding>.Register<TDeclaringType>(name, (writer, context, property, control) => {
+                var binding = control.GetCommandBinding(property);
```
exyi, line 21: This should throw an exception if the property value is a constant. You can use `(ICommandBinding)control.GetValueRaw` or add a custom error check, but `GetCommandBinding` returns null if there is a constant.

PR #1313, opened 2022-03-01
SetProperty added support for nullable ValueOrBinding

Duzij:


src/Framework/Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -50,7 +50,6 @@ public static TControl SetProperty<TControl>(this TControl control, string propN
             control.SetBinding(control.GetDotvvmProperty(propName), binding);
             return control;
         }
-
         /// <summary> Sets value or binding into the DotvvmProperty referenced in the lambda expression. Returns <paramref name="control"/> for fluent API usage. </summary>
         public static TControl SetProperty<TControl, TProperty>(this TControl control, Expression<Func<TControl, TProperty>> prop, ValueOrBinding<TProperty> value)
```
exyi, line 54: I think that you can just add the `?` to this method

exyi (approved):

PR #1314, opened 2022-03-04
Add binding property with information about used properties

exyi:
Created originally for DynamicEditor, but turns out it
could also be used for some BP controls.


exyi:
Superseded by #1385

PR #1319, opened 2022-03-07
Serializability analysis fix for generics and suppress on compilation errors

acizmarik:
This PR introduces some further changes / improvements to DotVVM analyzers.

1. Generic type parameters of viewmodels have been excluded from serializability analysis. It gets a bit more complicated when users define generic viewmodels. It can be implemented in future, but I doubt it is commonly used. Therefore, for now we assume that these are serializable

2. Serializability analysis is not longer run on incomplete type members (nodes with attached compilation errors). Otherwise this could get a bit annoying, for example in the following situation:
   - User tries to add a new property to viewmodel and types the following: `public int MyProperty`
   - We immediately issue a warning that user should not use fields for storing viewmodel state (notice that field definition is not completed - there is a missing semicolon, nor it is a property definition as it is missing accessors)
   - User finishes the definition by adding ` { get; set; }`
   - The warning is removed as it is now considered a serializable property

### DIFF
--- a/src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
+++ b/src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
@@ -101,6 +101,10 @@ namespace DotVVM.Analyzers.Serializability
             var semanticModel = context.SemanticModel;
             foreach (var property in viewModel.ChildNodes().OfType<PropertyDeclarationSyntax>())
             {
+                // Make sure there are not compilation errors attached
+                if (HasCSharpCompilerErrorsAttached(property))
+                    continue;
+
                 // Resolve property symbol
                 if (semanticModel.GetDeclaredSymbol(property) is not IPropertySymbol propertySymbol)
                     continue;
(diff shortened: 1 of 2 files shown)


src/Analyzers/Analyzers/Serializability/ViewModelSerializabilityAnalyzer.cs
```diff
@@ -222,6 +223,13 @@ private static void AnalyzePropertyType(ITypeSymbol propertyType, Location locat
                     }
                     break;
 
+                case TypeKind.TypeParameter:
+                    // Assume this is correct. This gets otherwise complicated. Possible solution:
+                    // 1.) Check for available generic type constraints (these might hint whether the type is serializable or not)
+                    // 2.) Check the context this viewmodel is used in (i.e. analyze only concrete instantiations of the generic type definition)
+                    // TODO: Maybe add (partial) support for this in future
+                    break;
```
exyi, line 231: yea, I doubt it's so commonly used and this is only a hint to users, the validation does not need to be bullet proof

exyi (approved):
:+1:

PR #1321, opened 2022-03-08
TextOrContentCapability Content can be used multiple times

tomasherceg:
I've found issues when the Content property of the capability are used multiple times (e. g. in `Repeater`).
This fix clones the controls so it should work.


exyi:
I think that ITemplate should be used instead of DotvvmControl when it's used multiple times.

In the very least, please don't clone the controls unless it's used for the second time.

tomasherceg:
In such case we should change it to `ITemplate` ASAP since we are planning to release 4.0.0 and this could be a big breaking change.

PR #1322, opened 2022-03-08
Null-value attributes and styles should not be rendered 

tomasherceg:
I am not sure if this is a good idea, but the typical scenario is when we need to conditionally include some attribute:

```
control.SetAttribute("disabled", enabled.Select(v => v ? null : "disabled"));
``` 

I want either to include the attribute (`disabled=disabled` value), or do not render it at all.

However, when the `enabled` is `true` on the server (the transformed expression evaluates to `null`), it renders the `disabled`:
```
data-bind="attr: { 'disabled': Enabled() ? null : 'disabled'}" disabled="" ...
```

When the page loads, the attribute is removed by Knockout because it supports this logic. But our server-side rendering doesn't respect this.

I am not sure how many things we can break by this, but I have a feeling that on some places we have this logic and on other places we do not.

### DIFF
--- a/src/Framework/Framework/Controls/DotvvmBindableObjectHelper.cs
+++ b/src/Framework/Framework/Controls/DotvvmBindableObjectHelper.cs
@@ -244,9 +244,11 @@ namespace DotVVM.Framework.Controls
         }
 
         /// <summary> Adds a css inline style - the `style` attribute. Returns <paramref name="control"/> for fluent API usage. </summary>
-        public static TControl AddCssStyle<TControl>(this TControl control, string name, string styleValue)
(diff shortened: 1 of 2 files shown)


exyi:
This is a rather problematic breaking change since null was interpreted as an attribute without value and I'm afraid we'll break quite a few things by this change. Simplest example is  that writing `<input enabled />` in dothtml should  output the same HTML. With your change, it will output `<input />`.

Maybe it would be better to find a better representation of attributes without value, but I'd like avoid squeezing such a major change into the v4 release. However, let's discuss this for the next major release.



For your use case of the disabled attribute, you can simply use the boolean as the binding. `false` removes the binding while `true` includes it (without value, I think)

PR #1324, opened 2022-03-09
Probably fix magic null reference exception

exyi:
The SelectMany call seems to cause null reference exception
on .NET Framework. I have no idea why, but this could fix it.

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/DataContextStack.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/DataContextStack.cs
@@ -57,18 +57,17 @@ namespace DotVVM.Framework.Compilation.ControlTree
 
         public ImmutableArray<Delegate> GetAllBindingPropertyResolvers()
         {
-            // most likely, there aren't any so just quickly scan the link list
+            var builder = ImmutableArray.CreateBuilder<Delegate>();
+
             var c = this;
             while (c is {})
             {
-                if (!c.BindingPropertyResolvers.IsEmpty)
-                    break;
+                builder.AddRange(c.BindingPropertyResolvers);
                 c = c.Parent;
             }
-            if (c is null)
-                return ImmutableArray<Delegate>.Empty;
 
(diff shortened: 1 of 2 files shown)


exyi:
@luboshl please have a look

src/Framework/Framework/Compilation/ControlTree/DataContextStack.cs
```diff
@@ -57,18 +57,17 @@ public IEnumerable<(int dataContextLevel, BindingExtensionParameter parameter)>
...
         {
-            // most likely, there aren't any so just quickly scan the link list
+            var builder = ImmutableArray.CreateBuilder<Delegate>();
+
             var c = this;
             while (c is {})
             {
-                if (!c.BindingPropertyResolvers.IsEmpty)
-                    break;
+                builder.AddRange(c.BindingPropertyResolvers);
```
luboshl, line 65: @exyi Is it possible that in the hierarchy (e.g. root -> child1) there is _child1_ with non-empty `BindingPropertyResolvers` and also _root_ with non-empty `BindingPropertyResolvers`? In that case this new implementation would be different than the original one - now `BindingPropertyResolvers` from both _chil1_ and _root_ are in the result where only `BindingPropertyResolvers` from _child1_ would be in the result in the original implementation.

exyi: This case is possible, but I think that the behavior is still the same.

> only BindingPropertyResolvers from child1 would be in the result in the original implementation.

The EnumerableItems() enumerated itself and all ancestors, so it returned both root and child1. The SelectMany would then also include resolvers from both of them.

tomasherceg (approved):

src/Framework/Framework/Compilation/ControlTree/DataContextStack.cs
luboshl: You are right. Than it's ok.

PR #1325, opened 2022-03-09
Add support for IEnumerable<T> in control properties

exyi:
Also improves the error messages when property type is not suported + a test that it works automatically (without MarkupOption) in CompositeControl.

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
@@ -236,7 +236,7 @@ namespace DotVVM.Framework.Compilation.ControlTree
             if (controlMetadata.DataContextConstraint != null && dataContext != null && !controlMetadata.DataContextConstraint.IsAssignableFrom(dataContext.DataContextType))
             {
                 ((DothtmlNode?)dataContextAttribute ?? element)
-                   .AddError($"The control '{controlMetadata.Type.Name}' requires a DataContext of type '{controlMetadata.DataContextConstraint.FullName}'!");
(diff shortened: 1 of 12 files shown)


src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
```diff
@@ -770,9 +758,29 @@ private IAbstractPropertySetter ProcessElementProperty(IAbstractControl control,
                     return treeBuilder.BuildPropertyControl(property, null, propertyWrapperElement);
                 }
             }
+            else if (IsCollectionProperty(property))
+            {
+                var collectionType = GetCollectionType(property);
```
exyi, line 763:
```suggestion
                var collectionType = GetElementType(property);
```

and use ReflectionUtils if possible

src/Framework/Framework/Controls/TextOrContentCapability.cs
```diff
@@ -61,5 +61,10 @@ public IEnumerable<DotvvmControl> ToControls()
             else
                 return Content.NotNull();
         }
+
+        public ITemplate ToTemplate()
+        {
+            return new CloneTemplate();
+        }
```
exyi, line 68:
```suggestion
```

exyi: I rather fixed this method, it might still be useful

src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
exyi: ReflectionUtils can't be used, because this works on the ITypeDescriptor. I don't want to rename the method, since it's method which already existed and is virtual

acizmarik (approved):

PR #1328, opened 2022-03-11
Fix error message when a method in binding does not exist

exyi:
It used to produce something like "must be reducible node",
which is really unhelpful. Now it will be something like
"Method X was not found"


acizmarik (approved):
👍

PR #1329, opened 2022-03-12
Information about current route is sent to the client

tomasherceg:


src/Framework/Framework/Resources/Scripts/dotvvm-base.ts
```diff
@@ -16,6 +16,10 @@ type DotvvmCoreState = {
     _viewModelCacheId?: string
     _virtualDirectory: string
     _initialUrl: string,
+    _routeName: string,
+    _routeParameters: {
+        [name: string]: any
+    },
```
exyi, line 22: I'm just thinking that it would maybe be bit better to place all these static values (culture, virtualDirectory, initialUrl, routeName, routeParameters, ...) into a common object, so we won't have to copy all the the property names on multiple places.

PR #1330, opened 2022-03-12
callNamedCommand JS function added

tomasherceg:


quigamdev:
Please can you explain what issue this PR solves? @tomasherceg

tomasherceg:
This PR aims to allow calling named commands via JS public API. 
I need this in the MAUI integration - I've changed this to a draft because maybe I'll have to change the API of this method so it doesn't get the viewModuleId (which is internal).

quigamdev:
Is this PR still relevant?

PR #1338, opened 2022-03-28
Dynamic assemblies excluded from traversing on OWIN

tomasherceg:
We were getting "The invoked member is not supported in a dynamic assembly." exception.


acizmarik (approved):

acizmarik:
I just tested this locally (OWIN, ASP.NET Core) and tests seem to pass. There is probably an unrelated problem on CI, however, that does not keep this PR from being merged and released

PR #1339, opened 2022-03-29
Added extensibility point to provide custom fetch implementation

tomasherceg:
I am not sure if it is a good idea to provide this extensibility point in the public API, but the attacker can replace `window.fetch` anyway so it's probably not such a big deal. 
Alternatively, we can bundle a special version of DotVVM for the MAUI hosted version - what do you think?


tomasherceg:
I've done this as part of this draft PR #1368. 

Depends on #1329 which should be merged first.

I've put there only the necessary things for DotVVM app to run inside WebView - the other things are still in the [external repo](https://github.com/riganti/dotvvm-maui) but I plan to integrate it into DotVVM repository soon.

PR #1340, opened 2022-03-30
Add helpers for compiling custom `value` and `staticCommand` bindings

exyi:
Note that I'm not adding helpers for compiling new
command and resource bindings,
because they are better created from Func<...>, since they
are only evaluated compile-time.


quigamdev (approved):
Nice 👍

PR #1341, opened 2022-03-30
Improve the Control used in another parent error

exyi:
The error now includes both of the related controls in properties,
and both control types are also mentioned in the error message.


src/Framework/Framework/Controls/DotvvmControlCollection.cs
```diff
@@ -431,5 +431,13 @@ public bool HasOnlyWhiteSpaceContent()
             }
             return true;
         }
+
+        public record ControlUsedInAnotherTree(DotvvmControl ParentControl, DotvvmControl AddedControl):
```
exyi, line 435:
```suggestion
        public record ControlUsedInAnotherTreeException(DotvvmControl ParentControl, DotvvmControl AddedControl):
```

acizmarik (approved):
👍

PR #1342, opened 2022-03-30
Update dotnet new template

exyi:
Updated the template to DotVVM 4 and .NET 6.
I have also added a few things from the official asp.net core template

* error handler, so the app displays at least something when it crashes
* https redirect and HSTS
* removed some code which is no longer needed

I also moved the viewmodels and .dothtml files into one
Views folder, people seem to prefer this way of organizing their files


src/Templates/content/EmptyWeb/Views/Error/ErrorViewModel.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public ErrorViewModel()
+        {
+        }
+
+        public override Task Init()
+        {
+            var aspcontext = Context.GetAspNetCoreContext();
+            var exceptionInfo = aspcontext.Features.Get<IExceptionHandlerFeature
+>();
```
cafour, line 33:
```suggestion
            var exceptionInfo = aspcontext.Features.Get<IExceptionHandlerFeature>();
```

cafour (commented):
Since other .NET 6 templates already use them, shouldn't we migrate to file-scoped namespaces as well?

quigamdev (approved):
Thanks 👍

PR #1344, opened 2022-03-31
SetDataContext helper methods added

Duzij:


src/Framework/Framework/Controls/DotvvmBindableObjectHelper.cs
```diff
@@ -16,6 +18,21 @@ namespace DotVVM.Framework.Controls
...
+            return control;
+        }
+
+        /// <summary> Sets binding into the DataContextTypeProperty. Returns <paramref name="control"/> for fluent API usage. </summary>
+        public static TControl SetDataContextType<TControl>(this TControl control, DataContextStack? stack)
+            where TControl : DotvvmBindableObject
+        {
+            control.properties.Set(Internal.DataContextTypeProperty, stack);
+            return control;
+        }
```
exyi, line 35: Didn't you forget to remove the existing `SetDataContextTypeFromDataSource` and `SetDataContext` from BindingHelper? Otherwise they will be ambiguous. I'd also prefer to modify the existing methods in BindingHelper than adding new ones to a different class.

Duzij: I fixed two methods as requested.

exyi (approved):

PR #1346, opened 2022-04-04
Set up source link

cafour:
Sets up Source link based on instructions from https://github.com/dotnet/sourcelink.

Also embeds the PDBs in the symbol package (including all C# sources) instead of the main package. (Visual Studio IMO just ignored them anyway because of the missing SourceLink package.) This makes the main package smaller by ~800 KB.

I tested it locally but I'll test it on our internal feed as well.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -33,10 +33,8 @@
 
   <PropertyGroup Label="SourceLink">
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <!-- Symbols are included in the main package. -->
+    <!-- NB: Don't generate symbol packages. Symbols and sources are packed into the dlls. -->
     <IncludeSymbols>false</IncludeSymbols>
-    <!-- <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder> -->
     <DebugType>embedded</DebugType>
     <EmbedAllSources>true</EmbedAllSources>
   </PropertyGroup>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -33,10 +33,11 @@
 
   <PropertyGroup Label="SourceLink">
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <DebugType>portable</DebugType>
+    <!-- Symbols are included in the main package. -->
+    <IncludeSymbols>false</IncludeSymbols>
+    <!-- <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder> -->
+    <DebugType>embedded</DebugType>
     <EmbedAllSources>true</EmbedAllSources>
   </PropertyGroup>
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -39,14 +39,6 @@
     <EmbedAllSources>true</EmbedAllSources>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


cafour:
Turns out, source link breaks unit tests. Who would have though. I'll fix it. :D

cafour:
@exyi  Can you please point to a place in DotVVM where we depend on there being pdbs in the NuGet package itself?

cafour:
Another issue with packing pdbs in the main NuGet package is that since .NET Core 3.0, it does not work at all. See https://github.com/dotnet/sourcelink/issues/628 and https://github.com/dotnet/sdk/issues/1458.

As far as I can see, the only approaches that work as of now are:
1. Using `snupkg` for the public NuGet packages and Azure DevOps Artifact Symbol Server for our internal feed.
2. Embedding the pdbs directly into the dlls by setting `DebugType` to `embedded`. (I haven't tested it yet. Will do so hopefully friday morning.)

As for the size of the Framework PDB, it's only ~800 KB, because `dotnet pack`'s `--include-source` is also thoroughly broken (see  https://github.com/dotnet/docs/issues/14165). We try to use this option in our publish script. If one were to use `EmbedAllSources`, _the other_ way of including all of source code in the package, the pdb grows to almost 3 MB. This settles the total size of the main DotVVM package to ~6 MB. However, do we even care about the size of the package at all?

cafour:
Update: `DebugType="embedded"` works fine! Both _Go to reference_ and _Step into_ use the sources embedded into the pdb (which is embedded into the dll). The only caveat is that the package has ~5.5 MB.

To try it out, use the `4.0.2-preview03-257319-sourcelink` packages on the private feed.

src/Directory.Build.props
```diff
@@ -11,24 +11,11 @@
...
-    <Exec Command="git rev-parse HEAD"
-          ConsoleToMsBuild="true"
-          ContinueOnError="true"
-          StandardOutputImportance="low" StandardErrorImportance="low">
-      <Output TaskParameter="ConsoleOutput" PropertyName="SourceRevisionId" />
-    </Exec>
-    <PropertyGroup>
-      <RepositoryCommit>$(SourceRevisionId)</RepositoryCommit>
-    </PropertyGroup>
-  </Target>
```
exyi, line 25: This is now handled by the `Microsoft.SourceLink.GitHub` package?

cafour: Yup. I just tested it and it added
```
<repository type="git" url="https://github.com/riganti/dotvvm.git" commit="f6d6f7a6101ea5adcb15f647b8f6c863ecb1f8ac" />
```
to the nuspec file in the package.

Nuget package explorer confirms this as well:

image

src/Directory.Build.props
```diff
@@ -42,10 +29,20 @@
...
-    <!-- Pack pdb symbol files -->
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
 
+  <PropertyGroup Label="SourceLink">
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- NB: Don't generate symbol packages. Symbols and sources are packed into the dlls. -->
+    <IncludeSymbols>false</IncludeSymbols>
+    <DebugType>embedded</DebugType>
+    <EmbedAllSources>true</EmbedAllSources>
```
exyi, line 39: Ok, just checking I understand it correctly: without this line we'd get the file names in exception stack traces, but Go To Definition in IDEs wouldn't work?

exyi (approved):
Awesome, I didn't know it's possible to pack the dlls into the main `dll`, it makes a lot of things easier :)

cafour:
**Explanation of the new build settings**

* `IncludeSymbols` -- Should I generate a symbol package?
* `DebugType` -- What should happen to the pdbs?
  * `full/pdbonly` -- use the only-pdb format on Windows, and the portable format on Linux
  * `portable` -- use the portable, cross-platform format
  * `embedded` -- use the portable format and embed it into the dll
* `EmbedAllSources` -- embed all version-controlled sources into the pdb (requires a `Microsoft.SourceLink.*` package)


We can't use `IncludeSymbols=true` because Azure Artifacts don't support them.

We can't use `DebugType=portable` because .NET Core >=3 kinda forgets to copy them from the NuGet package.

We can but don't have to use `EmbedAllSources=true`. It's just convenient to pack the sources as well, since then source link should work without internet connection.

src/Directory.Build.props
cafour: See https://github.com/riganti/dotvvm/pull/1346#issuecomment-1099103661 for some explanation.

I toyed for a bit with the error page: Without `DebugType=embedded`, we do get filenames but we **don't have line positions** because we **don't have access to the pdb** (nobody does, it isn't copied to `bin/`).

Our error pages have been broken for a while (apparently since .NET 3.0), we just didn't notice, I guess.

PR #1349, opened 2022-04-07
Fix package icons

cafour:
Currently, all our package icons on NuGet are broken since `PackageIconUrl` is deprecated and the link is dead anyway. This should fix it.


exyi (approved):
Great

exyi:
You can merge this to release/4.0

PR #1351, opened 2022-04-08
Fix Analyzers.Package build

cafour:
Adds a Analyzers.Package project reference to Framework with `ReferenceOutputAssembly=false`. This causes Analyzers.Package to be transitively built instead of it being assumed that it is has already been built.


acizmarik (approved):
Nice 👍

cafour:
@exyi Considering your comment on #1349, should this also go in `release/4.0`? Should all new PRs target `release/4.0`?

PR #1352, opened 2022-04-08
Feature/add dotvvm version serialized configuration

acizmarik:
This PR adds DotVVM Framework version to the JSON that contains serialized DotVVM configuration. Obtaining this information within extension code-base is harder. Moreover, this information does not change frequently.

### DIFF
--- a/src/Tests/Runtime/config-tests/ConfigurationSerializationTests.AuxOptions.json
+++ b/src/Tests/Runtime/config-tests/ConfigurationSerializationTests.AuxOptions.json
@@ -1,4 +1,5 @@
 {
+  "dotvvmVersion": "*.*.*.*",
   "config": {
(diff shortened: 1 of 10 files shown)


src/Tests/Runtime/config-tests/ConfigurationSerializationTests.AuxOptions.json
```diff
@@ -1,4 +1,5 @@
 {
+  "dotvvmVersion": "4.0.0.0",
```
exyi, line 2: We should replace it in the tests, otherwise it will make the tests fail every time we change the version.

acizmarik: Yes, I replaced the value with `"dotvvmVersion": "*.*.*.*"`

PR #1354, opened 2022-04-13
Allow multiple AddClass() even when some contain binding

exyi:
Since it's nearly impossible to concatenate two bindings or
a binding and a constant value, this patch adds support for a list
of binding or values directly to HtmlGenericControl.AddAttributesToRender

I added AttributeList, a special linked list for this use case. The reason is
that we need to append efficiently in the MergePlainValues method
(or prepend, that doesn't matter). It's also far more efficient to typecheck
for sealed class than an generic interface like `IEnumerable<T>`
(the typecheck will be done for each value in HtmlGenericControl)

In order to generate reasonable code, I needed to fix some bugs in
OperatorPrecdence:

* distinguish between no parens needed ever (0) and sequence expression (1)
   - for this reason, all values were shifted by one. And I made constants
* some expressions (parenthized expr, invocation, ...) have a different
   precence from the inside than from the outside. This used to be worked
   around by checking the parent node in `GetOperatorPrecedence`, but that
   doesn't work well when the expression is treated as independent from the
   its parent. I changed this by adding the `GetParentLevel` which gives
   the inner precedence of the parent node as seen from the current
   expression.
* JoinValuesOrBindings should just return the original binding if there are no fragments for joining - it just resulted in unnecesary parenthesis being added

cc @Duzij, you have been asking for this for some time, so feel free to test it out.

### DIFF
--- a/src/Framework/Framework/Controls/HtmlGenericControl.cs
+++ b/src/Framework/Framework/Controls/HtmlGenericControl.cs
@@ -12,6 +12,7 @@ using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Text;
 using DotVVM.Framework.Compilation.Javascript;
+using FastExpressionCompiler;
 
 namespace DotVVM.Framework.Controls
 {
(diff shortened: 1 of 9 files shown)


src/Framework/Framework/Controls/HtmlGenericControl.cs
```diff
@@ -344,9 +346,18 @@ private void AddCssStylesToRender(IHtmlWriter writer)
...
+                writer.AddAttribute(name, str, append: true);
+            }
+            else if (value is AttributeList list)
+            {
+                for (var i = list; i is not null; i = i.Next)
+                {
+                    AddHtmlAttribute(writer, name, i.Value);
+                }
             }
             else if (value is IEnumerable<string>)
```
tomasherceg, line 362: Remove this branch - the implementation is wrong and we should not support this.

tomasherceg (approved):

PR #1361, opened 2022-04-19
ClientID methods made public and allows specifying prefixes and suffixes

tomasherceg:


tomasherceg:
We should reimplement the methods returning `object` to return `ValueOrBinding<string>` as well.

tomasherceg:
We should reimplement the methods returning `object` to return `ValueOrBinding<string>` as well.

src/Framework/Framework/Controls/ItemsControl.cs
```diff
@@ -90,32 +90,36 @@ public static void OnCompilation(ResolvedControl control, BindingCompilationServ
...
-                            Expression.Parameter(typeof(int), "_index").AddParameterAnnotation(
-                                new BindingParameterAnnotation(dataContext, new CurrentCollectionIndexExtensionParameter()))
-                        )
-                    }
-                ));
+                        new ParsedExpressionBindingProperty(CreateIndexBindingExpression(dataContext))
+                    }));
             }
+
+            return new ValueOrBinding<int>(binding).Select(i => IndexToStringNoGlobalize(i));
```
exyi, line 121: I'm really not a fan of using the Select in core components at runtime. I don't know why can't we leave this logic as it was before, insisting on using `ValueOrBinding<string>` internally causes more problems than good things...

acizmarik (approved):

PR #1365, opened 2022-04-22
Don't render id twice when both ID property and id attribute are set

### LINKED ISSUE #1362: AddAttribute("id") collides with ID in HtmlCapability
In some controls, we need to make sure that the control will have some ID - either user-specified or the control-generated.
When a `CompositeControl` accepts a `HtmlCapability` and renders an `id` attribute at the same time, the ID is rendered twice (concatenated with semisolon): `id="krtek;krtek"`

```
public class MyControl : CompositeControl 
{
    public DotvvmControl GetContents(
       HtmlCapability html,
       ...
    ) {
        var id = CreateClientId() ?? GetDotvvmUniqueId();
        return new HtmlGenericControl("div", html)
            .AddAttribute("id", id);
    }
}
```

exyi:
resolves #1362

### DIFF
--- a/src/Framework/Framework/Controls/HtmlGenericControl.cs
+++ b/src/Framework/Framework/Controls/HtmlGenericControl.cs
@@ -404,6 +404,7 @@ namespace DotVVM.Framework.Controls
                 if (prop is not GroupedDotvvmProperty gprop || gprop.PropertyGroup != AttributesGroupDescriptor)
                     continue;
 
+                var attributeName = gprop.GroupMemberName;
                 var knockoutExpression = valueRaw switch {
                     AttributeList list => list.GetKnockoutBindingExpression(this, HtmlWriter.GetSeparatorForAttribute(gprop.GroupMemberName)),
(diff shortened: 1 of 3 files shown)


src/Framework/Framework/Controls/HtmlGenericControl.cs
```diff
@@ -412,19 +413,25 @@ private void AddHtmlAttributesToRender(ref RenderState r, IHtmlWriter writer)
...
-                        attributeBindingGroup.Add(gprop.GroupMemberName, knockoutExpression);
+                        attributeBindingGroup.Add(attributeName, knockoutExpression);
                     }
                     if (!r.RenderOnServer(this))
                         continue;
                 }
-                AddHtmlAttribute(writer, gprop.GroupMemberName, valueRaw);
+                AddHtmlAttribute(writer, attributeName, valueRaw);
+
+                if (attributeName == "id")
```
acizmarik, line 430: We should ensure that `attributeName` is always lower-case and invariant culture :)

acizmarik (approved):
Please just ensure that the unit test on Mac works before merging

PR #1366, opened 2022-04-22
Fix generated in Repeater created from code

### LINKED ISSUE #1363: Control IDs are generated wrong in Repeater created from code
When created from code, the `Repeater` control doesn't specify the `Internal.CurrentIndexBindingProperty` - it's set only during the [view compilation](https://github.com/riganti/dotvvm/blob/bb1ad2d70878738331710f35500bc8f8aabb059b/src/Framework/Framework/Controls/ItemsControl.cs#L89).
 
Therefore, the `Internal.ClientIDFragment` property is not set and the IDs generated in the child items are not correct in client-side rendering mode.

exyi:
Repeater used to rely on Internal.CurrentIndexBindingProperty being set
during view compilation, this patch adds a fallback path
for when the repeater is created at runtime.

resolves #1363

Turns out we already had a test covering this, but nobody
noticed that the test is also wrong :]


quigamdev (approved):
Thanks 👍

PR #1369, opened 2022-04-26
validation target: Use lambda function instead of a string

exyi:
This gets rid of one more eval in our codebase ;)

### DIFF
--- a/src/Framework/Framework/Controls/ValidationSummary.cs
+++ b/src/Framework/Framework/Controls/ValidationSummary.cs
@@ -65,11 +65,12 @@ namespace DotVVM.Framework.Controls
         {
             base.AddAttributesToRender(writer, context);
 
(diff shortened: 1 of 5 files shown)


src/Framework/Framework/Controls/ValidationSummary.cs
```diff
@@ -65,12 +65,13 @@ protected override void AddAttributesToRender(IHtmlWriter writer, IDotvvmRequest
         {
             base.AddAttributesToRender(writer, context);
 
-            var expression = KnockoutHelper.GetValidationTargetExpression(this);
-            if(expression == null)
+            if (!(bool)this.GetValue(Validation.EnabledProperty)!)
```
acizmarik, line 68: Maybe something like `false.Equals(...)`

acizmarik (approved):

PR #1371, opened 2022-04-27
Warn the user when view model is too large or request is slow

exyi:
I set the default thresholds at 5MB and 3 seconds, but it's
configurable (if you know your users are more impatient or
more patient than what I assumed).
It's enabled in both debug and release mode, since the
checks should have little overhead and they will be useful
to see in production logs.

The warning also includes information about which part of
the viewmodel is taking up significant space /
which stages of the request took the longest.

```
The serialized view model has 7.7MB, which may make your application quite slow. Property DefaultViewModel.RouteData.Bordel takes 70%, Property DefaultViewModel.RouteData.Bordel2 takes 26%, Class DefaultViewModel.RouteData takes 100%

The request took 5.7s to process, 2 slowest events: InitCompleted (5.01s), OutputRendered (0.50s).
```


acizmarik (approved):
Nice 👍

PR #1375, opened 2022-04-28
Add JS API for modifying the client-side error collection

exyi:
The new API has 2 new functions:
* `dotvvm.validation.addErrors([{ errorMessage: "Test", propertyPath: "/X/Y/Z" }])`
* `dotvvm.validation.removeErrors("/FirstName", "/LastName")`

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/validation/validation.ts
+++ b/src/Framework/Framework/Resources/Scripts/validation/validation.ts
@@ -1,6 +1,6 @@
 import * as evaluator from "../utils/evaluator"
 import { validators } from './validators'
-import { allErrors, detachAllErrors, ValidationError, getErrors } from "./error"
+import { allErrors, detachAllErrors, detachErrors, ValidationError, getErrors } from "./error"
 import { DotvvmEvent } from '../events'
 import * as spaEvents from '../spa/events'
 import { postbackHandlers } from "../postback/handlers"
@@ -21,7 +21,7 @@ type ValidationSummaryBinding = {
     hideWhenValid: boolean
 }
 
-type DotvvmValidationErrorsChangedEventArgs = PostbackOptions & {
+type DotvvmValidationErrorsChangedEventArgs = Partial<PostbackOptions> & {
     readonly allErrors: ValidationError[]
 }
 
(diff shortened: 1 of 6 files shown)


src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -280,24 +288,44 @@ function getValidationErrors<T>(
...
 
+export type ValidationErrorDescriptor = {
+    /** Error to be displayed to the user */
+    errorMessage: string
+    /** Path in the view model to the annotated property, for example `/LoginPage/Name` */
+    propertyPath: string
+}
+
+export type AddErrorsOptions = {
+    /** Root object from which are the property paths are resolved. By default it's the root view model of the page */
```
acizmarik, line 299:
```suggestion
    /** Root object from which are the property paths resolved. By default it's the root view model of the page */
```

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -280,24 +288,44 @@ function getValidationErrors<T>(
...
+    root?: KnockoutObservable<any> | any
+    /** When set to false, the validationErrorsChanged is not trigger. By default, the error is triggered. The event controls the `Validator`s in the DotHTML page, so when it's not trigger, the change won't be visible. */
```
acizmarik, line 301: I am confused by this comment. To me, it seems like the `triggerErrorsChanged` is set to `false` by default. Also when is it desirable to add errors but not propagate these changes to `Validator`s?

```suggestion
    /** When set to false, the validationErrorsChanged would not trigger. The event controls the `Validator`s state in the DotHTML page, so when it's not triggered, the change won't be visible. */
```

exyi: > Also when is it desirable to add errors but not propagate these changes to Validators?

it's "needed" when the method is called multiple times and you don't want to trigger the event for each invocation. Maybe it's just more confusing, I'll consider removing this (at least for now), it's probably just confusing.

exyi: Actually, we need it for our own usage in the `showValidationErrorsFromServer` function bellow, I couldn't deduplicate the code without this switch.

acizmarik (approved):
👍

PR #1377, opened 2022-05-02
Publish scripts for npm package with types

tomasherceg:


ci/scripts/npm/build-dotvvm-types.ps1
```diff
@@ -1,5 +1,19 @@
...
+$typescriptVersion = $packageJson.devDependencies.typescript
+
+npm run tsc-types
+copy ./obj/typescript-types/dotvvm.d.ts "$dir/dotvvm-types/index.d.ts"
+
+pop-location
+
+push-location dotvvm-types
+
+npm install typescript@$typescriptVersion --save-dev
```
exyi, line 16: why do we need to install typescript in this package? The typescript compiler is executed in src/Framework/Framework, not here.

tomasherceg: It's only to put the same version of Typescript dependency into package.json here.

tomasherceg: Oh, I see. The dependency was there before, so I just updated the script to push there the same version that is used in DotVVM codebase. If it's not needed, we can safely remove it.

ci/scripts/npm/build-dotvvm-types.ps1
```diff
@@ -1,5 +1,18 @@
-push-location ../../DotVVM.Framework
+param([Parameter(Mandatory=$true)] $version)
 
-tsc --declaration --emitDeclarationOnly --skipLibCheck --outFile ../Tools/npm/dotvvm-types/index.d.ts
+$dir = pwd
+push-location ../../../src/Framework/Framework
+
+$packageJson = Get-Content package.json | ConvertFrom-Json
+$typescriptVersion = $packageJson.devDependencies.typescript
```
exyi, line 7: this is also not needed anymore

exyi (approved):

quigamdev (approved):
No big issue found. Just please resolve the comments.

PR #1384, opened 2022-05-17
Fixed ToEnumString to work with flags

tomasherceg:


src/Framework/Framework/Utils/ReflectionUtils.cs
```diff
@@ -474,7 +474,18 @@ public static string GetTypeHash(this Type type)
 
         public static string ToEnumString<T>(this T instance) where T : Enum
         {
-            return ToEnumString(instance.GetType(), instance.ToString());
+            if (typeof(T).GetCustomAttribute<FlagsAttribute>() != null)
```
acizmarik, line 477: We should keep the behaviour used by Newtonsoft. Either using `JsonConvert` or fix the implementation to match their behaviour

exyi (approved):

PR #1385, opened 2022-05-17
Added support for more complex expressions to `Validator.Value`

acizmarik:
This PR adds support for complex expressions to `Validator.Value`. I am not even sure, whether this is a new feature or a fix. 

In BusinessPack, controls automatically set their `Validator.Value` according to the passed value binding (e.g. `DatePicker`'s `SelectedDate="{value: MyDateProperty}"` generates knockout binding `dotvvm-validation` with value `MyDateProperty`). However, this breaks once user tries to specify something more complex, for example `MyDateProperty.ToBrowserLocalTime().ToString("yyyy")`.

Therefore, this PR implements an unwrapping mechanism that tries to unwrap bindings so that errors can be propagated to the correct observable.


acizmarik:
There is one more place where I need to fix this. `ISortingOptions.SortExpression` can be populated with `DateTimeProperty.ToBrowserLocalTime()` when user tries to sort by specifying `AllowSorting="true"` and clicking on the column within a grid.

image

However, it is going to be separate PR as it is handled a bit differently there

src/Framework/Framework/Binding/Expressions/ValueBindingExpression.cs
```diff
@@ -95,7 +101,11 @@ public class OptionsAttribute : BindingCompilationOptionsAttribute
                         return new GlobalizeResourceBindingProperty();
                     }
                     return null;
-                })
+                }),
+                new Func<ValueBindingExpression, ParsedExpressionBindingProperty, ReferencedViewModelPropertiesBindingProperty>((binding, expr) =>
+                {
+                    return BindingPropertyResolvers.GetReferencedViewModelProperties(binding, expr);
+                }),
```
exyi, line 108: I think this is not needed, it should automatically discover the `GetReferencedViewModelProperties` method

src/Framework/Framework/Binding/Expressions/ValueBindingExpression.cs
```diff
@@ -49,6 +53,8 @@ private protected override void StoreProperty(object p)
         {
             if (type == typeof(KnockoutExpressionBindingProperty))
                 return knockoutExpressions.GetValue(this).GetValue(errorMode, this, type);
+            if (type == typeof(ReferencedViewModelPropertiesBindingProperty))
+                return referencedPropertyExpressions.GetValue(this).GetValue(errorMode, this, type);
```
exyi, line 57: It's not wrong, but this storing properties in fields instead of the ConcurrentDictionary is just an optimization for the most used binding properties. This one is probably not going to be used that heavily (or?), but also not super rarely, so whatever. Just noting it so you know that you don't have to do this if you find yourself adding another binding property in the future.

src/Framework/Framework/Compilation/Binding/GeneralBindingPropertyResolvers.cs
```diff
@@ -429,5 +430,47 @@ public IsMoreThanZeroBindingProperty IsMoreThanZero(ParsedExpressionBindingPrope
...
+                // unwrap type conversions, negations, ...
+                if (expr is UnaryExpression unary)
+                    expr = unary.Operand;
+                // unwrap some method invocations
+                else if (expr is MethodCallExpression boxCall && boxCall.Method.DeclaringType == typeof(BoxingUtils))
+                    expr = boxCall.Arguments.First();
+                else if (expr is MethodCallExpression { Method.Name: nameof(DateTimeExtensions.ToBrowserLocalTime) } dtMethodCall && dtMethodCall.Method.DeclaringType == typeof(DateTimeExtensions))
+                    expr = dtMethodCall.Object ?? dtMethodCall.Arguments.First();
+                else if (expr is MethodCallExpression { Method.Name: nameof(object.ToString) } toStringMethodCall)
+                    expr = toStringMethodCall.Object ?? toStringMethodCall.Arguments.First();
```
exyi, line 457: I'd also unwrap the ToEnumString method which we now use for enum->string implicit conversion by default

src/Framework/Framework/Controls/KnockoutHelper.cs
```diff
@@ -38,6 +34,22 @@ public static class KnockoutHelper
             }
         }
 
+        /// <summary>
+        /// Adds the data-bind attribute to the next HTML element that is being rendered. If in server rendering mode, the binding is also not rendered.
+        /// </summary>
+        /// <param name="renderEvenInServerRenderingMode">When set to true, the binding is rendered even in server rendering mode. By default, the data-bind attribute is only added in client-rendering mode.</param>
+        public static void AddKnockoutDataBind(this IHtmlWriter writer, string name, DotvvmBindableObject control, IValueBinding expression, string? valueUpdate = null, bool renderEvenInServerRenderingMode = false)
+        {
+            if (!control.RenderOnServer || renderEvenInServerRenderingMode)
```
exyi, line 43: I'd just make the `renderEvenInServerRenderingMode` the default. The signature of this method already takes in the `IValueBinding` expression, so I'd expect that it will just render it regardless the RenderSettings. Also we are gradually moving away from the "server rendering disables knockout expression", as users should prefer to use resource bindings for it.

src/Framework/Framework/Controls/Validator.cs
```diff
@@ -83,7 +86,22 @@ public class Validator : HtmlGenericControl
...
+                var referencedPropertyExpressions = binding.GetProperty<ReferencedViewModelPropertiesBindingProperty>();
+                var unwrappedPropertyExpression = referencedPropertyExpressions.UnwrappedBindingExpression;
+
+                // We were able to unwrap the the provided expression
+                writer.AddKnockoutDataBind(validationDataBindName, control, unwrappedPropertyExpression, renderEvenInServerRenderingMode: true);
+            }
+            else
+            {
+                // We are passing the binding "as-is"
+                writer.AddKnockoutDataBind(validationDataBindName, control, ValueProperty, renderEvenInServerRenderingMode: true);
```
exyi, line 103: I think it won't really work without the value binding, right? I'd just throw an exception here, unless there exists a scenario when validation without Validator.Value works...

src/Framework/Framework/Controls/KnockoutHelper.cs
```diff
@@ -38,6 +34,18 @@ public static class KnockoutHelper
             }
         }
 
+        /// <summary>
+        /// Adds the data-bind attribute to the next HTML element that is being rendered. If in server rendering mode, the binding is also not rendered.
```
exyi, line 38:
```suggestion
        /// Adds the data-bind attribute to the next HTML element that is being rendered.
```

 ;)

exyi (approved):

PR #1388, opened 2022-05-20
Added support for primary names and alternative names for controls

tomasherceg:
I've added an option to specify aliases for controls in markup. 
We'll need this in the AutoUI package where we want the control in markup would be named `auto:Editor`, but the underlying class name would be `AutoEditor`.

I've added the `PrimaryName` property in the `ControlMarkupOptions` - this will be important for the VS extension as this should be the name that the extension will suggest in IntelliSense. If this property is not specified, the primary name would be the name of the control class (current behavior).
Also, I've added the `AlternativeName` collection which allows specifying other aliases. This can be used for renaming controls.

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/DefaultControlResolver.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/DefaultControlResolver.cs
@@ -28,6 +28,7 @@ namespace DotVVM.Framework.Compilation.ControlTree
         private static object dotvvmLocker = new object();
         private static bool isDotvvmInitialized = false;
 
+        private static readonly Dictionary<string, string> controlNameMappings = new(StringComparer.OrdinalIgnoreCase);
 
         public DefaultControlResolver(DotvvmConfiguration configuration, IControlBuilderFactory controlBuilderFactory, CompiledAssemblyCache compiledAssemblyCache) : base(configuration.Markup)
         {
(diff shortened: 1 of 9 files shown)


tomasherceg:
I've found one issue here - the assembly name can be just the name or a fully qualified name. 
I am not sure what solution will be better - we can either make all assembly names in the configuration resolved when the configuration is getting frozen, or we can write the comparison so only the assembly name is being reflected.

src/Framework/Framework/Controls/ControlMarkupOptionsAttribute.cs
```diff
@@ -10,5 +10,32 @@ public class ControlMarkupOptionsAttribute : Attribute
...
+        /// </summary>
+        public string? PrimaryName { get; set; } = null;
+
+        /// <summary>
+        /// Represents a set of alternative names that are possible to use in the markup.
+        /// </summary>
+        public string[]? AlternativeNames { get; set; } = null;
+    }
+
+    public enum ControlPrecompilationMode
```
exyi, line 28: I think this shoudn't be here, but no big deal

src/Framework/Framework/Compilation/ControlTree/DefaultControlResolver.cs
```diff
@@ -45,6 +46,7 @@ public DefaultControlResolver(DotvvmConfiguration configuration, IControlBuilder
                         startupTracer?.TraceEvent(StartupTracingConstants.InvokeAllStaticConstructorsStarted);
                         InvokeStaticConstructorsOnAllControls();
                         ResolveAllPropertyAliases();
+                        ResolveAllControlAliases();
```
exyi, line 49: I'd slightly prefer if this method returned the `controlNameMappings` dictionary. It would be clearer that nobody can access it before it's actually initialized, or while it's being initialized.

exyi (approved):

PR #1390, opened 2022-05-21
Fix a race condition in capability registration

### LINKED ISSUE #1387: Duplicate key added to Dictionary in LoadProperties
We've probably found a bug in a control resolver. The weird thing is that it appears only sometimes.
It happens when the app is deployed in Azure App Service (running on a slow 2-CPU VM with a slow filesystem). When the app is freshly deployed, the issue doesn't occur and the application starts correctly. 
But when the app gets restarted, we are getting the following exception during startup:

```
DotVVM.Framework.Compilation.DotvvmCompilationException
---> System.ArgumentException: An item with the same key has already been added. Key: ID
at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
at DotVVM.Framework.Compilation.ControlTree.ControlResolverMetadata.LoadProperties(Dictionary`2 result)
at DotVVM.Framework.Compilation.ControlTree.ControlResolverMetadataBase.<.ctor>b__35_0()
at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
```

I've added detailed logging in the [`fix/loadproperties-error-message-4.0` branch](https://github.com/riganti/dotvvm/blob/fix/loadproperties-error-message-4.0/src/Framework/Framework/Compilation/ControlTree/ControlResolverMetadata.cs#L51) and the problem is with `HtmlGenericControl.ID` and `DotvvmControl.ID` properties:

```
The property with name ID (declared in DotVVM.Framework.Controls.HtmlGenericControl) was already registered for control DotVVM.Framework.Controls.HtmlGenericControl. Conflicting property: ID (declared in DotVVM.Framework.Controls.DotvvmControl).
```

In order to make sure the issue is not caused by Azure (it happens mostly after restart of the application so theoretically, some DLLs may stay in memory or be loaded twice), we've changed the ASP.NET Core hosting model to _out of process_ - this should kill the process and start it again. But I guess it's not the case.

We are not able to reproduce the issue locally. 
The weird thing is that HtmlGenericControl itself doesn't have the ID property declared - it's only declared in the HtmlCapability.

exyi:
The problem occured when HtmlGenericControl static constructor
was invoked before DotvvmControl
   - DotvvmControl.Id was not registered
   - HtmlGenericControl contains HtmlCapability which contains Id property
   - normally, the HtmlCapability.Id is mapped to the DotvvmControl.Id property
   - since DotvvmControl wasn't initialized, another Id property got created, leading to a crash later

This fix makes sure that base type static constructor is always invoked before attemting to register a capability or a CompositeControl property.

resolves #1387


acizmarik (approved):
Please fix the problem with nullability. After that I think we can merge

PR #1393, opened 2022-05-22
Refactored markup file loader to be extensible by third-party libraries

tomasherceg:


exyi:
What's the difference between this and  #1312? I remember we discussed it would be better to register the list of loaders in the IOptions instead of list of types which are then somehow automagically created

PR #1409, opened 2022-06-08
Use case insensitive comparisons for attributes

### LINKED ISSUE #1395: Change to case-insensitive comparisons
Changes in #1365

exyi:
Use OrdinalIgnoreCase comparer for all PropertyGroups - in general most things
in dothtml is case insensitive, all of our uses for property groups are
case insensitive. I think it's not worth it to add another config option.

resolves #1395


acizmarik (approved):
👍

PR #1412, opened 2022-06-11
deserializer: fix root viewmodels with constructor properties

exyi:
Problem was that we re-created the root viewmodel without setting it back to context.ViewModel and view.DataContext

cc @Ropack


quigamdev (approved):
👍

PR #1413, opened 2022-06-11
deserializer: don't touch getters irrelevant for deserialization

exyi:
In some convoluted scenarios when users had incomplete objects
in the viewmodel before deserialization,
we'd touch all property in order to copy them to
a potential new instance.
This could cause NullReferenceException, if the getters
computed something from missing nested objects.

cc @Ropack


quigamdev (approved):
👍

PR #1419, opened 2022-06-16
Fix empty templates on .NET old framework

exyi:


acizmarik (approved):
Thanks. I just tested this on .NET Framework and the issue is resolved

PR #1420, opened 2022-06-16
Fix race conditions in startup tracer

exyi:
hopefully without adding new ones...
the main point is the replacement of List with ImmutableList,
which is safe to read even when some other thread
is still reporting events

### DIFF
--- a/src/Framework/Framework/Diagnostics/DiagnosticsStartupTracer.cs
+++ b/src/Framework/Framework/Diagnostics/DiagnosticsStartupTracer.cs
@@ -18,7 +18,6 @@ namespace DotVVM.Framework.Diagnostics
 
         // list with concurrent read capability
         private ImmutableList<EventTiming> events = ImmutableList<EventTiming>.Empty;
-        private int startupCompleted;
 
         private long ElapsedMillisecondsSinceLastLog => events.Sum(e => e.Duration);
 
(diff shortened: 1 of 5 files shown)


src/Framework/Framework/Utils/ConcurrencyUtils.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Immutable;
+using System.Threading;
+
+namespace DotVVM.Framework.Utils
+{
+    static class ConcurrencyUtils
+	{
+		public static T CasChange<T>(ref T xRef, Func<T, T> f)
+			where T : class
+		{
```
acizmarik, line 11: IMO you can rewrite this method without the `goto` and `retry` label in a bit more cleaner way.

```csharp
T oldValue, newValue;

do
{
    oldValue = xRef;
    newValue = f(oldValue);
}
while (Interlocked.CompareExchange(ref xRef, newValue, oldValue) != oldValue);

return newValue;
```

src/Framework/Framework/Diagnostics/DiagnosticsStartupTracer.cs
```diff
@@ -30,34 +32,27 @@ public void TraceEvent(string eventName)
...
-            bool reportLateEvent;
-            lock (locker)
-            {
-                events.Add(eventTiming);
-                reportLateEvent = startupCompleted;
-            }
+            ConcurrencyUtils.CasChange(ref events, l => l.Add(eventTiming));
+            var reportLateEvent = startupCompleted > 0;
 
             if (reportLateEvent)
```
acizmarik, line 38: Technically, this is a race. Although, I am not sure how does this part of the framework works. It could be benign.

* T1: on L36 evaluates `reportLateEvent`
* T2: on L47 increments `startupCompleted`
* T1 then either executes this condition or not - depending on how the previous operations are scheduled.

exyi: You are right, should be gone now (there is only one variable which we need to synchronize)

acizmarik (approved):

PR #1421, opened 2022-06-16
Revert "Use case insensitive comparisons for attributes"

acizmarik:
We found out that this feature is not stable. We are not sure how to fix it


src/Framework/Framework/Controls/HtmlGenericControl.cs
```diff
@@ -413,7 +413,7 @@ private void AddHtmlAttributesToRender(ref RenderState r, IHtmlWriter writer)
 
                 if (knockoutExpression is {})
                 {
-                    if (attributeName.Equals("class", StringComparison.OrdinalIgnoreCase))
+                    if (attributeName == "class")
```
exyi, line 416: Maybe this line does not need to be reverted., otherwise it's fine

exyi (approved):

src/Framework/Framework/Controls/HtmlGenericControl.cs
acizmarik: You are right. I did not realize this - just performed a `git revert` and did not realize that this should definitely stay

PR #1424, opened 2022-06-20
Added support for running unit tests on .NET Framework

acizmarik:
This PR adds the possibility to run tests also on .NET Framework. Over the past few months, there have been a couple of issues manifesting only on .NET Framework. Therefore, we should always run unit tests also on .NET Framework as it seems to have discovered multiple issues (see the tests report).

- This change should not affect people on Linux/Mac as we are using conditional `TargetFrameworks`
- It also works on CI (`windows-2022` run has two reporting platforms - `net6` and `net472`)

Failing tests on .NET Framework will be fixed within a separate PR


acizmarik:
I will mark this as available for review once #1421 is merged

exyi (approved):

PR #1433, opened 2022-07-11
Updated Newtonsoft.Json

acizmarik:


exyi:
Do we really need to update the minimal required version of DotVVM.Framework and DotVVM.Core?

acizmarik:
We could keep the version. But then we should set the maximum recursion depth for JSON parsing. Not sure which approach is better for us

exyi:
I'd set the recursion limit, updating 3 version may break some people so they will avoid updating to 4.1... Setting the limit could also be released as 4.0.x

acizmarik:
Closing in favor of #1434

PR #1437, opened 2022-07-27
Backported fix 1385

acizmarik:
This PR backports a fix #1385 into 4.0.x.


acizmarik:
So... It turns out that I also needed to cherry-pick these: #1307, #1384, #1405

acizmarik:
This is not going anywhere... PR depends on too many changes not present in 4.0.x

PR #1441, opened 2022-08-04
markup control: Basic support for property groups

exyi:
If the property group is used in the markup,
it will be serialized among the other control properties as a dictionary.
This approach should allow any operations on the dictionary.

The main use case is for passing all the properties into a JS function


Mylan719 (commented):
one syntax sugar and serialization feedback

Mylan719 (approved):

PR #1443, opened 2022-08-04
Compile time warning for uppercase atrributes (i.e. mistyped properties)

exyi:
We have this warning at runtime, so people can see it in the dev console,
or in ASP.NET Core logs. With this change, the error should be also
displayed in VS Extension (we currently don't display warnings with DotVVM itself)


acizmarik (approved):
Very nice!

Issue #1438, opened 2022-07-29
Date/DateTime Textbox value resets if user changes year


Mylan719:
- I am trying to use: `<dot:TextBox Type="Date" Text={value: EndDate} />`
- I bind it to `public DateTime StartDate { get; set; }` 
- I type `10`,`10` for month and day into the date field like this: 
![image](https://user-images.githubusercontent.com/3776116/181761167-b5225a13-0835-4daa-8e1a-c302e2b195a8.png)
- If I now type `20` to write year 2022 the value resets to empty.
Another way to get this behavior is to go to filled in date select year and start typing.

With ordinary `<input type='date'>` this does not happen and I can fill the year no problem.

exyi:
This is solved by #1451

PR #1451, opened 2022-08-10
Resetting half written inputs

Mylan719:
Is part of solution for #1438 
We probably should not reset value of input if element has invalid value inside. It may be half filled date/number


Mylan719:
There might be better way to do it tho.

exyi (approved):
:+1:  Great, I think (or hope :sweat_smile:) this will fix most problems

PR #1454, opened 2022-08-14
Include a list of controls in the serialized config

### LINKED ISSUE #1431: Styles.Tag not known by VS Extension
Hi,

I'm on latest version and I would like to use `Styles.Tag` property but I'm getting warning that this property does not exist.

image

I found it here: https://www.dotvvm.com/docs/4.0/pages/concepts/dothtml-markup/server-side-styles

exyi:
I was considering moving the list of properties into each control,
but that would exclude attached properties from the list

resolves #1431


acizmarik:
https://github.com/riganti/dotvvm/blob/c6398833bf6b4d0df730614d2368ec0fa88d1898/src/Tests/Runtime/config-tests/ConfigurationSerializationTests.Markup.json#L51

There are now `FullName`, `CSharpName` and `CSharpFullName` properties. Some of them can be probably merged or something

acizmarik (approved):
Nice. Until release of 4.1 we should also resolve #1497

PR #1455, opened 2022-08-18
ListBox multiple item selection

Mylan719:
Currently in DotVVM there is no support for native HTML input `<select multiple>...</select>`
We do have `<dot:ListBox>` control but it does not support multiple selected items. 
ASP.NET Web Forms has `<ListBox SelectionMode="Multiple">` that can be used in this case.

I implemented  `SelectionMode` property for our `<dot:ListBox>`.


src/Framework/Framework/Controls/ListBox.cs
```diff
@@ -32,16 +60,54 @@ protected override void AddAttributesToRender(IHtmlWriter writer, IDotvvmRequest
...
+            {
+                writer.AddKnockoutDataBind("value", this, SelectedValueProperty, renderEvenInServerRenderingMode: true);
+            }
+            writer.AddKnockoutDataBind("valueAllowUnset", "true");
         }
 
-        public static readonly DotvvmProperty SizeProperty =
-            DotvvmProperty.Register<int, ListBox>(t => t.Size, defaultValue: 10);
+        [ControlUsageValidator(Override = true)]
+        public static new IEnumerable<ControlUsageError> ValidateUsage(ResolvedControl control)
```
tomasherceg, line 78: Can you please add some tests for the usage validation? 
I am not sure what will happen if the SelectionMode=Multiple and the SelectedValues is not IEnumerable - the GetEnumerableType will probably throw an exception so this method will fail to.

tomasherceg (changes requested):

src/Framework/Framework/Controls/ListBox.cs
Mylan719: good catch you are right

Mylan719: Done ;)

acizmarik:
We should reuse some parts of this implementation (ValidateUsage and tests for example).

PR #1456, opened 2022-08-21
Fixes in TypeScript declarations package

tomasherceg:
I've found several issues when using the TypeScript package:

1. Since we don't publish the types in the official `@types` organization, the user must set up the `typeRoots` section in the `tsconfig.json` like this:

```
{
  "compilerOptions": {
    ...
    "typeRoots": [
      "node_modules/@types",
      "node_modules/dotvvm-types"
    ]
  }
}
```

The problem here is that TypeScript looks only into subdirectories of `typeRoots`, so it didn't see the `node_modules/dotvvm_types/index.d.ts` file (it is necessary to see the `dotvvm` global object - it's not a concern when importing DotVVM as a module). 
When I moved it to a subfolder, it started working.

2. Another problem was the `/// <reference path="globalize.d.ts" />` - I removed it because it was not necessary and produced a warning when someone wanted to use the declarations package.

3. I've noticed that we have old Knockout declarations in the project. I've removed them and replaced them with the `@types/knockout` package. 

4. I've upgraded TypeScript and fixed about 20 warnings produced by `npm run tsc-check`.

5. I extracted all customizations of Knockout API to a separate declarations file `knockout.dotvvm.d.ts`.

**I am just wondering whether we shall make the `state`, `setState`, and `patchState` methods optional on the `KnockoutObservable<T>` interface.** They don't have to be there, but DotVVM adds them everywhere it can, and when the user uses `dotvvm.viewModels.root.viewModel.something` (which is a frequent scenario), they won't see these methods unless they will explicitly check for their presence.


src/Framework/Framework/Resources/Scripts/typings/knockout/knockout.dotvvm.d.ts
```diff
@@ -7,3 +7,21 @@ interface KnockoutDelaySync {
...
+}
+interface KnockoutObservableArrayStatic {
+    <T>(value: T[] | null, validator: (newValue: T[]) => void): KnockoutObservableArray<T>;
+}
+
+interface KnockoutObservable<T> {
+    readonly state?: T;
+    patchState?(patch: any): void;
+    setState?(newState: any): void;
+    readonly updater?: UpdateDispatcher<T>;
```
exyi, line 26: Using `T` in this definition is technically wrong, since T will have inner properties wrapped in knockout observables, while `state` and `updater` will be fully unwrapped.

We already have `DotvvmObservable<T>` which handles this in the "opposite" way - it wraps all properties of `T` in observables. 

I think it would be better to adjust methods like `ko.dataFor` and `ko.contextFor` to return object with `DotvvmObservable`s.

exyi (approved):

PR #1457, opened 2022-08-24
Fixed bug in RenderSettings.Mode=Server when a collection is set to null value

tomasherceg:
When a collection in the viewmodel was set to null, and there was a server-side rendered control bound to it, it was throwing an error in the `dotvvm-SSR-foreach` binding handler.
I've added a check for this situation and also added UI tests.


acizmarik (approved):
👍

PR #1460, opened 2022-09-01
Disable static value in IBinding property

VaclavKosak:
Disable static value in IBinding property


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/riganti/dotvvm?pullRequest=1460) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/riganti/dotvvm?pullRequest=1460) before we can accept your contribution.<br/><br/>:x: VaclavKosak [sign now](https://cla.dotnetfoundation.org/riganti/dotvvm?pullRequest=1460)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/riganti/dotvvm?pullRequest=1460) it.</sub>

exyi (approved):
good catch.

I guess it crashed anyway without this check when you used the hardcoded value? So we don't need to be afraid that it breaks any code.

tomasherceg:
Yes, it would crash later, but we need the metadata to be correct because we use them to generate the Properties table in the docs: https://www.dotvvm.com/docs/4.0/controls/builtin/ComboBox#Properties

For example, the ItemTextBinding states it supports static value, which is wrong. This PR should fix it.

And I suspect that a similar problem would occur in the VS Code extension if the serialized metadata is used there.

PR #1470, opened 2022-09-14
Improve generated code in capability property accessors

exyi:
This patch should fix number of problems with capability properties:
* it couldn't handle weird property types (fallback),
* it couldn't handle inherited properties correctly
* yet, it was using DotvvmBindableObject.GetValue for many properties, which is unnecesarily slow
* setting a big capability with 1 property set will set all the other properties to their default value - allocating unnecesary memory

Problem was that we was sometimes using direct access to the properties dictionary, and sometimes we were using GetValue (and friends). This patch decides for each property if it's ok to use direct access or not (i.e. it does not override GetValue & is not inherited), and then acts accordingly.

### DIFF
--- a/.github/setup/action.yml
+++ b/.github/setup/action.yml
@@ -28,9 +28,15 @@ runs:
   # .NET
   - uses: actions/setup-dotnet@v3
     with:
-      dotnet-version: 7.0.100
+      dotnet-version: |
+        7.0.x
+        6.0.x
+        3.1.x
   - if: ${{ runner.os == 'Windows' }}
     uses: microsoft/setup-msbuild@v1.1
+  - if: ${{ runner.os == 'Windows' }}
+    run: choco install dotnetcore-3.1-windowshosting -y
+    shell: pwsh
 
   # restore packages
   - if: ${{ runner.os == 'Windows' }}
--- a/.github/test-report/action.yml
+++ b/.github/test-report/action.yml
@@ -21,6 +21,10 @@ inputs:
 runs:
   using: composite
   steps:
+    - uses: actions/upload-artifact@v3
+      with:
+        name: ${{ inputs.report-name }}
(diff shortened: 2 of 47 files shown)


src/Framework/Framework/Binding/DotvvmCapabilityProperty.CodeGeneration.cs
```diff
@@ -109,22 +109,39 @@ public static (LambdaExpression getter, LambdaExpression setter) CreatePropertyG
...
             {
+                if (property is DotvvmPropertyAlias propertyAlias)
+                    return CreatePropertyAccessors(type, propertyAlias.Aliased);
+
+                // if the property does not override GetValue/SetValue, we'll use
+                // control.properties dictionary directly to avoid virtual method calls
+                var canUseDirectAccess =
+                    !property.IsValueInherited && (
+                    property.GetType() == typeof(DotvvmProperty) ||
+                    property.GetType().GetMethod("GetValue", new [] { typeof(DotvvmBindableObject), typeof(bool) }).DeclaringType == typeof(DotvvmProperty) &&
```
acizmarik, line 120: Maybe change for `nameof(DotvvmBindableObject.GetValue)`

acizmarik (commented):
It would be good to check whether this behaviour is tested or not

exyi:
I added the tests (and fixed few found issues :] ), @acizmarik would you have a look again?

src/Tests/Runtime/CapabilityPropertyTests.cs
```diff
@@ -140,6 +223,51 @@ public class TestControl5:
             IObjectWithCapability<TestNestedCapabilityWithPrefix>
         {       
         }
+        // public class TestControl6:
```
tomasherceg, line 226: Please remove

tomasherceg (approved):

PR #1471, opened 2022-09-15
Added support for `DateOnly` and `TimeOnly` types

acizmarik:
This PR adds support for newly added types in .NET 6

- In order to support these on .NET Framework, `Portable.System.DateTimeOnly` NuGet is used
- Formally, I did some breaking changes in `ReflectionUtils`, though I don't think that people can be affected. Moreover, it was theoretically possible to change our internal collections

This feature can be helpful for BusinessPack as well. Currently there are controls, such as `bp:DatePicker` that could benefit from supporting these types. Right now there are some weird edge-cases (for example using `.ToBrowserLocalTime()` can shift the selected date on `bp:DatePicker`). `DateOnly` and `TimeOnly` are not going to support conversions.


acizmarik:
I also found an issue that has been in the framework for some time probably. Whenever there is a value binding used within a control or enclosed, for example in a span, `data-bind` is generated. However, once a value binding is placed directly in body, the value is rendered using `<!-- ko text: .... -->`. This ignores any formatting, see an example below: (rendered plainly is using the default formatting)

image

acizmarik (commented):
We should refactor changes is `date.js` and maybe try and use `Date.parse` builtin function

src/Framework/Framework/Resources/Scripts/serialization/date.ts
```diff
@@ -35,6 +35,14 @@ export function parseDate(value: string | null, convertFromUtc: boolean = false)
     return null;
 }
 
+export function parseDateOnly(value: string | null): Date | null {
+    return parseDate(`${value}T00:00:00.00`, false);
+}
+
+export function parseTimeOnly(value: string | null): Date | null {
+    return parseDate(`0000-00-00T${value}`, false);
```
exyi, line 43: `0000-00-00` is not exactly a valid date :) I'd use `1970-01-01`

src/Framework/Framework/Resources/Scripts/serialization/date.ts
```diff
@@ -96,6 +104,41 @@ export function serializeDate(date: string | Date | null, convertToUtc: boolean
     return `${y}-${m}-${d}T${h}:${mi}:${s}.${ms}0000`;
 }
 
+export function serializeDateOnly(date: string | Date | null): string | null {
```
exyi, line 107: As far as I can tell, it's only called with parameter of type Date, so wouldn't bother with validating the `string`

exyi (approved):

PR #1472, opened 2022-09-16
Added missing GetDotvvmContext method for ASP.NET Core

tomasherceg:
We already had this method implemented for OWIN but not for ASP.NET Core.
Also, I have removed the word `Owin` from the `DotvvmRequestContextOwinKey` as it was used in ASP.NET Core version too.

### DIFF
--- a/src/Framework/Hosting.AspNetCore/Hosting/AspNetCoreDotvvmRequestContextExtensions.cs
+++ b/src/Framework/Hosting.AspNetCore/Hosting/AspNetCoreDotvvmRequestContextExtensions.cs
@@ -35,5 +35,14 @@ namespace DotVVM.Framework.Hosting
             }
             return new AuthenticationManager(concreteContext.OriginalContext);
         }
+
+
+        /// <summary>
+        /// Gets the <see cref="IDotvvmRequestContext"/> bound to the specified <see cref="HttpContext"/>.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/Framework/Hosting.AspNetCore/Hosting/AspNetCoreDotvvmRequestContextExtensions.cs
```diff
@@ -35,5 +35,14 @@ public static AuthenticationManager GetAuthentication(this IDotvvmRequestContext
             }
             return new AuthenticationManager(concreteContext.OriginalContext);
         }
+
+
+        /// <summary>
+        /// Gets the <see cref="IDotvvmRequestContext"/> bound to the specified <see cref="HttpContext"/>.
+        /// </summary>
+        public static IDotvvmRequestContext? GetDotvvmContext(this HttpContext httpContext)
+        {
```
exyi, line 44: :thinking: We forgot to add null annotations to this package. Anyway, we should add it later, not in this PR

exyi (approved):

src/Framework/Hosting.AspNetCore/Hosting/AspNetCoreDotvvmRequestContextExtensions.cs
acizmarik: Removed the annotation, we can do the refactoring later

PR #1473, opened 2022-09-19
Added sample for _js.Invoke in value binding.

Mylan719:
I realized that `_js.Invoke` does not work in value bindings in markup control. It works nicely in the page. I added sample for problematic case.


Mylan719 (commented):
This approach seems nicer and more general.

cafour (approved):
I am no expert on the translation of bindings to JS but this looks alright to me.

acizmarik:
Can we add an UI test for this please? After that I think we can merge

Mylan719:
I added new more complicated sample and UI tests for both samples.

exyi:
Awesome, thank you

PR #1480, opened 2022-09-28
property directive: cache the generated type

### LINKED ISSUE #1474: Hot reload breaks markup controls with @property directive
When a markup control with the `@property` directive is used, and its file is changed with the Hot reload turned on, we are probably not deleting some caches - I am getting the following exception:

```
Unable to cast object of type 'DotvvmMarkupControl-690dd7db-a01e-41a5-9897-b4bfda8c472d' to type 'DotvvmMarkupControl-4f6bb064-cc07-4b26-9b8d-9534a0433f50'.
```

exyi:
this should avoid changing the type when only body of the markup control changes. Should mostly fix  #1474

### DIFF
--- a/src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
+++ b/src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
@@ -11,6 +11,8 @@ using DotVVM.Framework.Controls.Infrastructure;
 using DotVVM.Framework.Utils;
 using System.Linq;
 using System.Collections.Immutable;
+using System.Security.Cryptography;
+using System.Text;
 
 namespace DotVVM.Framework.Compilation.Directives
 {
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
```diff
@@ -63,22 +65,42 @@ protected override IAbstractBaseTypeDirective Resolve(DothtmlDirectiveNode direc
...
             {
-                wrapperType = CreateDymanicDeclaringType(wrapperType) ?? wrapperType;
+                wrapperType = CreateDynamicDeclaringType(wrapperType, propertyDirectives) ?? wrapperType;
             }
 
             return wrapperType;
         }
 
-        protected virtual ITypeDescriptor? CreateDymanicDeclaringType(ITypeDescriptor? originalWrapperType)
+        /// <summary> Gets or creates dynamic declaring type based on the  </summary>
```
acizmarik, line 76: Incomplete comment

acizmarik (approved):

PR #1481, opened 2022-10-09
Public API cleanup in Auto UI

tomasherceg:
I've reviewed entire code base of Dynamic Data to make sure there is no unused code and that the public API looks fine.

I've done just two tiny changes in the AutoUI options Fluent API:
* All methods start with verb - `Set`, `Add`, `Show`, `Hide`, `Enable` and so on.
* The naming tries to match the naming in attributes.
* `Hide()` is a shortcut for `SetAutoGenerateField(false)` - it is probably the most used use case so I've tried to make it easy for the user.


src/AutoUI/Core/Metadata/Builder/PropertyDisplayMetadataModifier.cs
```diff
@@ -51,12 +51,18 @@ public PropertyDisplayMetadataModifier SetNullDisplayText(string nullDisplayText
...
 
+    public PropertyDisplayMetadataModifier SetAutoGenerateField(bool value)
+    {
+        actions.Add(m => m.AutoGenerateField = value);
+        return this;
+    }
+
     public PropertyDisplayMetadataModifier ShowForViews(string viewNames)
     {
         actions.Add(m =>
```
exyi, line 68: ShowForViews and ShowForRoles should appends to the VisibleAttributes

src/AutoUI/Core/Metadata/Builder/PropertyDisplayMetadataModifier.cs
```diff
@@ -51,12 +51,18 @@ public PropertyDisplayMetadataModifier SetNullDisplayText(string nullDisplayText
         return this;
     }
 
-    public PropertyDisplayMetadataModifier Ignore()
+    public PropertyDisplayMetadataModifier Hide()
```
exyi, line 54: It would be nice to add doccomments with the equivalent attribute

exyi (approved):

src/AutoUI/Annotations/SelectionAttribute.cs
```diff
@@ -2,13 +2,21 @@
 
 namespace DotVVM.AutoUI.Annotations;
 
+/// <summary>
+/// Indicates that the user will select a value from a list of SelectionType provided by ISelectionProvider&lt;SelectionType&gt;
```
exyi, line 6:
```suggestion
/// Indicates that the user will select a value from a list of SelectionType provided by <see cref="ISelectionProvider{SelectionType}" />
```

You can use this to avoid these XML entities

PR #1484, opened 2022-10-13
Improvements of control property metadata for docs

tomasherceg:
I've added XML comments to builtin capabilities so they show up in the docs.
Also, I have updated the default mapping mode for properties of type control, collection of controls, or template. The user can override the default by placing his own `MarkupOptionsAttribute`.

### DIFF
--- a/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
+++ b/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
@@ -336,20 +336,22 @@ namespace DotVVM.Framework.Binding
                     dotvvmProperty = new DotvvmProperty(propertyName, type, declaringType, boxedDefaultValue, false, attributeProvider);
                     dotvvmProperty.OwningCapability = declaringCapability;
 
-                    var isNullable = propertyType.IsNullable() || propertyType.UnwrapValueOrBinding().IsNullable();
+                    if (!attributeProvider.IsDefined(typeof(MarkupOptionsAttribute), true))
+                    {
+                        var isNullable = propertyType.IsNullable() || propertyType.UnwrapValueOrBinding().IsNullable();
+                        if (!defaultValue.HasValue && !isNullable)
+                            dotvvmProperty.MarkupOptions.Required = true;
 
-                    if (!defaultValue.HasValue && !isNullable)
-                        dotvvmProperty.MarkupOptions.Required = true;
+                        if (typeof(IBinding).IsAssignableFrom(propertyType))
+                            dotvvmProperty.MarkupOptions.AllowHardCodedValue = false;
+                        else if (!typeof(ValueOrBinding).IsAssignableFrom(propertyType.UnwrapNullableType()))
+                            dotvvmProperty.MarkupOptions.AllowBinding = false;
 
-                    if (typeof(IBinding).IsAssignableFrom(propertyType))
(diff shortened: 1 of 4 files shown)


src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
```diff
@@ -336,20 +336,22 @@ internal static IControlAttributeDescriptor InitializeArgument(ICustomAttributeP
...
+                    if (!attributeProvider.IsDefined(typeof(MarkupOptionsAttribute), true))
+                    {
+                        var isNullable = propertyType.IsNullable() || propertyType.UnwrapValueOrBinding().IsNullable();
+                        if (!defaultValue.HasValue && !isNullable)
+                            dotvvmProperty.MarkupOptions.Required = true;
+
+                        if (typeof(IBinding).IsAssignableFrom(propertyType))
+                            dotvvmProperty.MarkupOptions.AllowHardCodedValue = false;
+                        else if (!typeof(ValueOrBinding).IsAssignableFrom(propertyType.UnwrapNullableType()))
+                            dotvvmProperty.MarkupOptions.AllowBinding = false;
```
exyi, line 348: This is problematic, if you add `MarkupOptions` to change the name or mapping mode, you'll make the property optional/bindable even when it must not be. Or you add it to make the property required, and suddenly it's not mapped as InnerElement

src/Framework/Framework/Controls/HtmlGenericControl.cs
```diff
@@ -597,11 +598,11 @@ public sealed record HtmlCapability
             Visible.HasValue && Visible.ValueOrDefault == true &&
             ID.HasValue && ID.ValueOrDefault == null;
 
-        /// <summary> A dictionary of css styles which will be placed in the `style` attribute. </summary>
+        /// <summary> Gets or sets a dictionary of CSS styles which will be placed in the `style` attribute. </summary>
         [PropertyGroup("Style-")]
         public IDictionary<string, ValueOrBinding<object?>> CssStyles { get; init; } = new Dictionary<string, ValueOrBinding<object?>>();
 
-        /// <summary> When set to false, `style="display: none"` will be added to this control. </summary>
+        /// <summary> Gets or sets whether the control is visible. When set to false, `style="display: none"` will be added to this control. </summary>
```
exyi, line 605: I'm not a fan of adding fluff to the comments. It's obvious that has getter, setter and controls if the control is visible (actually the comment is even partially wrong about the setter...)

exyi (dismissed):

src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
tomasherceg: So the `Required`, `AllowHardCodedValue`, and `AllowBinding` properties shall not be overridden by the attribute, while the other properties can be. I wonder whether it is intuitive, but I guess it's better than forcing the user to specify all these properties whenever they decide to use the attribute.

exyi: No it's not intuitive, but we don't have a way to tell if the user actually set the property, or only used the attribute for other reason... It's not great that when you apply `[MarkupOptions(Required=true)]`, it will disable MappingMode.InnerElement :facepalm:

exyi (approved):

PR #1486, opened 2022-10-16
Removed warning when a plain object is pushed in the observable array

tomasherceg:
When the user calls `observableArray.push(someItem)`, DotVVM was always producing a warning "Replacing old knockout observable with a new one, just because it is not created by DotVVM. Please do not assign objects into the knockout tree directly."

I believe the problem can occur only when the user pushes observable objects in the array but not when it is just a plain object.

I've also added tests for this warning.

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/state-manager.ts
+++ b/src/Framework/Framework/Resources/Scripts/state-manager.ts
@@ -327,7 +327,7 @@ function createWrappedObservable<T>(initialValue: DeepReadonly<T>, typeHint: Typ
                     if (newContents[index] && newContents[index][notifySymbol as any]) {
                         continue
                     }
-                    if (compileConstants.debug && newContents[index]) {
+                    if (compileConstants.debug && ko.isObservable(newContents[index])) {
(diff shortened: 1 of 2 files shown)


src/Framework/Framework/Resources/Scripts/state-manager.ts
```diff
@@ -327,7 +327,7 @@ function createWrappedObservable<T>(initialValue: DeepReadonly<T>, typeHint: Typ
                     if (newContents[index] && newContents[index][notifySymbol as any]) {
                         continue
                     }
-                    if (compileConstants.debug && newContents[index]) {
+                    if (compileConstants.debug && ko.isObservable(newContents[index])) {
```
exyi, line 330: I think we should warn when the object contains any observables. The problem is that DotVVM replaces all of them, so if your observables won't get updated.

tomasherceg:
I've moved the check into `unmapKnockoutObservable` and ensured that only one warning is emitted for the entire object.
Also, I've enhanced the tests to check that the warnings are produced only when appropriate.

acizmarik (approved):

PR #1488, opened 2022-10-18
Fixed issue with rejecting requests when restoring closed tabs on chrome

acizmarik:
This PR suggests relaxing our security checks, as right now one of them is too strict. Some customers started noticing this issue since 4.0 and I was able to replicate it as well on our samples using chrome.

1. Run samples and open any page (does not need to contain `SpaContentPlaceHolder`)
2. Close the tab
3. Restore the tab
4. Open developer tools

At this point a request is issued with `sec-fetch-destination`=`empty` and `sec-fetch-mode`=`no-cors`

### DIFF
--- a/src/Framework/Framework/Hosting/DotvvmPresenter.cs
+++ b/src/Framework/Framework/Hosting/DotvvmPresenter.cs
@@ -498,14 +498,17 @@ namespace DotVVM.Framework.Hosting
                 { // fine, this is allowed even cross-site
                 }
                 // if SPA is used, dest will be empty, since it's initiated from JS
-                // we only allow this with the X-DotVVM-SpaContentPlaceHolder header
+                // every SPA request has the X-DotVVM-SpaContentPlaceHolder header
                 // we "trust" the client - as if he lies about it being a SPA request,
                 // he'll will just get a redirect response, not anything useful
+                //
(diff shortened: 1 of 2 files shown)


exyi:
I guess we can allow this for SPA, but this really seems like a chromium bug, Shouldn't we report it? I couldn't find anything reported

This patch allows loading the page using JS, which is exactly what I originally wanted to prevent (as defense in depth against XSS)

acizmarik:
Honestly, I am not really familiar with these headers. If you feel like this is a bug, we should probably report it. I was just wondering that it is a little weird that we would be the only ones experiencing this issue.

I can maybe test also other browsers, I am wondering how this works on Firefox

exyi:
I can't reasonably replicate it even in Chromium, it happened like once in 100 attempts. In Firefox i didn't hit it at all, but maybe I just didn't try enough times.

acizmarik:
So my observations are following:
* Chrome: happens everytime
* Edge: happens everytime
* Firefox: could not reproduce

acizmarik:
Closing. The `CopyTo` was cherry-picked in #1547. 

The other change won't be merged as it effectively turns off the XSS protection. Instead we rewrote the warning to also notify about these issues in #1546

PR #1489, opened 2022-10-23
Removed usage of globalize from coercer

acizmarik:
This PR removes coercer's dependency on globalize library


src/Framework/Framework/Resources/Scripts/tests/coercer.test.ts
```diff
@@ -305,7 +305,7 @@ test("string - valid, converted from boolean", () => {
 test("string - valid, converted from Date", () => {
     const result = tryCoerce(new Date(2020, 0, 10, 12, 34, 56), "String");
```
exyi, line 306: You want to initialize the date as `new Date("2022-1-10T12:34:56Z")`

exyi (approved):

Issue #809, opened 2020-03-04
Static command don't resolve viewmodel correctly


martindybal:
```
public class SomeDependency
{
    public string SomeProperty { get; set; }
}

public class MasterViewModel : DotvvmViewModelBase
{
    private string dependencySomeProperty;

    public MasterViewModel(SomeDependency dependency)
    {
        //System.NullReferenceExceptionObject reference not set to an instance of an object.
        dependencySomeProperty = dependency.SomeProperty;
    }
}

public class PageViewModel : MasterViewModel
{
    public PageViewModel(SomeDependency dependency) : base(dependency)
    {
    }

    [AllowStaticCommand]
    public void StaticCommand()
    {
    }
}
```

Dependencies are `null` when static command is executed so I get NullReferenceException in  constructor. 

DotVVM version: 2.4.0.1

tomasherceg:
Calling instance methods on viewmodels from static commands is not supported. You should use [Static Command Services](https://www.dotvvm.com/docs/tutorials/basics-static-command-services/2.0).

Since the static commands don't send the viewmodel to the server, property values of the viewmodel are unknown and the object will be in a strange and probably inconsistent state. The same applies to the dependencies passed in the constructor - for historical reasons, they are set to null because it was implemented before DotVVM was using `IServiceProvider`. 
We could "fix" this by resolving them from the `IServiceProvider`, but still - the viewmodel will be in a weird state, the lifecycle methods won't be called, and the method may rely on data which are not there. 
We **should** throw an exception in this case - I think that the probability of using this feature by mistake is greater than the probability that someone is using it intentionally and is happy with how it behaves.

martindybal:
Oh, I see. I think the exception is best solution of this

exyi:
> dependencies passed in the constructor - for historical reasons, they are set to null because it was implemented before DotVVM was using IServiceProvider.

I don't think so, the arguments of a static command are simply deserialized using Newtonsoft.Json and, apparently it passes nulls to constructors (if a property of the same name does not exist).

This is how arguments are processed: https://github.com/riganti/dotvvm/blob/master/src/DotVVM.Framework/Hosting/DotvvmPresenter.cs#L332

exyi:
@tomasherceg What can we do about this? Shall we also use the dotvvm serializer static command arguments? (this is certainly a breaking change) Or call it a "feature"?

tomasherceg:
[...]

PR #1498, opened 2022-10-27
Support for custom primitive types

tomasherceg:
We've added support for registering custom primitive types.

There is one unresolved issue - the static command wasn't deserializing method arguments using converters from DotVVM. 
We've introduced this, but:
1. It is a breaking change.
2. The code needs to be refactored as the argument deserialization is in `DotvvmPresenter` that should not depend on `Newtonsoft.Json`. I think it should be moved to `IViewModelSerializer`, or maybe added as another interface.

Also, unit tests and UI tests are missing.


exyi:
> There is one unresolved issue - the static command wasn't deserializing method arguments using converters from DotVVM

Awesome, this is the subject of #809. We will have to leave it for 5.0, since this breaks quite many things. Or maybe put it behind a feature flag.

src/Framework/Framework/Configuration/DotvvmRuntimeConfiguration.cs
```diff
@@ -16,12 +16,21 @@ public class DotvvmRuntimeConfiguration
         public IList<IActionFilter> GlobalFilters => _globalFilters;
         private IList<IActionFilter> _globalFilters;
 
+        /// <summary>
+        /// Gets types that are treated as primitive by DotVVM components and serialization.
+        /// </summary>
+        [JsonIgnore()]
```
exyi, line 22: VS Extension will definitely need to know about this

src/Framework/Framework/Routing/UrlHelper.cs
```diff
@@ -30,7 +31,7 @@ public static string BuildUrlSuffix(string? urlSuffix, object? query)
                     }
                     break;
                 case IEnumerable<KeyValuePair<string, object>> keyValueCollection:
-                    foreach (var item in keyValueCollection)
+                    foreach (var item in keyValueCollection.Where(i => i.Value != null))
```
exyi, line 34: :+1: I guess we should do the same for the other two cases

src/Samples/Common/ViewModels/FeatureSamples/CustomPrimitiveTypes/SampleId.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace DotVVM.Samples.Common.ViewModels.FeatureSamples.CustomPrimitiveTypes
+{
+    public record SampleId : TypeId<SampleId>
+    {
+        public SampleId(Guid idValue) : base(idValue)
+        {
+        }
```
exyi, line 10: :thinking: 

We need to be able work with the id server-side and client-side in a similar way. Making the ID record solves comparison. We'd also need to override `ToString()` to be the same as `IdValue.ToString()`, otherwise `{{resource: Id}}` will behave differently than `{{value: Id}}` (even the value binding should evaluate to the same value server-side and client-side).

src/Framework/Framework/Resources/Scripts/validation/validation.ts
```diff
@@ -332,11 +332,15 @@ export function removeErrors(...paths: string[]) {
...
         const propertyPath = prop.propertyPath;
-        const property = evaluator.traverseContext(root, propertyPath);
+        try {
+            // find the property
+            const property = evaluator.traverseContext(root, propertyPath);
 
-        ValidationError.attach(prop.errorMessage, propertyPath, property);
+            ValidationError.attach(prop.errorMessage, propertyPath, property);
+        } catch (err) {
+            logWarning("validation", `Unable to find viewmodel property ${propertyPath}. If you've added this validation error using Context.AddModelError, make sure the property path is correct and that the object hasn't been removed from the viewmodel.`);
```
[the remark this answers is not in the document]
acizmarik: This will is resolved by #1500

src/Framework/Framework/Routing/UrlHelper.cs
acizmarik: This will be resolved in #1502

src/Framework/Framework/Configuration/CustomPrimitiveTypeAttribute.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace DotVVM.Framework.Configuration;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct)]
+public class CustomPrimitiveTypeAttribute : Attribute
+{
+    /// <summary>
+    /// Gets a type which will be used for this custom type on the client side. Only types from ReflectionUtils.PrimitiveTypes (or their nullable versions) are supported.
```
acizmarik, line 9: This type can not be actually nullable. We handle those differently

src/Framework/Framework/Utils/ReflectionUtils.cs
```diff
@@ -348,19 +358,37 @@ public static bool IsCollection(Type type)
...
+            {
+                throw new DotvvmConfigurationException($"The type {type} is marked with {nameof(CustomPrimitiveTypeAttribute)} but its {nameof(CustomPrimitiveTypeAttribute.ClientSideType)} is not valid. The client-side representation can only be one of the built-in primitive types (string, numbers, date and time types), or a nullable version of such type.");
+            }
+            if (!typeof(ICustomPrimitiveTypeConverter).IsAssignableFrom(attribute.ConverterType))
+            {
+                throw new DotvvmConfigurationException($"The type {type} is marked with {nameof(CustomPrimitiveTypeAttribute)} but its {nameof(CustomPrimitiveTypeAttribute.ConverterType)} doesn't implement the {nameof(ICustomPrimitiveTypeConverter)} interface!");
+            }
 
-            return PrimitiveTypes.Contains(type);
+            var converter = (ICustomPrimitiveTypeConverter)Activator.CreateInstance(attribute.ConverterType)!;
```
acizmarik, line 390: We should wrap this into a more descriptive exception. Right now if the converter can not be constructed using default ctor, the exception will be ugly

src/Framework/Framework/Configuration/CustomPrimitiveTypeAttribute.cs
tomasherceg: What is the problem with nullable types here? In the sample, we are using `Guid?` since the typed id is a reference type and can be null.

src/Framework/Framework/Configuration/DotvvmExperimentalFeaturesConfiguration.cs
```diff
@@ -17,11 +17,15 @@ public class DotvvmExperimentalFeaturesConfiguration
         [JsonProperty("explicitAssemblyLoading")]
         public DotvvmGlobalFeatureFlag ExplicitAssemblyLoading { get; private set; } = new DotvvmGlobalFeatureFlag();
 
+        [JsonProperty("useDotvvmSerializationForStaticCommandArguments")]
+        public DotvvmGlobalFeatureFlag UseDotvvmSerializationForStaticCommandArguments { get; private set; } = new DotvvmGlobalFeatureFlag();
```
tomasherceg, line 21: Rename to `UseDotvvmSerializationForStaticCommands`
Add XML comment
Will be true by default in the next release
Make sure it applies also for static command response

src/Framework/Framework/Configuration/CustomPrimitiveTypeAttribute.cs
acizmarik: This use-case seems to work nicely with reference types. If `SampleId`, which is a reference type, maps to `Guid?` - then properties that are of type `SampleId` work correctly.

The problem comes when `SampleId` is a value type (struct). Then we can have also a nullable and non-nullable property. This can generate funny stuff like this: 
image

acizmarik: We should probably add a check when generating metadata for value types so that such situation can not happen. Otherwise users should not registrate these with nullable underlying type, otherwise it can be doubled.

acizmarik:
We just discussed with @exyi that there is another potential issue - when user creates a custom primitive type and binds it into a textbox. How to treat this expression?

* We said that on client-side we would use the underlying types (which might have special formatting)
* However, expressions are generated on server from the custom type (that does not have special formatting by default)

I am not sure that this behaviour is clear or generally expected by users. What is the correct approach here?

acizmarik:
Furthermore, format overriding is not trivial (see our textbox binding handler 😄). The easiest solution would be IMO to format based on the underlying type

tomasherceg:
> user creates a custom primitive type and binds it into a textbox
For now, I would not support this. We'd have to provide `ToString(format)` implementation and JS translation, and somehow extend the globalize library to support formatting.
If these types don't work with `TextBox`, we should throw a compilation error.

tomasherceg:
The last thing we wanted to do is to change the `[CustomPrimitiveType]` attribute to a marker interface. Its presence is easier to detect without the need for caching.

exyi (approved):

PR #1499, opened 2022-11-01
Remove colors handed to console.log as arguments

Ropack:
The methods console.log/warn/error already have appropriate coloring in browser's console and doesn't need overriding. Current implementation creates log entries such as this:
image


exyi (approved):
Ok, I agree this is better

I think the original goal was to print `area` in different color than rest of the message. But that apparently doesn't work, and it's unnecessary overengineering anyway.

PR #1504, opened 2022-11-03
Improved error message on DataContext mismatch

acizmarik:
This PR improves error message when user does something weird with data contexts. Example:

## ViewModel
```csharp
    public class CollectionIndexViewModel : DotvvmViewModelBase
    {
        public List<int> Collection { get; set; } = Enumerable.Range(0, 4).ToList();
        public SomeOtherType Property { get; set; }
    }
```

## View
```html
<div class="container" DataContext="{value: Property}">
   <dot:Repeater DataSource="{value: Collection}" RenderSettings.Mode="Client">
      ...
   </dot:Repeater>
</div>
```

### DIFF
--- a/src/Framework/Framework/Binding/ControlPropertyBindingDataContextChangeAttribute.cs
+++ b/src/Framework/Framework/Binding/ControlPropertyBindingDataContextChangeAttribute.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using DotVVM.Framework.Binding.Expressions;
 using DotVVM.Framework.Compilation.ControlTree;
 using DotVVM.Framework.Controls;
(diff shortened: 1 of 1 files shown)


exyi (approved):

src/Framework/Framework/Binding/ControlPropertyBindingDataContextChangeAttribute.cs
```diff
@@ -29,11 +31,26 @@ public ControlPropertyBindingDataContextChangeAttribute(string propertyName, int
                 throw new Exception($"The property '{PropertyName}' was not found on control '{control.Metadata.Type}'!");
             }
 
+            [DoesNotReturn]
+            void ThrowDataContextMismatch(IAbstractBinding binding)
+            {
+                var sb = new StringBuilder();
+                sb.Append($"The '{controlProperty.Name}' property contains invalid data-binding. ");
+                sb.Append($"Binding '{binding}' could not be constructed on current context ({dataContext})");
+                throw new Exception(sb.ToString());
```
exyi, line 40: On ResolvedBinding, there is a `Errors.GetErrorMessage` method which could be useful, but we'd have to expose it on the IAbstractBinding interface

PR #1505, opened 2022-11-03
Added logging area type definition

acizmarik:
This PR improves our JS logging API by suggesting available areas in Intellisense. Moreover, I am considering whether to merge or remove certain areas. For example:

* `log` area seems too general and could be eliminated
* `evaluator` is only used by `validation` and as such can be eliminated in favor of `validation`

What do you think?

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/utils/logging.ts
+++ b/src/Framework/Framework/Resources/Scripts/utils/logging.ts
@@ -4,21 +4,37 @@ type LogLevel = "normal" | "verbose";
 
 export const level = getLogLevel();
 
-export function logInfoVerbose(area: string, ...args: any[]) {
+export type DotvvmLoggingArea = (
+    | "debug"
+    | "log"
+    | "postback"
+    | "spa"
+    | "static-command"
+    | "binding-handler"
+    | "resource-loader"
+    | "coercer"
+    | "state-manager"
+    | "validation"
+    | "evaluator"
+    | "events"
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Resources/Scripts/utils/logging.ts
```diff
@@ -4,21 +4,37 @@ type LogLevel = "normal" | "verbose";
...
 }
 
-export function logWarning(area: string, ...args: any[]) {
+export function logWarning(area: DotvvmLoggingArea, ...args: any[]) {
     console.warn(`%c${area}`, ...args);
 }
 
-export function logError(area: string, ...args: any[]) {
+export function logError(area: DotvvmLoggingArea, ...args: any[]) {
     console.error(`%c${area}`, ...args);
```
exyi, line 38:
```suggestion
    console.error(area, ...args);
```

Actually, the PR removing colors had to break this. The browser would interpret `args[0]` as a color.

exyi (approved):

exyi:
> evaluator is only used by validation and as such can be eliminated in favor of validation

I guess "evaluator" is also not very telling for most users, validation would be much better if it's only used for that. Maybe "validation-paths" if we want to be specific.

> log area seems too general and could be eliminated

if it's easy to get rid of, why not. `log` is really something you couldn't guess when you are reading a log :)

exyi (approved):

PR #1447, opened 2022-08-05
Implemented List.Contains for bindings

Mylan719:
Method works for primitive types.

usage:
`SelectedCategories.Contains(Category.Green)`
`NumberList.Contains(1)`
`NameList.Contains("Igor")`

PR #1507, opened 2022-11-04
Fix Array.Contains translation

exyi:
It just wasn't exported 🙃


acizmarik (approved):
Nice 😄. The translation was not fully released yet: https://github.com/riganti/dotvvm/pull/1447/files. We should add it to documentation though

PR #1509, opened 2022-11-04
Fix DataContextType handling in control precompilation

exyi:
When setting controls into a template property which has
a data context change attribute, we need to adjust
the expected data context. The data context type is
being tracked by `ResolvedControlHelper`, which does
the translation of runtime controls to `ResolvedControl`s.
The tracking need to be done so that the compiler can insert
correct `Internal.DataContextTypeProperty`.

See the RepeaterSetsCorrectDataContextInside test,
on what went wrong


acizmarik (approved):
👍

PR #1513, opened 2022-11-14
Fix of API refreshing

tomasherceg:


src/Samples/Tests/Tests/Feature/ApiTests.cs
```diff
@@ -241,6 +246,44 @@ public void Feature_Api_BindingSharing()
             });
         }
 
+        [Fact(Skip = "Doesn't work on OWIN because it relies on _apiCore.")]
```
exyi, line 249: This will skip it on both platforms, right?

exyi (approved):

PR #1515, opened 2022-11-14
Bump AspNetCoreLatest sample projects to .NET 7

tomasherceg:


acizmarik (approved):
Nice. We should fix those warnings in another PR

PR #1521, opened 2022-11-24
Migrate ListBox Multiple to MultiSelect

exyi:
I migrated @Mylan719's changes (validators, tests) of ListBox to the existing MultiSelect control.


src/Framework/Framework/Controls/SelectorBase.cs
```diff
@@ -87,15 +88,51 @@ public bool Enabled
...
                 if (!t.IsPrimitiveTypeDescriptor())
                 {
                     yield return new ControlUsageError("Return type of ItemValueBinding has to be a primitive type!", valueBinding.DothtmlNode);
                 }
             }
+
+            if (control.GetValue(SelectorBase.ItemTextBindingProperty) is ResolvedPropertyBinding textBinding)
+            {
+                // TODO: just change the property type to IValueBinding<string>?
+                var t = textBinding.Binding.ResultType;
```
acizmarik, line 103: This should be resolved so that types, such as `DateTime` are displated correctly

acizmarik (approved):

PR #1523, opened 2022-11-28
Fixed REST API behavior in SPA apps

tomasherceg:


src/Framework/Framework/Resources/Scripts/api/api.ts
```diff
@@ -98,10 +100,13 @@ export function invoke<T>(
     }
 
     refreshArgs()
-    ko.computed(() =>
-        refreshTriggers(args).map(trigger => typeof trigger == "string" ? eventHub.get(trigger)() : trigger())
-    )
+    const subscription = ko.computed(
+            () => refreshTriggers(args).map(trigger => typeof trigger == "string" ? eventHub.get(trigger)() : trigger())
+        )
```
exyi, line 105: You can use  https://github.com/riganti/knockout/blob/master/src/binding/defaultBindings/ifIfnotWith.js#L67
```suggestion
        , null, { disposeWhenNodeIsRemoved: lifetimeElement })
```

src/Framework/Framework/Resources/Scripts/state-manager.ts
```diff
@@ -404,3 +404,21 @@ function createWrappedObservable<T>(initialValue: DeepReadonly<T>, typeHint: Typ
     });
     return obs
 }
+
+function compareObjectTypes(currentValue: any, newVal: any): boolean {
+    if (currentValue["$type"] && currentValue["$type"] === newVal["$type"]) {
+        // objects with type must have a same type
+        return true;
+    }
+    else if (!currentValue["$type"] && !newVal["$type"]) {
```
exyi, line 413: don't dynamic objects have `$type: { type: "dynamic" }`?

src/Framework/Framework/Resources/Scripts/state-manager.ts
```diff
@@ -404,3 +404,21 @@ function createWrappedObservable<T>(initialValue: DeepReadonly<T>, typeHint: Typ
...
+        // dynamic objects must have the same properties
+        let currentValueKeys = keys(currentValue);
+        let newValKeys = keys(newVal);
+        if (currentValueKeys.length == newValKeys.length) {
+            currentValueKeys.sort();
+            newValKeys.sort();
+            return currentValueKeys.every((v, i) => newValKeys[i] === v);
+        }
```
exyi, line 421:
```suggestion
       return new Set([...currentValueKeys, ...newValKeys]).size == currentValueKeys.length
```
less error prone to indexes, probably less readable... I though the Set class would have some equals method, but no :|

src/Framework/Framework/Resources/Scripts/api/api.ts
```diff
@@ -29,7 +30,8 @@ export function invoke<T>(
     refreshTriggers: (args: any[]) => Array<KnockoutObservable<any> | string>,
     notifyTriggers: (args: any[]) => string[],
     element: HTMLElement,
```
exyi, line 32: I'd rename this one to `cacheElement` or something, and it should be nullable. I was actually quite confused about how it works when I was going through this function last week

exyi (approved):

src/Framework/Framework/Resources/Scripts/state-manager.ts
tomasherceg: No. 
The `{ type: "dynamic" }` can appear only in type metadata, not in the ViewModel objects
In the ViewModel objects, the `$type` property is always `string` and is always a hash of the .NET type name, which is subsequently looked up in type metadata. Missing `$type` there means that it is a dynamic object.

exyi: Ok.

PR #1526, opened 2022-12-04
Calling .NET WebAssembly code from views

tomasherceg:
This PR enables the user to reference a C# project compiled as a web assembly and invoke methods on imported "modules".

There are several things that need to be done:
[ ] Decide if we'll allow importing multiple modules in the same view; currently, only one is supported
[ ] More tests for SPA, combination with JS modules
[ ] Production builds and trimming - in Debug mode, we are using reflection, and trimming is disabled
[ ] Publishing experience - MSBuild task that will take compiled app bundle and ship it as part of the web app
[ ] Better indication of errors and error handling - invoking the error screen from the client
[ ] Serialization of enums with EnumMember attribute


src/Framework/Framework/Compilation/ControlTree/BindingExtensionParameter.cs
```diff
@@ -211,6 +211,52 @@ public ViewModuleAnnotation(string id, bool isMarkupControl)
...
+        public override Expression GetServerEquivalent(Expression controlParameter)
+        {
+            var type = ResolvedTypeDescriptor.ToSystemType(this.Type);
+            var constructors = type.GetConstructors(BindingFlags.Public);
+            if (constructors.Length != 1 || constructors[0].GetParameters().Length != 1)
+            {
+                throw new DotvvmCompilationException($"The type {type} referenced in the @dotnet directive must have exactly one public constructor with one parameter of IViewModuleContext!");
+            }
+            // TODO: check parameter type
+            return Expression.New(constructors[0], Expression.Constant(null, typeof(object)))
```
exyi, line 236:
```suggestion
            return Expression.New(constructors[0], Expression.Constant(null, typeof(IViewModuleContext)))
```

It won't compile with `object`.

However, I'm thinking if we want to allow invoking the methods server-side. I guess it makes sense, but passing null into the constructor will just make it fail on the unexpected null reference. Maybe only allow it if the type also has parameterless constructor?

src/Framework/Framework/Compilation/Directives/DotnetViewModuleDirectiveCompiler.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public DotnetViewModuleDirectiveCompiler(IReadOnlyDictionary<string, IReadOnlyList<DothtmlDirectiveNode>> directiveNodesByName, IAbstractTreeBuilder treeBuilder, IAbstractControlBuilderDescriptor? masterPage, bool isMarkupControl, ImmutableList<NamespaceImport> imports)
+        : base(directiveNodesByName, treeBuilder)
+    {
+        this.masterPage = masterPage;
+        this.isMarkupControl = isMarkupControl;
+        this.imports = imports;
+    }
+
+    public override string DirectiveName => ParserConstants.CsharpViewModuleDirective;
```
exyi, line 28: You might want to rename CsharpViewModuleDirective to DotnetViewModuleDirective

src/Framework/Framework/Resources/Scripts/dotvvm-root.ts
```diff
@@ -107,6 +107,7 @@ const dotvvmExports = {
         call: viewModuleManager.callViewModuleCommand,
         registerMany: viewModuleManager.registerViewModules
     },
+    get virtualDirectory() { return getVirtualDirectory() },
```
exyi, line 110: I can't find where this is needed.

exyi (commented):
I find the changes in the framework fine (except for the `dotvvm-with-view-modules` omission).

It's harder to tell how well will the WASM related code work, It's a bit sad we use reflection to invoke the methods, but I don't see any way around it :shrug:

PR #1527, opened 2022-12-06
CompositeControl: merge MarkupOption attribute with implicit metadata

exyi:
Added internal nullable field to each MarkupOptions property, so
we can distinguish if the user used the property explicitly or
it only contains the default value.

### DIFF
--- a/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
+++ b/src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
@@ -349,7 +349,7 @@ namespace DotVVM.Framework.Binding
                     if (typeof(IDotvvmObjectLike).IsAssignableFrom(type) ||
                         typeof(ITemplate).IsAssignableFrom(type) ||
                         typeof(IEnumerable<IDotvvmObjectLike>).IsAssignableFrom(type))
-                        dotvvmProperty.MarkupOptions._mappingMode = MappingMode.InnerElement;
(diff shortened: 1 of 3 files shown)


src/Framework/Framework/Binding/DotvvmCapabilityProperty.cs
```diff
@@ -338,20 +338,17 @@ internal static IControlAttributeDescriptor InitializeArgument(ICustomAttributeP
...
-                            typeof(ITemplate).IsAssignableFrom(type) ||
-                            typeof(IEnumerable<IDotvvmObjectLike>).IsAssignableFrom(type))
-                            dotvvmProperty.MarkupOptions.MappingMode = MappingMode.InnerElement;
-                    }
+                        dotvvmProperty.MarkupOptions._allowBinding ??= false;
+
+                    if (typeof(IDotvvmObjectLike).IsAssignableFrom(type) ||
+                        typeof(ITemplate).IsAssignableFrom(type) ||
+                        typeof(IEnumerable<IDotvvmObjectLike>).IsAssignableFrom(type))
+                        dotvvmProperty.MarkupOptions._mappingMode = MappingMode.InnerElement;
```
tomasherceg, line 351: Shall we overwrite `mappingMode` here, or set it only when it's null via `??=`? 
What if someone sets it to `Exclude`?

tomasherceg (approved):

PR #1541, opened 2023-01-04
Add documentation comments to various binding-related classes

exyi:
Part is taken from #1392 which won't be merged for a while.


src/Framework/Core/ViewModel/Direction.cs
```diff
@@ -5,19 +5,31 @@
 namespace DotVVM.Framework.ViewModel
 {
     ///<summary>
-    /// ServerToClient, ServerToClient on postback, ClientToServer, C2S iff in command path
+    /// Specifies on which requests should the property should serialized and sent. Default is <see cref="Direction.Both">Both</see>.
```
acizmarik, line 8:
```suggestion
    /// Specifies on which requests should the property be serialized and sent. Default is <see cref="Direction.Both">Both</see>.
```

src/Framework/Core/ViewModel/Direction.cs
```diff
@@ -5,19 +5,31 @@
...
+    /// Set to <see cref="Direction.None">None</see> to disable serialization of the property.
+    /// This enums is flags, the directions can be arbitrarily combined.
```
acizmarik, line 10:
```suggestion
    /// This enumeration can be treated as flags, the directions can be arbitrarily combined.
```

acizmarik (approved):
Nice!

PR #1542, opened 2023-01-04
MarkupFileLoader will not reload in Release and not leak in Debug

exyi:
MarkupFileLoader did check the LastWriteTimeUtc on each request and would recompile the page if the page changed. While we have worse performance issues, this is quite unnecesary. IMHO it's more likely to break production server when new source code is copied there.

Moreover, the loader would allocate new IControlBuilder, but didn't unreference the old one, leaking some memory. I didn't test that this commit fixes it, though.

### DIFF
--- a/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
+++ b/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
@@ -31,7 +31,7 @@ namespace DotVVM.Framework.Compilation
         public DefaultControlBuilderFactory(DotvvmConfiguration configuration, IMarkupFileLoader markupFileLoader, CompiledAssemblyCache compiledAssemblyCache)
         {
             this.configuration = configuration;
-            this.allowReload = configuration.Debug; // TODO: do we want another option for this?
+            this.allowReload = configuration.Debug;
 
             // WORKAROUND: there is a circular dependency
             // TODO: get rid of that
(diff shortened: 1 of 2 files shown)


src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
```diff
@@ -20,16 +20,18 @@ namespace DotVVM.Framework.Compilation
...
 
         public Func<IViewCompiler> ViewCompilerFactory { get; private set; }
 
-        private ConcurrentDictionary<MarkupFile, Lazy<(ControlBuilderDescriptor, Lazy<IControlBuilder>)>> controlBuilders = new ConcurrentDictionary<MarkupFile, Lazy<(ControlBuilderDescriptor, Lazy<IControlBuilder>)>>();
+        private ConcurrentDictionary<string, Lazy<(ControlBuilderDescriptor, Lazy<IControlBuilder>)>> controlBuilders = new();
 
         public DefaultControlBuilderFactory(DotvvmConfiguration configuration, IMarkupFileLoader markupFileLoader, CompiledAssemblyCache compiledAssemblyCache)
         {
             this.configuration = configuration;
+            this.allowReload = configuration.Debug; // TODO: do we want another option for this?
```
acizmarik, line 34: I don't think additional setting is necessary

acizmarik (approved):

acizmarik:
There is still a race condition but affects only debug experience (file might be removed at a wrong time and it could theoretically cause a crash)

PR #1545, opened 2023-01-09
Added `UnsupportedCallSiteAttribute` and an analyzer

acizmarik:
This PR adds `UnsupportedCallSiteAttribute` that can be placed on a method to signalize that a specific invocation (either from server-side or client-side) is not supported. Moreover, there is also a Roslyn Analyzer that reports violation whenever someone tries to invoke a method marked as only client-side.

Our previous solution was to annotate the method with the `ObsoleteAttribute`. IMO there are the following advantages to the proposed solution for this problem:
* The API no longer looks like it is obsoleted
* VSExtension can not easily filter compiler warnings. Right now it issues false-positives whenever someone uses these methods (for example, `ToBrowserLocalTime`). It is easier to ignore analyzer.
* Users can also use the `UnsupportedCallSiteAttribute` if they wish to make for example a client-side-only method

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/UnsupportedCallSiteCheckingVisitor.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/UnsupportedCallSiteCheckingVisitor.cs
@@ -2,7 +2,7 @@
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-using DotVVM.Core.CodeAnalysis;
+using DotVVM.Framework.CodeAnalysis;
 using DotVVM.Framework.Binding.Expressions;
 using DotVVM.Framework.Compilation.ControlTree.Resolved;
 
--- a/src/Framework/Framework/Compilation/ControlTree/UnsupportedCallSiteCheckingVisitor.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/UnsupportedCallSiteCheckingVisitor.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using DotVVM.Core.CodeAnalysis;
+using DotVVM.Framework.Binding.Expressions;
+using DotVVM.Framework.Compilation.ControlTree.Resolved;
+
+namespace DotVVM.Framework.Compilation.ControlTree
+{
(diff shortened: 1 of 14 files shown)


acizmarik:
Client-side-only methods can be guarded by the proposed Roslyn Analyzer.
Server-side-only methods can be checked by VSExtension (emit a warning when used inside a binding). We could also issue a warning when compiling the page using framework.

acizmarik:
Furthermore, we should maybe also check client-side-only methods on server because some behaviours might not be expected by users. For example using `ToBrowserLocalTime` in a resource binding.

* Should it silently render UTC?
* Should it render UTC and issue a warning during compilation?
* It probably should not error-out as that would be a breaking change

exyi (approved):

src/Framework/Framework/Compilation/ControlTree/UnsupportedCallSiteCheckingVisitor.cs
```diff
@@ -0,0 +1,43 @@
...
+    public class UnsupportedCallSiteCheckingVisitor : ResolvedControlTreeVisitor
+    {
+        class ExpressionInspectingVisitor : ExpressionVisitor
+        {
+            public event Action<MethodInfo>? InvalidCallSiteDetected;
+
+            protected override Expression VisitMethodCall(MethodCallExpression node)
+            {
+                var callSiteAttr = node.Method.CustomAttributes.FirstOrDefault(a => a.AttributeType == typeof(UnsupportedCallSiteAttribute));
+                if (callSiteAttr is not null && callSiteAttr.ConstructorArguments.Any())
```
exyi, line 20: I'd maybe check it using `node.Method.IsDefined(typeof(UnsupportedCallSiteAttribute))`, it's significantly faster and this will be called for all methods in all bindings. Easier to "optimize it" than measure the impact on startup time...

exyi:
One more note: we could disable this warning in Linq.Expressions-typed lambdas, since it may be the ValueOrBinding.Select or a similar method which produces a client-side binding. I found this snippet which checks if we are in Linq.Expression https://github.com/dotnet/roslyn-analyzers/pull/4621/files#diff-5fbef35187753a9b6db0c89fe8d00a9da02b9f8887454aaea458fbb041b9e080R518

exyi (approved):
Great, thank you :+1:

PR #1549, opened 2023-01-15
Instrument DotVVM with System.Diagnostics.Metrics 

exyi:
The aim is to add prometheus-compatible metrics
to places which can signal a potential problem with
the application (too large VM, compiling too many
bindings or views, ...). The samples project now includes
reference to prometheus-net, so you can go to `/metrics` to see them.

I couldn't find what naming convention should be used for
these .NET metrics, so I used the prometheus conventions
(maybe except label values are in PascalCase, because it's easier that way).
However, prometheus-net seems to have some naming convertor,
which suggest the convention for System.Diagnostics.Metrics
are different (or non-existent...)

### DIFF
--- a/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
+++ b/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
@@ -79,10 +79,12 @@ namespace DotVVM.Framework.Compilation
             {
                 var sw = ValueStopwatch.StartNew();
                 var (descriptor, factory) = ViewCompilerFactory().CompileView(file.ReadContent(), file.FileName);
+                var phase1Ticks = sw.ElapsedTicks;
 
                 var lazyBuilder = new Lazy<IControlBuilder>(() => {
                     try
                     {
+                        sw.Restart();
                         var result = factory();
 
                         // register the internal resource after the page is successfully compiled,
@@ -111,7 +113,7 @@ namespace DotVVM.Framework.Compilation
                     }
                     finally
                     {
-                        Interlocked.Add(ref DotvvmMetrics.BareCounters.ViewsCompilationTime, sw.ElapsedTicks);
+                        Interlocked.Add(ref DotvvmMetrics.BareCounters.ViewsCompilationTime, phase1Ticks + sw.ElapsedTicks);
                     }
                 });
 
--- a/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
+++ b/src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using DotVVM.Framework.Compilation.ViewCompiler;
 using DotVVM.Framework.Configuration;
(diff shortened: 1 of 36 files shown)


src/Framework/Framework/Hosting/Middlewares/DotvvmLocalResourceMiddleware.cs
```diff
@@ -23,6 +23,7 @@ public DotvvmLocalResourceMiddleware(ILocalResourceUrlManager urlManager, Dotvvm
 
         public async Task<bool> Handle(IDotvvmRequestContext request)
         {
+            var sw = ValueStopwatch.StartNew(isActive: DotvvmMetrics.ResourceServeDuration.Enabled);
```
acizmarik, line 26: We should add a `finally` block to capture also serve failures

src/Framework/Framework/Compilation/DefaultControlBuilderFactory.cs
```diff
@@ -76,6 +77,7 @@ void editCompilationException(DotvvmCompilationException ex)
             }
             try
             {
+                var sw = ValueStopwatch.StartNew();
```
acizmarik, line 80: The stopwatch is started too soon. The compilation might not begin immediately

src/Framework/Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -243,8 +244,21 @@ public async Task ProcessRequestCore(IDotvvmRequestContext context)
...
+                    {
+                        commandResult = await ExecuteCommand(actionInfo, context, methodFilters);
+                    }
+                    finally
+                    {
+                        DotvvmMetrics.CommandInvocationDuration.Record(
+                            commandTimer.ElapsedSeconds,
+                            new KeyValuePair<string, object?>("command", actionInfo.Binding!.ToString()),
+                            new KeyValuePair<string, object?>("result", context.CommandException is null ? "Ok" :
+                                                                        context.IsCommandExceptionHandled ? "Exception" :
```
acizmarik, line 259:
```suggestion
                                                                        context.IsCommandExceptionHandled ? "HandledException" :
```

src/Framework/Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -377,7 +391,21 @@ public async Task ProcessStaticCommandRequest(IDotvvmRequestContext context)
...
+                {
+                    result = await ExecuteCommand(actionInfo, context, filters);
+                }
+                finally
+                {
+                    DotvvmMetrics.StaticCommandInvocationDuration.Record(
+                        commandTimer.ElapsedSeconds,
+                        new KeyValuePair<string, object?>("command", executionPlan.ToString()),
+                        new KeyValuePair<string, object?>("result", context.CommandException is null ? "Ok" :
+                                                                    context.IsCommandExceptionHandled ? "Exception" :
```
acizmarik, line 406:
```suggestion
                                                                    context.IsCommandExceptionHandled ? "HandledException" :
```

acizmarik (commented):
Let's also investigate redirects for static commands. Otherwise looks nice

exyi:
I checked that the static command redirects work the same way as before (although I don't get it why :sweat_smile:). We have a test for it (`Redirect_StaticCommand`)

acizmarik (approved):
👍

acizmarik (changes requested):
There are still some issues with assembly bindings to be solved.

acizmarik (approved):

PR #1550, opened 2023-01-16
Fixed directives to be case-insensitive again

acizmarik:
This PR fixes a bug that forces users to specify directives with exact casing (for example, `@viewModel` is valid and `@viewmodel` is not). The breaking change was introduced in one of the preview builds (4.0 - 4.1-preview)

image


exyi (approved):
:+1:

I think a unit test for this would be preferable. There is even ViewModelDirectiveTest class for it (or you can just change one of the existing to use lowercase `@viewmodel`)

acizmarik:
I agree. I overlooked the file when originally writing the test. It is fixed now.

PR #1551, opened 2023-01-16
REST API binding - IncludedInPage behavior fixed

tomasherceg:
I fixed another problem in REST API bindings reported by a customer.

When the binding is used in an element with `IncludeInPage` set to `false`, the binding is not evaluated - that is correct.
When the element appears on the page, the binding is evaluated at that moment - that is also correct.

However, when the element disappears and re-appears on the page, the binding wasn't refreshed.
The problem was that the reference counting wasn't working well as the binding often evaluates multiple times on the same element so it was incremented multiple times, but decremented only once. I changed that to an array of elements so it is easier to recognize whether the binding was already counted or not.

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/api/api.ts
+++ b/src/Framework/Framework/Resources/Scripts/api/api.ts
@@ -14,27 +14,24 @@ class CachedValue {
     _stateManager?: StateManager<any>
     _isLoading?: boolean
     _promise?: PromiseLike<any>
-    _elements: HTMLElement[] = []
+    _elements: Set<HTMLElement> = new Set()
 
     constructor(public _cache: Cache) {
     }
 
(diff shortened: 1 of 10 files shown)


src/Framework/Framework/Resources/Scripts/api/api.ts
```diff
@@ -10,18 +10,39 @@ type ApiComputed<T> =
         refreshValue: () => PromiseLike<any>
     };
 
-type CachedValue = {
+class CachedValue {
     _stateManager?: StateManager<any>
     _isLoading?: boolean
     _promise?: PromiseLike<any>
-    _referenceCount: number
+    _elements: HTMLElement[] = []
```
acizmarik, line 17: We should change array to set

exyi (approved):
Just the Array -> Set, then it's good to go

PR #1553, opened 2023-01-17
Fixed issue in `Validator` in connection with BP

acizmarik:
This PR fixes a breaking change that was introduced in PR #1385.

I added there a condition that was supposed to throw an exception if the `Validator.Value` can not be resolved to a valid value binding. A customer found out that it is in fact possible to enter this branch when using BP and its `FormItem` control. BP automatically sets validator properties and I am guessing that in some cases it sets there something a little weird.

This change should revert to the previous behaviour while keeping the improved one if the binding can be actually resolved.

image

### DIFF
--- a/src/Framework/Framework/Controls/Validator.cs
+++ b/src/Framework/Framework/Controls/Validator.cs
@@ -99,7 +99,9 @@ namespace DotVVM.Framework.Controls
             }
             else
             {
-                throw new DotvvmControlException($"Could not resolve {nameof(ValueProperty)} to a valid value binding.");
+                // Could not resolve ValueProperty to a valid value binding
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Controls/Validator.cs
```diff
@@ -99,7 +99,9 @@ private static void AddValidatedValue(IHtmlWriter writer, IDotvvmRequestContext
             }
             else
             {
-                throw new DotvvmControlException($"Could not resolve {nameof(ValueProperty)} to a valid value binding.");
```
acizmarik, line 102: Either remove this or check that there is not a hard-coded value

exyi (approved):

PR #1559, opened 2023-01-21
Initialization of DotVVM in a single file

tomasherceg:
Added extension methods which support configuration of the app in one file without `DotvvmStartup`.

```
var builder = WebApplication.CreateBuilder(args);
builder.AddDotVVM(options =>
{
    // register DotVVM services
    options.AddAutoUI();
    ...
});

var app = builder.Build();
app.UseDotVVM(..., config => 
{
   // fine-tune DotvvmConfiguration 
});
app.UseStaticFiles();
app.Run();
```

### DIFF
--- a/src/Framework/Hosting.AspNetCore/WebApplicationBuilderExtensions.cs
+++ b/src/Framework/Hosting.AspNetCore/WebApplicationBuilderExtensions.cs
@@ -1,0 +1,56 @@
+﻿#if NET6_0
+#nullable enable
+using System;
+using DotVVM.Framework.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Builder;
+
+public static class WebApplicationBuilderExtensions
+{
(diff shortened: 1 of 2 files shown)


acizmarik:
We should keep the `DotvvmStartup` class so that we can revive compiler one day

src/Framework/Hosting.AspNetCore/WebApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        return builder;
+    }
+
+    private static void AddDotvvmServiceDependencies(WebApplicationBuilder builder)
+    {
+        builder.Services.AddDataProtection();
+        builder.Services.AddAuthorization();
+        builder.Services.AddWebEncoders();
+        builder.Services.AddAuthentication();
+        builder.Services.AddHttpContextAccessor();
```
exyi, line 54: We only really need DataProtection, and sometimes Authorization. I guess there is no harm in adding the other services (?).
```suggestion
```

However, I think we shouldn't enable ~~advanced async debugging practicals~~ IHttpContextAccessor automatically. (I hope none of DotVVM code depends on it)

exyi (approved):

PR #1563, opened 2023-01-23
styles: support controls in capabilities

exyi:


exyi:
@tomasherceg You can add your test case. I will write some unit tests for this, but I don't know how to use AutoUI to trigger this codepath

tomasherceg (approved):

PR #1566, opened 2023-01-25
Update the template to DotVVM 4.1 and net7

exyi:


acizmarik (approved):
We should rename the folder `Views` to `Pages`. It will also work better with VS Extension

PR #1567, opened 2023-01-27
Add the DotvvmGlobalExtensions interface

cafour:
Adds an extension point for the `window.dotvvm` global object. This is would benefit BusinessPack, which could then use `window.dotvvm.bp`.

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/dotvvm-root.ts
+++ b/src/Framework/Framework/Resources/Scripts/dotvvm-root.ts
@@ -140,7 +140,7 @@ declare global {
     const dotvvm: DotvvmGlobalExtensions & typeof dotvvmExports & {debug?: true, isSpaReady?: typeof isSpaReady, handleSpaNavigation?: typeof handleSpaNavigation};
 
     interface Window {
-        dotvvm: typeof dotvvmExports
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Resources/Scripts/dotvvm-root.ts
```diff
@@ -132,11 +132,15 @@ if (compileConstants.debug) {
...
+
 declare global {
-    const dotvvm: typeof dotvvmExports & {debug?: true, isSpaReady?: typeof isSpaReady, handleSpaNavigation?: typeof handleSpaNavigation};
+    interface DotvvmGlobalExtensions {}
+
+    const dotvvm: DotvvmGlobalExtensions & typeof dotvvmExports & {debug?: true, isSpaReady?: typeof isSpaReady, handleSpaNavigation?: typeof handleSpaNavigation};
 
     interface Window {
-        dotvvm: typeof dotvvmExports
+        dotvvm: DotvvmGlobalExtensions & typeof dotvvmExports
```
exyi, line 143: Not your mistake, but I'd guess we didn't want `const dotvvm` and `window.dotvvm` to have different types, would you mind to unify them (using `dotvvm: typeof dotvvm` could work)

exyi (approved):

PR #1573, opened 2023-02-02
Relax the constraint on React version

exyi:
We used to assert that React 17 is used. We don't use any advanced React feature, so this constraint is irrelevant. I set it to 16, since that was the first open source release (with the patents)


acizmarik (approved):

acizmarik:
Please fix tests before merging. Something related to React is failing in UI tests

PR #1575, opened 2023-02-03
Fixed warning when building analyzers

acizmarik:
This PR removes the false-positive when building analyzers

See: https://github.com/dotnet/roslyn-analyzers/issues/5828 for more info


exyi (approved):
ok, wtf :D

PR #1579, opened 2023-02-10
Fixed various issues in BootstrapForm

tomasherceg:
Fixed issues with `BootstrapForm` rendering and added tests to ensure compatibility with Bootstrap 3, 4, and 5.


exyi (approved):
You just need to accept the new test outputs

PR #1582, opened 2023-02-13
Added test for collection as Validation.Target

### LINKED ISSUE #508: Validation doesn't work when Target is a Collection
When I bind Validation.Target to a **Collection** in a ViewModel, client validation ignores the collection entirely, however, server validation yields errors correctly.

Objects in the collection have the following shape:
```C#
public class PersonalDataDTO
{
    [Required(ErrorMessage = "I am required")]
    public string Name { get; set; }
}
```
The interesting is that every next client validation fails quietly (ValidationSummary doesn't show anything).

IMHO we are dealing with 2 different bugs:
* One is the lack of support for arrays in client validation
* Second is connected to the `dotvvm.validation.errors` array in dotvvm js: 
    * It contains errors that should have been cleared before the postback

Thanks to @ENgateman and @exyi for advice with this bug.

acizmarik:
Resolves #508. This feature was probably fixed/implemented within the big validation rewrite. However, it was not covered by any tests.


exyi (approved):
:+1:

Issue #1576, opened 2023-02-07
Error on refreshing the page


khurram4u2:
Hi,

I am using dotvvm to modernize the asp.net web form page. I am able to load the data in view first time, however when I refresh the page, I am getting following error.

![image](https://user-images.githubusercontent.com/56805276/217158362-9e0b6852-45c3-49b2-b9ba-1b4b7b14cdae.png)

could you please advise what could be the problem?

Thanks Regards,

khurram4u2:
Hi @acizmarik, yes I am using custom loader to use the DI (castle windsor). My custom loader code is 

![image](https://user-images.githubusercontent.com/56805276/217378295-90483332-c4cd-44f7-87c9-9b6856ae51d3.png)

I am using following code to register view modal and windsor castle. 

`CustomHttpApplication.WindsorContainer.AddComponent<CentreOfLearningViewModel>();
            options.Services.AddSingleton<IViewModelLoader>(serviceProvider => new WindsorViewModelLoader((WindsorContainer)CustomHttpApplication.WindsorContainer));'

Could you please let me know if I am using the correct way to register view modal

khurram4u2:
Hi @acizmarik, Thanks for your advise. I was able to register the component with transit life style. I release by defaul windsor castle use the singleton life style.

Thanks.

PR #1586, opened 2023-02-14
Added a better error when someone tries to reuse viewmodels

acizmarik:
This PR fixes a generic exception `An item with the same key has already been added` when trying to reuse instances of viewmodels. Even though this is still not supported, users will at least receive a message that tells what went wrong and possibly how to fix it.

Relevant to: #1576


exyi (approved):
:+1: 

I'd maybe include type of the view model which caused the problem. Since it could be a nested viewmodel, it may not be immediately clear which class is causing the problem. Also, lets put this into 4.1 as bugfix

PR #1588, opened 2023-02-15
Fixed issue where duplicities were returned after upgrading to SDK 7.0.200

acizmarik:
This PR fixes a weird breaking change where `DotVVM.Framework`, `DotVVM.Framework.Hosting` and `DotVVM.Framework.Testing` were returned twice. I am not sure whether something changed in our configuration but the only change that I was able to observe is GitHub Actions started using 7.0.200 instead of 7.0.102 SDKs


acizmarik:
This is a (temporary) work-around until we investigate the real source of the problem. It looks like a breaking change but also affects only our assemblies...

exyi (approved):
:+1:  The Distinct can't hurt, but are right that we might be only masking another issue...

PR #1591, opened 2023-02-18
Fix AttributeList in precompiled control

exyi:
This is a stupid fix, since it propagates the AttributeList into the runtime HtmlGenericControl, instead of merging the attributes compile-time. However, doing that requires other significant changes in AttributeValueMergers, and I'd better not do it in a patch release


acizmarik (approved):
👍 

We could create an issue for 4.2 about the mentioned improvement

PR #1594, opened 2023-02-19
Extract generic type data context change attribute

tomasherceg:
This PR adds a new `ExtractGenericArgumentDataContextChangeAttribute`, which looks at base types or implemented interfaces of the current data context type (e. g. `IGroupingDataSet<T>`) and extracts the type argument on a specified index.

### DIFF
--- a/src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
+++ b/src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using DotVVM.Framework.Compilation.ControlTree;
 using DotVVM.Framework.Controls;
 using DotVVM.Framework.Utils;
+using FastExpressionCompiler;
 
 namespace DotVVM.Framework.Binding;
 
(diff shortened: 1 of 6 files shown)


src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+    public override Type? GetChildDataContextType(Type dataContext, DataContextStack controlContextStack, DotvvmBindableObject control, DotvvmProperty? property = null)
+    {
+        var implementations = ReflectionUtils.GetBaseTypesAndInterfaces(dataContext).ToList();
+        if (implementations.Count == 0)
+        {
+            throw new Exception($"The data context {dataContext} doesn't implement {GenericType}!");
+        }
+        else if (implementations.Count > 1)
+        {
+            throw new Exception($"The data context {dataContext} has multiple implementations of {GenericType}! Cannot decide which one to extract:\n" + string.Join("\n", implementations));
```
exyi, line 54:
```suggestion
            throw new Exception($"The data context {dataContext} has multiple implementations of {GenericType}! Cannot decide which one to extract:\n" + string.Join("\n", implementations.Select(t => t.ToCode())));
```

acizmarik (approved):

exyi:
It would be better to cover this by an unit test.

PR #1597, opened 2023-02-19
Fix regression with empty required control properties

exyi:
DotVVM 4.0 didn't care if a required property is
actually empty, but 4.1 removes the required
properties in the StyleTreeShufflingVisitor, which triggers the validator. Fix is to just avoid removing properties which are required, even when they are empty.


src/Framework/Framework/Compilation/Styles/StyleTreeShufflingVisitor.cs
```diff
@@ -42,12 +42,13 @@ public override void VisitControl(ResolvedControl control)
 
             foreach (var (prop, setter) in control.Properties.ToArray())
             {
-                // remove empty properties
+                // remove empty properties, but preserve empty collections if they are marked as Required
```
acizmarik, line 45: You mentioned that we want to preserve empty collections. Is it valid also for templates? It seems like something related is failing in unit tests

acizmarik (approved):

PR #1598, opened 2023-02-21
Replace typeMetadata after cleaning controls in SPA navigate

exyi:
If the control touched viewmodel in the dispose logic, it would fail with "Cannot find type metadata..." error.

cc @martindybal


acizmarik (approved):
👍

PR #1603, opened 2023-03-01
Added _api.IsLoading property

tomasherceg:
If a `Repeater` or `GridView` is bound from REST API bindings, the `EmptyDataTemplate` is displayed before the data are loaded into the control.
I've added a simple property `_api.IsLoading` to be able to determine this state so the user can hide the empty template content or replace it with a custom loading animation. 

In the future, we can add more detailed API to detect on which API client the loading occurs, or maybe filter by URL prefix. But for now it is sufficient.


exyi (approved):

exyi:
I don't think we need a new way to tell IsLoading for specific API endpoint - you can take the API call and compare it with `null`.

```
<div Visible={value: _myApi.GetWhatever() == null}>
   Loading...
</div>
```

PR #1611, opened 2023-03-06
Improved support for dynamics in JS

acizmarik:
This PR fixes a couple of bugs I found while attempting to fix some bugs in BP in connection with filters and usage of dynamics.

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/metadata/typeMap.ts
+++ b/src/Framework/Framework/Resources/Scripts/metadata/typeMap.ts
@@ -3,7 +3,7 @@ import { keys } from "../utils/objects";
 
 let types: TypeMap = {};
 
-export function getTypeInfo(typeId: string): TypeMetadata {
+export function getTypeInfo(typeId: string | object): TypeMetadata {
 
     if (typeof typeId === "string") {
         var result = types[typeId];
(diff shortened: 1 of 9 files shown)


src/Framework/Framework/Resources/Scripts/metadata/typeMap.ts
```diff
@@ -3,18 +3,27 @@ import { keys } from "../utils/objects";
...
+    if (typeof typeId === "string") {
+        var result = types[typeId];
+        if (result) {
+            return result;
+        }
+    }
+    else if (typeof typeId === "object") {
+        var typeInfo = typeId as any;
+        if (ko.unwrap(typeInfo?.type) === "dynamic")
+            return { type: "dynamic" };
```
exyi, line 17:
```suggestion
            return { type: "dynamic", properties: {} };
```

I think this could eliminate number of conditions bellow. AFAIK are supposed to treat any additional properties as dynamic (to not fall apart on inheritance), so dynamic type is essentially a type without any information about its properties.

exyi (approved):

src/Framework/Framework/Resources/Scripts/metadata/typeMap.ts
acizmarik: I agree but it causes problems in some other places. We can refactor it a little if we want in another PR

PR #1614, opened 2023-03-09
Changed typehash computation to be less volatile

acizmarik:
Currently, when accessing `typeof(..).FullName` it can contain assembly versions in case generics are present. Using FastExpressionCompiler's `ToCode(..)` should improve the situation a bit as assembly versions no longer contribute when computing hash.


exyi (approved):
:+1:

Issue #1557, opened 2023-01-18
Can't add StylesheetResource to Head, it is already rendered


acizmarik:
A customer got the following exception: `DotVVM.Framework.ResourceManagement.ResourceManager+ResourcePositionUnavailableException: Can't add StylesheetResource 'dotvvm.businesspack-default-css' to Head, it is already rendered.`

## Control Hierarchy
* <bp:FileUpload
* ...
* <dot:DataItemContainer
* <dot:GridView
* <dot:PlaceHolder
* <dot:DotvvmMarkupControl
* <bs:TabItem
* ...
* <bs:TabControl
* ...
* <dot:Content
* <dot:PlaceHolder
* <dot:ContentPlaceHolder
* ...
* <dot:DotvvmView

## Stack Trace

> DotVVM.Framework.ResourceManagement.ResourceManager.AddRequiredResource(String name, IResource resource)
DotVVM.BusinessPack.Controls.ResourceManagerExtensions.AddBusinessPackStyleResourcesBasedOnBrowser(ResourceManager resourceManager, IDotvvmRequestContext context, String resourceName)
DotVVM.BusinessPack.Controls.ResourceManagerExtensions.AddBusinessPackStylesheetResources(ResourceManager resourceManager, IDotvvmRequestContext context)
DotVVM.BusinessPack.Controls.ResourceManagerExtensions.AddBusinessPackRequiredResources(ResourceManager resourceManager, IDotvvmRequestContext context)
DotVVM.BusinessPack.Controls.CompositeHtmlControlBase.OnInit(IDotvvmRequestContext context)
DotVVM.BusinessPack.Controls.FileUpload.OnInit(IDotvvmRequestContext context)
DotVVM.Framework.Controls.DotvvmControlCollection.InvokeMissedPageLifeCycleEvent(IDotvvmRequestContext context, LifeCycleEventType targetEventType, Boolean isMissingInvoke, DotvvmControl& lastProcessedControl)
DotVVM.Framework.Controls.DotvvmControlCollection.InvokeMissedPageLifeCycleEvent(IDotvvmRequestContext context, LifeCycleEventType targetEventType, Boolean isMissingInvoke, DotvvmControl& lastProcessedControl)
DotVVM.Framework.Controls.DotvvmControlCollection.InvokeMissedPageLifeCycleEvent(IDotvvmRequestContext context, LifeCycleEventType targetEventType, Boolean isMissingInvoke, DotvvmControl& lastProcessedControl)
[...]

PR #1615, opened 2023-03-09
grid view: Initialize client templates in PreRender

exyi:
Initializing them in Render leads to problems with resource
registrations.

Fix for https://github.com/riganti/dotvvm/issues/1557

The GridViewColumn tests were only moved to the new GridViewTests file.


acizmarik (approved):
👍

PR #1619, opened 2023-03-19
Add InlineScriptResource.RenderScript helper method

exyi:
It is useful when inheriting from the InlineScriptResource

cc @martindybal

### DIFF
--- a/src/Framework/Framework/ResourceManagement/InlineScriptResource.cs
+++ b/src/Framework/Framework/ResourceManagement/InlineScriptResource.cs
@@ -65,14 +65,19 @@ namespace DotVVM.Framework.ResourceManagement
         /// </summary>
         public override void Render(IHtmlWriter writer, IDotvvmRequestContext context, string resourceName)
         {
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/ResourceManagement/InlineScriptResource.cs
```diff
@@ -65,14 +65,19 @@ static bool InlineScriptContentGuard(string? code)
         /// </summary>
         public override void Render(IHtmlWriter writer, IDotvvmRequestContext context, string resourceName)
         {
-            var code = Code;
+            RenderScript(writer, Code, Defer);
+        }
+
+        /// <summary> Renders a &lt;script&gt; element with the <paramref name="code"/> content. </summary>
+        public static void RenderScript(IHtmlWriter writer, string code, bool defer)
```
martindybal, line 72: I'm voting for `virtual` method.

acizmarik (approved):

PR #1631, opened 2023-03-31
Always replace in history on redirect in SPA navigation

exyi:
Otherwise, two pages are placed into the history during navigation, making it impossible (or at least hard) to navigate back.


acizmarik (approved):
Let's also add a test before merging

PR #1632, opened 2023-04-04
Remove runtime dependency from compiler pipeline

acizmarik:
This PR fixes a couple of problems introduced to the compiler pipeline's internal API in version 4.1. These fixes make it significantly easier to adopt DotVVM 4.1 for VS Extension.

**What was not ideal**:
* `IMarkupDirectiveCompilerPipeline` required `IControlBuilderFactory` whose implementation directly depends on runtime information (`CompiledAssemblyCache` and others)
* The runtime information was needed to achieve three things:
  * Master page is being resolved almost immediately
  * Resolved master page is used to validate its viewmodel
  * Resolved master page is used to assign view modul IDs (if any modules are present)

**Changes in this PR**:
* Master page is being resolved by `DefaultControlTreeResolver`, which is called by `ControlTreeResolverBase.ResolveTree(...)`
* Master page is validated in the `ResolveTree(...)` method as well
* ViewModul IDs are assigned in the `ResolveTree(...)` method too

### DIFF
--- a/src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
+++ b/src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
@@ -11,6 +11,8 @@ using DotVVM.Framework.Utils;
 using DotVVM.Framework.Binding;
 using System.Diagnostics.CodeAnalysis;
 using DotVVM.Framework.Compilation.Directives;
+using DotVVM.Framework.Compilation.Binding;
+using DotVVM.Framework.Compilation.ViewCompiler;
 
 namespace DotVVM.Framework.Compilation.ControlTree
 {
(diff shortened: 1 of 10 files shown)


src/Framework/Framework/Compilation/ControlTree/ControlTreeResolverBase.cs
```diff
@@ -58,22 +60,68 @@ public virtual IAbstractTreeRoot ResolveTree(DothtmlRootNode root, string fileNa
...
+                masterPage = ResolveMasterPage(fileName, masterPageDirective);
+                ValidateMasterPage(directiveMetadata.ViewModelType, masterPageDirective, masterPage);
+            }
+
+            var view = treeBuilder.BuildTreeRoot(this, viewMetadata, root, dataContextTypeStack, directiveMetadata.Directives, masterPage);
             view.FileName = fileName;
 
             if (directiveMetadata.ViewModuleResult is { })
             {
+                // Resolve viewmodul IDs
```
exyi, line 76:
```suggestion
                // Resolve viewmodule IDs
```

exyi (approved):

PR #1638, opened 2023-04-17
Add initial support for validation in static commands

acizmarik:
This PR adds support for manual validation in static commands.

### Changes
* `ModelState` now checks that `DotvvmRequestType` is `Command`
* `StaticCommandModelState` now checks that `DotvvmRequestType` is `StaticCommand`
   * `AddArgumentError(string argName, string message)` attaches error to the specified argument
   * `AddArgumentError<T>(Expression<Func<T>> expr, string message)` attaches error to a property of the specified argument
   * `AddRawError(string propertyPath)` attaches error to the specified object determined by the property path (identical property path notation is used as in command validation)
* `showValidationErrorsFromServer` now triggers `validationErrorsChanged` event
* All manually added validation errors (either from command or static command) are now displayed in JS console
* New API analyzer: _(DotVVM05) Possible reference of a local variable that is not an argument of a static command invocation_

### Example

```csharp
public class ViewModel
{
    public int Counter { get; set; } 
    public TestClass SomeObject { get; set; }
}

// Assume the following static command: {staticCommand: Method(Counter, SomeObject)}

[AllowStaticCommand]
public static void Method(int arg1, TestClass arg2)
{
    var modelState = new StaticCommandModelState();
    
    // Generates the following property path: '/Counter' (bound to first argument)
    modelState.AddArgumentError(nameof(arg1), "Invalid value");
    
    // Generates the following property path: '/SomeObject/Property' (bound to second argument)
    modelState.AddArgumentError(() => arg2.Property, "Another invalid value");

    // Interrupt execution if model state is not valid (contains errors)
    modelState.FailOnInvalidModelState();
}
```

Every argument is processed and mapped to a viewmodel property (if it can be determined). Additionally, there is a `AddRawError(...)` method that can be used to specify property paths manually. These paths will be always sent to client as-is and no preprocessing will take place.

### Tooling

It is not possible to enforce certain properties with type system. Therefore, there is now new analyzer:
validation1

* **DotVVM05** guards that any invocation of `AddArgumentError(...)` references actual method argument
* In order to avoid false positives, it is only active within a method that is annotated with `AllowStaticCommandAttribute`


src/Framework/Framework/Hosting/StaticCommandModelState.cs
```diff
@@ -0,0 +1,106 @@
...
+            ErrorsInternal.Add(error);
+            return error;
+        }
+
+        /// <summary>
+        /// Adds a new validation error with the given message on a property of the argument determined by its name
+        /// </summary>
+        /// <param name="propertyPath">Property path determining the nested value where to attach error</param>
+        /// <param name="message">Validation error message</param>
+        public StaticCommandArgumentValidationError AddRawArgumentError(string propertyPath, string message)
```
acizmarik, line 45:
```suggestion
        public StaticCommandValidationError AddRawArgumentError(string propertyPath, string message)
```

src/Framework/Framework/ViewModel/Validation/StaticCommandArgumentValidationError.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using DotVVM.Framework.Configuration;
+using Newtonsoft.Json;
+
+namespace DotVVM.Framework.ViewModel.Validation
+{
+    public class StaticCommandArgumentValidationError
```
acizmarik, line 7:
```suggestion
    public class StaticCommandValidationError
```

src/Framework/Core/ViewModel/AllowStaticCommandAttribute.cs
```diff
@@ -10,5 +9,11 @@ namespace DotVVM.Framework.ViewModel
     /// the attribute offers a decent protection against RCE in case the Asp.Net Core encryption keys are compromised. </remarks>
     public class AllowStaticCommandAttribute : Attribute
```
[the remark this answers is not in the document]
acizmarik: We decided that `ServerSide` and `ClientSide` is misleading. Better names are probably: `None`, `Manual` and `Automatic`

src/Framework/Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -376,6 +380,7 @@ public async Task ProcessStaticCommandRequest(IDotvvmRequestContext context)
                 CsrfProtector.VerifyToken(context, context.CsrfToken);
 
                 var knownTypes = postData["knownTypeMetadata"].Values<string>().ToArray();
+                var argumentPaths = postData["paths"]?.Values<string?>().ToArray();
```
tomasherceg, line 383: Can we rename this to `argumentPaths`?

src/Framework/Framework/Hosting/DotvvmPresenter.cs
```diff
@@ -447,6 +454,11 @@ public async Task ProcessStaticCommandRequest(IDotvvmRequestContext context)
                 
                 return commandResultOrNotYetComputedAwaitable;
             }
+            catch (TargetInvocationException e) when (e.InnerException is DotvvmInvalidStaticCommandModelStateException ex)
```
tomasherceg, line 457: Move this to the `(Exception ex)` branch after unwrapping the `TargetInvocationException`.

src/Framework/Framework/Hosting/DotvvmRequestContext.cs
```diff
@@ -75,7 +75,7 @@ public bool IsPostBack
         public ResourceManager ResourceManager => _resourceManager ?? (_resourceManager = Services.GetRequiredService<ResourceManager>());
 
         /// <summary>
-        /// Gets the <see cref="ModelState"/> object that manages validation errors for the viewmodel.
+        /// Gets the <see cref="ModelState"/> object that manages validation errors for the and static command arguments.
```
tomasherceg, line 78: Revert this, and explicitly mention that static commands are using a different class.

src/Framework/Framework/Resources/Scripts/postback/postbackCore.ts
```diff
@@ -101,7 +101,7 @@ export async function postbackCore(
                 return await processPostbackResponse(options, context, postedViewModel, initialState, response.result, response.response!);
             } catch (err: any) {
                 if (err instanceof DotvvmPostbackError) {
-                    throw err;
+                    throw new DotvvmPostbackError(err.reason);
```
tomasherceg, line 104: We are losing stack trace here.

tomasherceg (approved):

src/Framework/Framework/Resources/Scripts/postback/postbackCore.ts
exyi: I don't know why I made the change, it could be some hack since the method is async. Rethrowing using `throw err` works even in Chromium, so I'm reverting the change. If I rediscover the hack sometime, I'll make sure to add a comment explaining what it does :)

Issue #1479, opened 2022-09-28
Make it error to use DotvvmView instead of DotvvmMarkupControl


exyi:
When the user writes a dotvvm markup control with `.dothtml` extension and doesn't use `@baseType` directive, DotVVM will use DotvvmView as a base type. When the control is then initialized, there are multiple `DotvvmView`s in the page.

Easier solution is to make it an error to declare a markup control in such way. It's obviously breaking, so milestone 5.0

Alternative would be to use DotvvmMarkupControl as default base type, whenever the page is registered in DotvvmMarkupConfiguration. This is breaking too, if someone is using the page as view and also declared it as a control.

PR #1639, opened 2023-04-27
Fix issue where property directive compiled without errors on a regular page

acizmarik:
This PR fixes an issue where `@property ...` directive compiles just fine in a regular page. This feature is meant to be within markup controls only

### DIFF
--- a/src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
+++ b/src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
@@ -14,7 +14,7 @@ namespace DotVVM.Framework.Compilation.Directives
 
     public class BaseTypeDirectiveCompiler : DirectiveCompiler<IAbstractBaseTypeDirective, ITypeDescriptor>
     {
-        private readonly string fileName;
+        private readonly bool isMarkupControl;
         private readonly ImmutableList<NamespaceImport> imports;
 
         public override string DirectiveName => ParserConstants.BaseTypeDirective;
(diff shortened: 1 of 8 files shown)


src/Framework/Framework/Compilation/Directives/BaseTypeDirectiveCompiler.cs
```diff
@@ -107,19 +108,15 @@ IEnumerable<DothtmlDirectiveNode> propertyDirectives
...
         /// <summary>
         /// Gets the default type of the wrapper for the view.
         /// </summary>
         private ITypeDescriptor GetDefaultWrapperType()
-        {
-            if (fileName.EndsWith(".dotcontrol", StringComparison.Ordinal))
-            {
-                return new ResolvedTypeDescriptor(typeof(DotvvmMarkupControl));
-            }
+            => new ResolvedTypeDescriptor(IsMarkupControl() ? typeof(DotvvmMarkupControl) : typeof(DotvvmView));
```
exyi, line 116: Determining the wrapper type based on file extension is fragile. Now everything worked even when the base type was incorrectly inferred as DotvvmView (#1479), so people can use dothtml or html extension for controls. I think it would be better to determine if a page is control based on the registration in dotvvm configuration:

```
configuration.Markup.Controls.Any(c => fileName.Equals(c.Src, StringComparison.OrdinalIgnoreCase))
```

However, that would run into problem every time a new control is added in VS Extension, so I'd use an "extension OR configuration" condition in there (this class is overwritten anyway, so if this method is abstract it should be easy, right?)

tomasherceg (approved):

PR #1640, opened 2023-04-29
AutoUi GetPropertiesToDisplay no longer internal

mirecad:
It would be nice to access `GetPropertiesToDisplay` in all derived classes when building new custom form. I am building form that is based on grid, so I do not want to inherit from `AutoForm`.

### DIFF
--- a/src/AutoUI/Core/Controls/AutoFormBase.cs
+++ b/src/AutoUI/Core/Controls/AutoFormBase.cs
@@ -61,7 +61,7 @@ namespace DotVVM.AutoUI.Controls
         /// <summary>
         /// Gets the list of properties that should be displayed.
         /// </summary>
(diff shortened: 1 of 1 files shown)


exyi (commented):
:+1: I definitely agree, I think it's an omission. Would you want this as a 4.1 patch or 4.2-preview is fine?

We also use this method from another class (`AutoGridViewColumns`), you have to make it `protected internal` (or `public`) to make it compile.

mirecad:
@exyi 4.2-preview is fine. I did a workaround...

mirecad:
> We also use this method from another class (AutoGridViewColumns), you have to make it protected internal (or public) to make it compile.

@exyi  I did fix on web and havent tried the build. Fixed now...

exyi (approved):
:

PR #1641, opened 2023-05-02
Add try catch for deleting old files

Ropack:
`Directory.GetFiles()` invocation can cause DirectoryNotFoundException and crash the application


exyi (commented):
I think we should handle the "temp directory disappeared" problem more rigorously and also create the missing directory in the `StoreFileAsync` function

tomasherceg:
Many thanks for the contribution. I've embedded this change into #1655 so I am closing this PR.

PR #1643, opened 2023-05-10
Fix/range client validation

### LINKED ISSUE #1642: Client-side validation: `Nullable<>` properties with `Range` attribute are implicitly required for numeric inputs
If a property of type `int?` has a `Range` attribute, the validation always fails when the field is left empty, even though the property is not `Required`. This is probably specific to inputs with `Type="Number"`.

Please note that the value is sent correctly to the server as `null`. The issue is with only the client-side validation treating the value as `0`.
(However, non-numeric values are both validated as and sent to the server as `0`.)

Example code:
```C#
public class MyForm
{
    [Range(1, 999, ErrorMessage = "Invalid number")]
    public int? SomeNumber { get; set; }
}

public class MyPageViewModel : MasterPageViewModel
{
    // Constructor...

    public override async Task Init()
    {
        await base.Init();
        Form = new(); // Load from service
    }

    public MyForm Form { get; }

    public async Task SaveForm()
    {
        // Send form to service
    }
}
```
```html
<%--...--%>
<dot:Content ContentPlaceHolderID="MainContent">
        <div class="form-group" Validator.InvalidCssClass="has-error" Validator.Value="{value: Form.SomeNumber}">
            <label for="SomeNumber">Some number:</label>
            <dot:TextBox html:id="SomeNumber" Text="{value: Form.SomeNumber}"
                         class="form-control" Type="Number" min="1" max="999" />
        </div>
        <div class="d-flex align-items-center">
            <dot:Button html:id="submitBtn" Click="{command: SaveForm()}"
                        ButtonTagName="button" IsSubmitButton="true" PostBack.Concurrency="Deny"
                        class="btn btn-danger align-self-start">
                Save
            </dot:Button>
            <dot:ValidationSummary class="text-danger mb-0" />
        </div>
</dot:Content>
```

If the input is left empty, the error `Invalid number` appears.

Workaround 1: Disable client-side validation for this field.
Workaround 2: Instead of using `Range`, make `MyForm` implement `IValidatableObject` and customize the validation.

tomasherceg:
Resolves #1642

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/validation/validators.ts
+++ b/src/Framework/Framework/Resources/Scripts/validation/validators.ts
@@ -16,13 +16,6 @@ export const regex: DotvvmValidator = {
     }
 }
 
-export const intRange: DotvvmValidator = {
-    isValid(val: any, context): boolean {
-        const [from, to] = context.parameters;
-        return val % 1 === 0 && val >= from && val <= to;
-    }
-}
-
 export const enforceClientFormat: DotvvmValidator = {
     isValid(value: any, context, property): boolean {
         const [allowNull, allowEmptyString, allowEmptyStringOrWhitespaces] = context.parameters;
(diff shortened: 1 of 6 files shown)


exyi (approved):
:+1:

PR #1647, opened 2023-05-15
UI Tests for ko.options.deferUpdates

exyi:


exyi:
I'd say it works. One test which is related is failing on CI in Firefox, but it works when I run it locally.

tomasherceg (approved):

PR #1650, opened 2023-05-31
Unify client/server evaluation in HtmlGenericControl

### LINKED ISSUE #1635: Cannot use _js.Invoke in Visible binding
```
<div class="d-f fd-c color-text ai-c w-100 mt-md mb-md video-status"
     Visible="{value: _js.Invoke<VideoContentState>("getVideoContentState", State, _root.CallerStatus) == VideoContentState.WaitForCaller}">
    <img src="/Images/Incident/video-loading.svg" alt="Video loading" />
    <p class="text-md">{{resource: PortalTexts.Incident_Video_StartingVideoStream}}</p>
    <p class="text-sm ta-c">{{resource: PortalTexts.Incident_Video_WaitingForUserAcceptStream}}</p>
</div>
```

Cannot invoke JS command server-side: getVideoContentState

Of course, `_js.Invoke` cannot be called on the server :-), but it is not necessary. RenderSettings.Mode="Client" doesn't help either.

Workaround: `class="{value: _js.Invoke<VideoContentState>("getVideoContentState", State, _root.CallerStatus) == VideoContentState.WaitForCaller ? "d-f" : "d-n"}"`

exyi:
Unifies how client/server rendering mode behaves in all HtmlGenericControl properties to "always render both".

* Applies to Visible, Attributes, CssClasses, CssStyles
* When property contains value binding, knockout binding is added
* The property is always rendered server-side
* When value binding evaluation fails on server, the exception is suppressed.

The behavior is unchanged in `CssClasses` and `CssStyles`, I have added the try catch to Visible and Attributes are now rendered in server-side mode. Technically, this is a **breaking change**, so we should definitely announce it, but the impact should be small - In most cases, what is rendered server-side is irrelevant to normal page behavior, it only affect indexers and page load performance.

I have also changed how returning `null` behaves in Attributes. Before it added an empty attribute, which is inconsistent with the client-side behavior of knockout JS (and it is also annoying, since most attributes in client-side templates will return `null` and thus would be present empty). Now, `null` returned from a binding skips rendering of that attribute. Behavior of hardcoded `null`s is unchanged - it will still render an empty attribute.

resolves #1635

### DIFF
--- a/src/Tests/ControlTests/HtmlGenericControlTests.cs
+++ b/src/Tests/ControlTests/HtmlGenericControlTests.cs
@@ -43,6 +43,8 @@ namespace DotVVM.Framework.Tests.ControlTests
             }
         }
 
+        // Test that all HtmlGenericControl properties are bindable and the value binding is evaluated both client-side and server-side, regardless of the RenderMode.
+        // Errors in value bindings are suppressed, to allow bindings which can't be evaluated server-side
         [DataTestMethod]
         [DataRow(RenderMode.Client)]
         [DataRow(RenderMode.Server)]
@@ -55,16 +57,6 @@ namespace DotVVM.Framework.Tests.ControlTests
             var str = RenderToString(control);
             Assert.AreEqual("""<div data-bind='attr: { "data-test": dotvvm.viewModules.call("p1", "testMethod", [], false) }'></div>""", str);
         }
-
-        // [TestMethod]
-        // public void ServerSideInvokeFails_Attribute()
-        // {
-        //     var control = new HtmlGenericControl("div")
-        //         .SetProperty(RenderSettings.ModeProperty, RenderMode.Server)
-        //         .SetAttribute("data-test", bindingService.Cache.CreateValueBinding<string>("_js.Invoke<string>('testMethod')", defaultDataContext));
-
-        //     var exception = Assert.ThrowsException<DotvvmControlException>(() => RenderToString(control));
-        // }
 
         [DataTestMethod]
         [DataRow(RenderMode.Client)]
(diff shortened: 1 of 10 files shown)


src/Tests/ControlTests/HtmlGenericControlTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        [DataRow(RenderMode.Client)]
+        [DataRow(RenderMode.Server)]
+        public void ValueBinding_Attribute(RenderMode renderMode)
+        {
+            var control = new HtmlGenericControl("div")
+                .SetProperty(RenderSettings.ModeProperty, renderMode)
+                .SetAttribute("data-test", bindingService.Cache.CreateValueBinding<string>("String", defaultDataContext));
+
+            var str = RenderToString(control);
+            // TODO: should it behave differently for server-side rendering?
```
tomasherceg, line 116: Please remove this comment - it's already resolved

src/Tests/ControlTests/HtmlGenericControlTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            var control = new HtmlGenericControl("div")
+                .SetProperty(RenderSettings.ModeProperty, renderMode)
+                .SetAttribute("data-test", bindingService.Cache.CreateValueBinding<string>("_js.Invoke<string>('testMethod')", defaultDataContext));
+
+            var str = RenderToString(control);
+            Assert.AreEqual("""<div data-bind='attr: { "data-test": dotvvm.viewModules.call("p1", "testMethod", [], false) }'></div>""", str);
+        }
+
+        // [TestMethod]
```
tomasherceg, line 59: Please remove

tomasherceg (approved):

PR #1655, opened 2023-06-10
Fixes in uploaded file storage

tomasherceg:
Adapts changes from #1641, and also makes sure that the temp directory exists before the file is stored.

### DIFF
--- a/src/Framework/Framework/Storage/FileSystemUploadedFileStorage.cs
+++ b/src/Framework/Framework/Storage/FileSystemUploadedFileStorage.cs
@@ -25,16 +25,27 @@ namespace DotVVM.Framework.Storage
         public FileSystemUploadedFileStorage(string tempDirectory, TimeSpan autoDeleteInterval)
         {
             TempDirectory = tempDirectory;
-            if (!Directory.Exists(tempDirectory))
-            {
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Storage/FileSystemUploadedFileStorage.cs
```diff
@@ -24,23 +25,35 @@ public class FileSystemUploadedFileStorage : IUploadedFileStorage, IDisposable
...
+        {
+            if (!Directory.Exists(TempDirectory))
+            {
+                try
+                {
+                    Directory.CreateDirectory(TempDirectory);
+                }
+                catch (IOException)
+                {
+                    throw new Exception($"The {nameof(FileSystemUploadedFileStorage)} couldn't create a directory {TempDirectory}. Make sure the application has write permissions for this path.");
```
exyi, line 45:
```suggestion
                catch (IOException e)
                {
                    throw new Exception($"The {nameof(FileSystemUploadedFileStorage)} couldn't create a directory {TempDirectory}. Make sure the application has write permissions for this path.", e);
```

exyi (approved):
:+1:

PR #1674, opened 2023-07-14
Added ASP.NET Web Forms adapters package

tomasherceg:
This package helps in the process of [migrating Web Forms apps to the new .NET using DotVVM](https://www.dotvvm.com/modernize).

* The `<webforms:HybridRouteLink>` control has the same functionality as the `RouteLink` control, but it falls back to Web Forms routes when a requested DotVVM route doesn't exist (which means that the page hasn't been migrated yet).
* The `RedirectToRouteHybrid` extension method of the `IDotvvmRequestContext` also falls back to the Web Forms route in case the DotVVM route doesn't exist.

The package also works on the .NET Standard 2.1 branch, where these fallbacks are disabled (because of the absence of `System.Web`). This allows to keep using the API even after the migration is finished.


src/Adapters/WebForms/Controls/RouteLinkCapability.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel;
+using DotVVM.Framework.Binding;
+
+namespace DotVVM.Adapters.WebForms.Controls
+{
+    [DotvvmControlCapability()]
+    public sealed class RouteLinkCapability
+    {
```
exyi, line 9: I think we could add this capability to the built-in RouteLink control.

src/Adapters/WebForms/RedirectHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+
+#if NETFRAMEWORK
+using System.Web.Routing;
+#endif
+
+namespace DotVVM.Framework.Hosting
+{
+    public static class DotvvmRequestContextExtensions
+    {
+        public static void RedirectToRouteHybrid(this IDotvvmRequestContext context, string routeName, object routeValues = null, string urlSuffix = null, object query = null)
```
exyi, line 15: I'd prefer to name it something like `RedirectToWebFormsRoute`, or at least name the class as WebForms...Extensions. The control is fine at will always have the `webforms:` prefix.

src/Adapters/WebForms/Controls/RouteLinkCapability.cs
tomasherceg: Maybe put this in a separate PR.

src/Adapters/WebForms/RedirectHelper.cs
tomasherceg: Let's rename to `WebFormsExtensions` and make this method obsolete on .NET Core branch.

src/Adapters/WebForms/Controls/HybridRouteLink.cs
```diff
@@ -0,0 +1,61 @@
...
+                throw new DotvvmControlException($"Route '{routeLinkCapability.RouteName}' does not exist.");
+            }
+        }
+
+        private static DotvvmControl GenerateDotvvmRouteLink(HtmlCapability htmlCapability, TextOrContentCapability textOrContent, RouteLinkCapability routeLinkCapability)
+        {
+            var link = new RouteLink()
+                .SetCapability(htmlCapability)
+                .SetCapability(textOrContent)
+                .SetProperty(r => r.RouteName, routeLinkCapability.RouteName)
```
exyi, line 53: You should now be able to just set the routelink capability

```suggestion
                .SetCapability(routeLinkCapability)
```

src/Adapters/WebForms/Controls/WebFormsLinkUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+            var routeUrlExpression = $"dotvvm.buildRouteUrl({KnockoutHelper.MakeStringLiteral(webFormsRoute.Url)}, {{{parametersExpression}}})";
+            return routeUrlExpression;
+        }
+
+        private static string GenerateUrlSuffixExpression(DotvvmControl container, RouteLinkCapability routeLinkCapability)
+        {
+            var urlSuffixBase = routeLinkCapability.UrlSuffix?.GetJsExpression(container) ?? "\"\"";
+            var queryParams = routeLinkCapability.QueryParameters
+                .OrderBy(p => p.Key)
+                .Select(p => $"{KnockoutHelper.MakeStringLiteral(p.Key.ToLowerInvariant())}: {p.Value.GetJsExpression(container)}")
```
exyi, line 113: Are you sure we want to lowercase all query parameters? Those might be case sensitive.

exyi (approved):

PR #1682, opened 2023-08-01
Switch custom primitive types to a marker interface

exyi:
This allows for efficient checks if a value is primitive without the need for cache hashtable containing all types we encounter

### DIFF
--- a/src/Framework/Core/ViewModel/CustomPrimitiveTypeAttribute.cs
+++ b/src/Framework/Core/ViewModel/CustomPrimitiveTypeAttribute.cs
@@ -3,6 +3,7 @@
 namespace DotVVM.Framework.ViewModel;
 
 [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct)]
+[Obsolete("Implement IDotvvmPrimitiveType.", error: true)]
 public class CustomPrimitiveTypeAttribute : Attribute
 {
 }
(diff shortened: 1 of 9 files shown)


src/Framework/Core/ViewModel/CustomPrimitiveTypeAttribute.cs
```diff
@@ -3,6 +3,7 @@
 namespace DotVVM.Framework.ViewModel;
```
tomasherceg, line 3: Remove the class completely

src/Framework/Framework/Utils/ReflectionUtils.cs
```diff
@@ -374,31 +375,30 @@ public static bool IsDotvvmNativePrimitiveType(Type type)
...
             return PrimitiveTypes.Contains(type)
-                || (IsNullableType(type) && IsDotvvmNativePrimitiveType(type.UnwrapNullableType()))
+                || (IsNullableType(type) && IsPrimitiveType(type.UnwrapNullableType()))
                 || type.IsEnum
-                || TryGetCustomPrimitiveTypeRegistration(type) is {};
+                || IsCustomPrimitiveType(type);
         }
 
-        private static CustomPrimitiveTypeRegistration? TryDiscoverCustomPrimitiveType(Type type)
+        private static CustomPrimitiveTypeRegistration TryDiscoverCustomPrimitiveType(Type type)
```
tomasherceg, line 399: Remove Try from the name since the method can crash

tomasherceg (approved):

PR #1683, opened 2023-08-01
Run module initialization in dotvvm.init event

### LINKED ISSUE #1592: page view module is initialized before dotvvm
DotVVM currently initializes a view module as soon as `dotvvm.viewModules.init` is called. Since we render the script calling `dotvvm.viewModules.init({ ... })` in the header, it will be called before dotvvm has been initialized. This makes it impossible to use `dotvvm.state` in the module init function without wrapping it with `dotvvm.events.init.subscripe`.

cc @martindybal

exyi:
This allows using dotvvm.state (etc.) already in the module constructor.

resolves  #1592

### DIFF
--- a/src/Framework/Framework/ResourceManagement/ViewModuleInitResource.cs
+++ b/src/Framework/Framework/ResourceManagement/ViewModuleInitResource.cs
@@ -29,7 +29,15 @@ namespace DotVVM.Framework.ResourceManagement
             this.ReferencedModules = referencedModules.ToArray();
             this.Dependencies = dependencies;
 
-            this.registrationScript = string.Join("\r\n", this.ReferencedModules.Select(m => $"dotvvm.viewModules.init({KnockoutHelper.MakeStringLiteral(m)}, {KnockoutHelper.MakeStringLiteral(viewId)}, document.body);"));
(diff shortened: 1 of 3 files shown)


src/Framework/Framework/ResourceManagement/ViewModuleInitResource.cs
```diff
@@ -29,7 +29,15 @@ public ViewModuleInitResource(string[] referencedModules, string name, string vi
...
             this.Dependencies = dependencies;
 
-            this.registrationScript = string.Join("\r\n", this.ReferencedModules.Select(m => $"dotvvm.viewModules.init({KnockoutHelper.MakeStringLiteral(m)}, {KnockoutHelper.MakeStringLiteral(viewId)}, document.body);"));
+            var initCalls = this.ReferencedModules.Select(m => $"dotvvm.viewModules.init({KnockoutHelper.MakeStringLiteral(m)}, {KnockoutHelper.MakeStringLiteral(viewId)}, document.body);");
+
+            // Run the module init in the init event
+            // * dotvvm.state will be available
+            // * executed before applying bindings to the controls, so the page module will initialize before control modules 
+            this.registrationScript =
+                "dotvvm.init.subscribeOnce(() => {\n" +
```
tomasherceg, line 38: `dotvvm.events.init`

tomasherceg (approved):

PR #1686, opened 2023-08-04
ExtractGenericArgumentDataContextChangeAttribute attrribute VS Extension compatibility

Mylan719:
It used naked type making it impossible to integrate in VS Extension.

### DIFF
--- a/src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
+++ b/src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
@@ -47,11 +47,11 @@ public class ExtractGenericArgumentDataContextChangeAttribute : DataContextChang
     public override Type? GetChildDataContextType(Type dataContext, DataContextStack controlContextStack, DotvvmBindableObject control, DotvvmProperty? property = null)
     {
         var implementations = ReflectionUtils.GetBaseTypesAndInterfaces(dataContext)
-            .Where(i => i.IsGenericType && i.GetGenericTypeDefinition() == GenericType)
+            .Where(i => i.IsGenericType && new ResolvedTypeDescriptor(i.GetGenericTypeDefinition()).IsEqualTo(GenericType))
             .ToList();
         if (implementations.Count == 0)
         {
-            throw new Exception($"The data context {dataContext} doesn't implement {GenericType.CSharpFullName}!");
+            throw new Exception($"The data context {dataContext.ToCode()} doesn't implement {GenericType.CSharpFullName}!");
         }
         else if (implementations.Count > 1)
         {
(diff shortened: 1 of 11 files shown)


src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
```diff
@@ -48,11 +49,11 @@ public ExtractGenericArgumentDataContextChangeAttribute(Type genericType, int ty
         var implementations = ReflectionUtils.GetBaseTypesAndInterfaces(dataContext).ToList();
         if (implementations.Count == 0)
         {
-            throw new Exception($"The data context {dataContext} doesn't implement {GenericType}!");
+            throw new Exception($"The data context {dataContext} doesn't implement {GenericType.CSharpFullName}!");
```
exyi, line 52:
```suggestion
            throw new Exception($"The data context {dataContext.ToCode()} doesn't implement {GenericType.CSharpFullName}!");
```

Not your mistake, just noticed :]

src/Framework/Framework/Compilation/ControlTree/Resolved/ResolvedTypeDescriptor.cs
```diff
@@ -138,10 +138,11 @@ public ITypeDescriptor MakeGenericType(params ITypeDescriptor[] typeArguments)
             return new ResolvedTypeDescriptor(genericType);
         }
 
-        public IEnumerable<ITypeDescriptor> FindGenericImplementations(Type genericType)
+        public IEnumerable<ITypeDescriptor> FindGenericImplementations(ITypeDescriptor genericType)
         {
-            return ReflectionUtils.GetBaseTypesAndInterfaces(Type)
-                .Where(i => i.IsGenericType && i.GetGenericTypeDefinition() == genericType)
+            var generic = (genericType as ResolvedTypeDescriptor)?.Type ?? throw new InvalidOperationException($"Only {nameof(ResolvedTypeDescriptor)} sould be used as a parameter.");
```
exyi, line 143: I think we prefer to use the `ToSystemType` function. In case this method is called in VS Extension on a stray ResolvedTypeDescriptor and the Roslyn-backed TypeDescriptor, it will somewhat work.

```suggestion
            var generic = ResolvedTypeDescriptor.ToSystemType(genericType);
```

exyi (approved):

src/Framework/Framework/Binding/ExtractGenericArgumentDataContextChangeAttribute.cs
```diff
@@ -24,7 +25,7 @@ public ExtractGenericArgumentDataContextChangeAttribute(Type genericType, int ty
             throw new ArgumentOutOfRangeException(nameof(typeArgumentIndex), $"The {nameof(typeArgumentIndex)} is not a valid index of a type argument!");
         }
 
-        GenericType = genericType;
+        GenericType = new ResolvedTypeDescriptor(genericType);
         TypeArgumentIndex = typeArgumentIndex;
         Order = order;
     }
```
exyi, line 31: I think you'll need add `ExtractGenericArgumentDataContextChangeAttribute(ITypeDescriptor genericType, int typeArgumentIndex, int order = 0)` overload

exyi (approved):

PR #1696, opened 2023-09-04
More reasonable unclosed tag handling

exyi:
Enabling the warnings in #1659 reveals that our handling of unclosed tags is not ideal. This patch is attempt to improve it, while maintaining reasonable level of backwards compatibility.

The only behavior change is what happens when `</x>` has no opening tag
* previous: close currently open elements, then insert `<x />`
* new: insert `<x />` into the current element

a warning is issued in both cases. The new logic avoids a major problem of one accidental end tag closing the entire hierarchy, including the `<body>` and `<html>` elements. Since we insert script references at the end of body, this puts the scripts and viewmodel wherever the closing tag was. This change has the potential to break something

Other changes only affect the issued warning:
* no warning on an unclosed tags like img, meta, br (which are always self-closing in HTML)
* no warning on automatically closed tags like p, td. In HTML it's perfectly valid let these tags close automatically, so we can also consider this valid.

### DIFF
--- a/src/Tests/Parser/Dothtml/DothtmlParserTests.cs
+++ b/src/Tests/Parser/Dothtml/DothtmlParserTests.cs
@@ -260,9 +260,12 @@ test";
         {
             var markup = @"<dot:ContentPlace Holder DataContext=""sdads"">
 </ dot:ContentPlaceHolder > ";
-            var nodes = ParseMarkup(markup).Content;
+            var root = ParseMarkup(markup);
+            var nodes = root.Content;
 
-            Assert.AreEqual(3, nodes.Count);
+            Assert.IsTrue(root.EnumerateNodes().Any(n => n.HasNodeErrors || n.Tokens.Any(t => t.HasError)), "this is invalid code, there should be error on some node");
+
+            Assert.AreEqual(2, nodes.Count);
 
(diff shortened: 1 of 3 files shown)


src/Tests/Parser/Dothtml/DothtmlParserTests.cs
```diff
@@ -260,21 +260,23 @@ public void DothtmlParser_Invalid_SpaceInTagName()
...
             var markup = @"<dot:ContentPlace Holder DataContext=""sdads"">
 </ dot:ContentPlaceHolder > ";
-            var nodes = ParseMarkup(markup).Content;
+            var root = ParseMarkup(markup);
+            var nodes = root.Content;
 
-            Assert.AreEqual(3, nodes.Count);
+            Assert.IsTrue(root.EnumerateNodes().Any(n => n.HasNodeErrors || n.Tokens.Any(t => t.HasError)), "this is invalid code, there should be error on some node");
+
+            Assert.AreEqual(2, nodes.Count);
```
exyi, line 268: I also slightly changed what happens in the case of incomplete close tag - it now closes the current element, regarless of its name. However, this cannot break any running code, since incomplete end tag has a token error and thus the compilation fails.

tomasherceg (approved):

PR #1704, opened 2023-09-28
Fixed bug in ExtractGenericArgumentDataContextChangeAttribute

tomasherceg:
There was a bug in `ExtractGenericArgumentDataContextChange` attribute - the method overload with `Type` wasn't comparing the generic arguments correctly.
Also, when the DataContext was directly the requested type, it didn't match because we inspected only implemented interfaces and base types, but not self.


exyi (approved):
:+1:

exyi:
I'd not mind if we had a unit test for this, I think it should be quite easy to just create the attribute, and call the two methods.

PR #1713, opened 2023-10-17
support Linq methods on ImmutableArray + added Errors tab to Compilation Page

exyi:
Original goal was to add an "Errors" tab to the compilation which lists all failed pages, controls and master pages. Saves few clicks and scrolling when checking the compilation page.

image

However, I had to fix Linq translations on `ImmutableArray`. ImmutableArray is a struct so it has custom Linq method overloads to avoid allocations. These are preferred by the compiler over the `IEnumerable<T>` methods.

I also added .AsEnumerable and few other conversion methods to help users work around similar issues.


tomasherceg (approved):
Please update the [docs page](https://github.com/riganti/dotvvm-docs/blob/4.0/Pages/concepts/data-binding/supported-expressions.md)

PR #1742, opened 2023-11-22
Added repro for a bug with passing commands in markup controls

tomasherceg:
We found a strange bug - probably caused by missing validation of delegates passed to DotVVM properties.

I've added a repro `/FeatureSamples/MarkupControl/CommandAsPropertyPage` but I cannot find the place where to fix it.

### DIFF
--- a/src/Framework/Framework/Compilation/Javascript/ParametrizedCode.cs
+++ b/src/Framework/Framework/Compilation/Javascript/ParametrizedCode.cs
@@ -50,9 +50,14 @@ namespace DotVVM.Framework.Compilation.Javascript
 
         // TODO(exyi): add WriteTo(StringBuilder)
         /// <summary>
-        /// Converts this to string and assigns all parameters using `parameterAssignment`. If there is any missing, exception is thrown.
+        /// Converts this to string and assigns all parameters using `parameterAssignment`.
         /// </summary>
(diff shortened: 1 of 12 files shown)


exyi:
I fixed the error message, otherwise the behavior looks "fine" to me.

1. Normally, it should definitely error out, since the `Click` command has type `Func<string, bool, Task>` - i.e. it requires arguments, and the button provides none.
2. You call the command from a value binding, so it makes sense that evaluating the value binding will call the command. As far as I understand it, _control.Click.ToString() invokes the value binding, which invokes the command.

Using the value binding to call commands is definitely not recommended, commands should only be invoked from command or staticCommand bindings. `staticCommand` is preferred to call control properties, since it works for calling both `command` and `staticCommand`s

tomasherceg:
The `ToString()` workaround in the debug mode probably works because of the bug in .NET - see https://github.com/dotnet/runtime/issues/96385

tomasherceg:
I have to add tests here.

exyi (approved):

tomasherceg:
We need to add additional checks:

1. If DotVVM property is `Command/Action/Func` or any delegate, we should only allow `command/staticCommand` bindings.
2. If DotVVM control property value is delegate, throw a reasonable exception.
3. Disable lambda conversion except for in commands:
    image

PR #1743, opened 2023-11-23
Remove reference to `Microsoft.CodeAnalysis.CSharp` in AspNetCore package

OronDF343:
This PR removes an unnecessary reference to `Microsoft.CodeAnalysis.CSharp` in the project `DotVVM.Framework.Hosting.AspNetCore`. This reference is unused yet brings a lot of transitive references to legacy `System` packages from .NET Core 1.x.


exyi:
Thank you! We forgot to remove the reference after we rewrote the compiler to target Linq.Expressions instead of C#.

We'll have to release this as a 4.3 preview version. We should be rather strict about breaking changes in 4.2 patches, and somebody could depend on `Microsoft.CodeAnalysis.CSharp` being a transitive dependency of DotVVM

PR #1746, opened 2023-11-28
Added a failing test for conditional operator nullable conversions

tomasherceg:
I have added a test case `BindingCompiler_ImplicitConversion_ConditionalNullableAndNonNullable` for conditional operators that cannot compile:

```
boolValue ? nullableDouble : double
```

It seems to me that the [VisitConditionalExpression](https://github.com/riganti/dotvvm/blob/main/src/Framework/Framework/Compilation/Binding/ExpressionBuildingVisitor.cs#L314) can do both implicit conversions (from nullable double to double and vice versa) and cannot decide on the direction.

Shall we be even able to do an implicit conversion from nullable to non-nullable? 
I am not sure if I can fix it in the logic of implicit conversion, or if it is a special case here.

### DIFF
--- a/src/Tests/Binding/BindingCompilationTests.cs
+++ b/src/Tests/Binding/BindingCompilationTests.cs
@@ -836,6 +836,17 @@ namespace DotVVM.Framework.Tests.Binding
         }
 
         [TestMethod]
+        public void BindingCompiler_ImplicitConversion_ConditionalNullableAndNonNullable()
+        {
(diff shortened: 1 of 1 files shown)


exyi:
I fixed it by adding a special case. I didn't know we allow the "from nullable" implicit conversion, but we can't disallow this in a patch release, so I added a special case to the VisitConditionalExpression function.

Also, I had to adjust the testcase for `B ? null : DoubleProp`. This is not allowed by standard C#. It (unfortunately) works in DotVVM, but returns `object`, not `double?`. Again, we can't disallow it in a patch version. Changing the return type to `double?` is also a bit risky (and not trivial), but doable. We can either forbid this in v5, or make it return `double?` in v4.3 or v5, whichever comes sooner.

PR #1756, opened 2023-12-22
Fixed globalize resource visitor and added the test

tomasherceg:
The `GlobalizeResourceVisitor` had a bug that sometimes did not correctly detect the need to add `globalize` JS resource.

When the first binding that requires `globalize` resource was detected, a boolean flag was set, but the `RequiredResource` control was being added when the visitor reached the end of a template property, or the end of a `Content` control, or the end of the entire view. 

```html
{{value: BindingWithGlobalizeRequirement}}
<dot:SomeControl>
  <SomeTemplate>
     <!-- the resource is added here -->
  </SomeTemplate>
</dot:SomeControl>
```

This was leading to a behavior when the resource was added into a random template which may not get into the page control tree eventually (e. g. when it is in `AuthenticatedView` or something like that).

I am not sure if the fix is good - I removed adding the resource in the templates, so it is added at the end of the `Content` or the view itself. Maybe it would be better to add it before or after the control which required the particular binding.

### DIFF
--- a/src/Framework/Framework/Compilation/GlobalizeResourceVisitor.cs
+++ b/src/Framework/Framework/Compilation/GlobalizeResourceVisitor.cs
@@ -39,11 +39,6 @@ namespace DotVVM.Framework.Compilation
             base.VisitControl(control);
         }
 
-        public override void VisitPropertyTemplate(ResolvedPropertyTemplate propertyTemplate)
-        {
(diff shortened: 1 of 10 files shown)


exyi (approved):
Good catch! The intent was to analyze the template separately to avoid adding resources in the (rare) case it's required only in an unused template. However, for that we'd need to nest the flags in the `private void Visit<TNode>` method. I guess it's not worth the effort/complexity...

PR #1758, opened 2024-01-04
Attribute value initializer of PropertyDirective changes

Mylan719:
Now the initializer of the attribute supports half written values for the VS extension.
The validation before was way too strict and replaced invalid values with empty literal.

### DIFF
--- a/src/Tests/AssertEx.cs
+++ b/src/Tests/AssertEx.cs
@@ -10,7 +10,7 @@ namespace DotVVM.Framework.Tests
 {
     public static class AssertEx
     {
-        public static void AssertNode(BindingParserNode node, string expectedDisplayString, int start, int length, bool hasErrors = false)
+        public static void BindingNode(BindingParserNode node, string expectedDisplayString, int start, int length, bool hasErrors = false)
         {
             Assert.AreEqual(expectedDisplayString, node.ToDisplayString(), $"Node {node.GetType().Name}: display string incorrect.");
             Assert.AreEqual(start, node.StartPosition, $"Node {node.GetType().Name}: Start position incorrect.");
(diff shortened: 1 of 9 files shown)


src/Tests/AssertEx.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using DotVVM.Framework.Compilation.Parser.Binding.Parser;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DotVVM.Framework.Tests
+{
+    public static class AssertEx
+    {
+        public static void AssertNode(BindingParserNode node, string expectedDisplayString, int start, int length, bool hasErrors = false)
```
exyi, line 13: I'd rename this to `BindingNode`, it's on longer in "binding" class, but it's now in "assert" class.

src/Tests/Runtime/ControlTree/DefaultControlTreeResolver/PropertyDirectiveTests.cs
```diff
@@ -45,19 +50,36 @@ @viewModel object
...
+            var nodes = property.InitializerSyntax.EnumerateChildNodes();
             Assert.IsFalse(nodes.Any(n => n == property.InitializerSyntax));
+        }
+
+        [TestMethod]
+        public void ResolvedTree_PropertyDirectiveArrayInicializerAndAttributes_ResolvedCorrectly()
+        {
+            var root = ParseSource(@$"
+@viewModel object
+@property string[] MyProperty=["""",""""], MarkupOptionsAttribute.Required = true, MarkupOptionsAttribute.AllowBinding = false");
```
exyi, line 73: You could use the `"""` string literal to avoid the double-quoting

exyi (approved):

src/Tests/AssertEx.cs
Mylan719: Done

exyi (approved):
Yes, it's good to go. I already approved of this yesterday, the comments were just nitpicks :]

PR #1762, opened 2024-01-16
Refactor compilation diagnostics, add them to compilation status page

exyi:
This is a large commit of various interdependent changes with the simple goal of adding warnings to the compilation status page.

* Non-critical errors from tokens are now reported as warnings. Before, they were ignored completely.
* Added IDiagnosticsCompilationTracer interface
   - it allows watching the compilation process, for example observing changes made by various visitors or in this case getting the warnings when the compilation otherwise succeeds
* Added DotvvmCompilationDiagnostic and DotvvmCompilationSourceLocation records
   - it designed to be a generic class for warnings and errors from the DotHTML compilation. It contains references to the syntax and resolved trees, so it should not be stored long-term, but provides all the possible detail about each reported error/warning.
* BindingCompilationException is refactored to use the DotvvmCompilationDiagnostics and support any number of them.
   - plus, IDotvvmException compatibility is added
   - the exception will present as the "first" / primary error for better compatibility
* ErrorCheckingVisitor is now smarter about binding errors, the BindingToken ranges are mapped into the DothtmlToken so we can underline only the affected part of the binding. This is somewhat important for displaying diagnostics of multiline bindings in the compilation page. Multiple errors can be collected from the error.
* ViewCompilationService now collects the warnings and errors from the tracer and DotvvmCompilationException
* And finally, the compilation displays up to 8 errors and warnings encountered during the page compilation under each row. Warnings tab was also added which shows only the views with at least one warning/error

image
image


src/Framework/Framework/Diagnostics/CompilationDiagnostic.dotcontrol
```diff
@@ -0,0 +1,14 @@
+<%-- @property DotVVM.Framework.Compilation.DotHtmlFileInfo.CompilationDiagnosticViewModel Diagnostic --%>
```
tomasherceg, line 1: Remove commented-out code

PR #1772, opened 2024-02-12
Properties with the same name defined by `@property` directive

Mylan719:
If user defines two properties of the same name in the same markup file, there should be an error.


exyi (approved):
:+1:

PR #1776, opened 2024-02-18
New dot:ModalDialog control, wrapper for <dialog>

### LINKED ISSUE #1708: dot:ModalDialog control
There is a native HTML dialog component, which would be useful to wrap into a dot:ModalDialog control. The documentation for the dialog component is at this link: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

exyi:
The control closes or shows the modal based on an Open property. The modal is always shown using the .showModal() method, non-modal dialog is already accessible by binding the open attribute of dialog HTML element.
The Open property may either be a boolean or a nullable object, the dialog is shown if the value isn't false nor null.

On close event, false or null is written back into the Open property. Otherwise, we'd quickly have inconsistent viewModel whenever the user closes the dialog with ESC.
Close event is also provided for explicit event handling.

We also optionaly provide a helper for implementing "close after backdrop click" functionality. It is not supported by the dialog element natively and could not be otherwise implemented without writing custom JS.
It is enabled by setting CloseOnBackdropClick=true

Resolves  #1708

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/binding-handlers/modal-dialog.ts
+++ b/src/Framework/Framework/Resources/Scripts/binding-handlers/modal-dialog.ts
@@ -1,0 +1,41 @@
+export default {
+    "dotvvm-modal-open": {
+        init(element: HTMLDialogElement, valueAccessor: () => any) {
+            element.addEventListener("close", () => {
+                const value = valueAccessor();
+                if (ko.isWriteableObservable(value)) {
+                    // if the value is object, set it to null
+                    value(typeof value.peek() == "boolean" ? false : null)
+                }
+            })
+        },
+        update(element: HTMLDialogElement, valueAccessor: () => any) {
+            const value = ko.unwrap(valueAccessor()),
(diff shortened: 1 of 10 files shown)


src/Framework/Framework/Resources/Scripts/binding-handlers/modal-dialog.ts
```diff
@@ -0,0 +1,41 @@
...
+            if (shouldOpen != element.open) {
+                if (shouldOpen) {
+                    element.showModal()
+                } else {
+                    element.close()
+                }
+            }
+        },
+    },
+    "dotvvm-model-backdrop-close": {
```
tomasherceg, line 24:
```suggestion
    "dotvvm-modal-backdrop-close": {
```

tomasherceg (commented):
Please do not fire `Closed` event when the value in the ViewModel changes - fire it only when the user makes some action to close the dialog (click on backdrop, press escape, and so on).

tomasherceg (approved):

PR #1777, opened 2024-02-19
Partial TimeSpan support in TextBox

tomasherceg:
I have added partial support for `TimeSpan` (basically redirecting to `TimeOnly` and hoping it will work at least on values between 0:00 and 23:59.59). 
The `TimeSpan` type is weird and uses a different syntax for format strings, so there may be issues in some cases. However, the basic scenario (using it in TextBox) works.


tomasherceg:
TimeSpan format strings work in a different way than DateTime (do not use AM/PM, support more than one day), so this needs to be implemented in another way.

PR #1807, opened 2024-04-18
Use grouped operation name in application insights telemetry

mirecad:
Application insights request telemetry on DOTVVM pages is identified by request URL. If there are path variables app insights considers them as different operations and does not apply grouping. Results are hard to read with tables consisted of many rows pointing to the same route.

image

I was inspired by Web API where operation name ignores variables and used DOTVVM route URL as new name. I was also considering DOTVVM route name, but from my experience URL is always provided by problem reporter (imagine non IT users as reporters) but route name can be more problematic to guess at first look. 

image

As you can see on second screenshot "vyuctovanie" was grouped and I can see overal behavior of this page.


exyi (commented):
Thanks! I agree about using the URL template instead of route name (I'll consider including them in our OpenTelemetry metric labels too)

tomasherceg (approved):

tomasherceg:
Many thanks for the contribution!

PR #1817, opened 2024-05-08
Use relative paths in automatic control registration

exyi:


quigamdev:
Why is this PR labeled as bug? Why would be absolute paths considered bug?

quigamdev (approved):

exyi:
I marked it as a bug, since it lead to inconsistent behavior on different machines (when people/servers have the project in different directory). I agree that it's a borderline bug, and we are not going to backport it to 4.2, as it might in theory break someones code.

PR #1819, opened 2024-05-13
Data pager extensibility improvements

Mylan719:
We should make it possible to implement:
- `aria-label`s 
- custom classes on links
- reader only spans inside navigation and page number buttons
I implemented extensibility points


exyi (approved):

tomasherceg:
I believe this should have the Version 5.0 milestone.

exyi:
@tomasherceg You are right, thanks for pointing it out

PR #1823, opened 2024-05-28
Allow type=module in InlineScriptResource

exyi:
1. type=module scripts are deferred and allow inline content, allowing us to bypass the base64 trick
2. it is necessary if someone wants to import other modules from the inline script

Unfortunatelly, we cannot make it the default when defer is requested, as type=module also enabled JS strict mode.


tomasherceg (approved):
Fix unit tests

PR #1824, opened 2024-05-29
Added support for localizable routes

tomasherceg:
This PR adds support to define multiple route URLs for the same route that will apply in different request cultures.
A default route URL, which is applied for any unmatched language, is required. 
The primary motivation for this is using the same route name for all language versions of the same page. Without that, the route name for each language would have to be different and thus dynamically computed for all `RouteLink` controls in the application).

```
config.RouteTable.Add("FeatureSamples_Localization_LocalizableRoute", "FeatureSamples/Localization/LocalizableRoute/{lang?}", "Views/FeatureSamples/Localization/LocalizableRoute.dothtml",
    localizedUrls: new LocalizedRouteUrl[] {
            new("cs-CZ", "cs/FeatureSamples/Localization/lokalizovana-routa"),
            new("de", "de/FeatureSamples/Localization/lokalisierte-route"),
    });
```

The `RouteLink` control was extended with the `Culture` property to generate a link to the specified route for a different culture (to allow easy language switching).

The route matching assumes that the request culture is already set. 
Detection of the language must be done at the beginning of the request pipeline:
* In ASP.NET Core, the easiest way is to use the standard Request Localization middleware.
* In OWIN, there is no built-in mechanism. It is sufficient to write a middleware, which sets `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture`, but you must also call `context.Set(HostingConstants.OwinDoNotSetRequestCulture, true);` to prevent DotVVM from overwriting the current culture to `configuration.DefaultCulture`.

`LocalizablePresenter` is not supported for this feature as it detects the request culture too late.

Also, the localization is currently unavailable for URL segments defined in route groups.

### DIFF
--- a/src/Framework/Framework/Hosting/HostingConstants.cs
+++ b/src/Framework/Framework/Hosting/HostingConstants.cs
@@ -28,5 +28,7 @@ namespace DotVVM.Framework.Hosting
         public const string DotvvmFileUploadAsyncHeaderName = "X-DotVVM-AsyncUpload";
 
         public const string HostAppModeKey = "host.AppMode";
+
+        public const string OwinDoNotSetRequestCulture = "OwinDoNotSetRequestCulture";
     }
 }
--- a/src/Framework/Framework/Hosting/HostingConstants.cs
+++ b/src/Framework/Framework/Hosting/HostingConstants.cs
@@ -29,6 +29,11 @@ namespace DotVVM.Framework.Hosting
 
         public const string HostAppModeKey = "host.AppMode";
 
+        /// <summary>
+        /// When this key is set to true in the OWIN environment, the request culture will not be set by DotVVM to config.DefaultCulture.
+        /// Use this key when the request culture is set by the host or the middleware preceding DotVVM.
+        /// See https://github.com/riganti/dotvvm/blob/93107dd07127ff2bd29c2934f3aa2a26ec2ca79c/src/Samples/Owin/Startup.cs#L34
+        /// </summary>
         public const string OwinDoNotSetRequestCulture = "OwinDoNotSetRequestCulture";
     }
 }
--- a/src/Framework/Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
+++ b/src/Framework/Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
@@ -38,10 +38,9 @@ namespace DotVVM.Framework.Hosting.Middlewares
             return false;
         }
 
-        public static RouteBase? FindMatchingRoute(IEnumerable<RouteBase> routes, IDotvvmRequestContext context, out IDictionary<string, object?>? parameters, out bool isPartialMatch)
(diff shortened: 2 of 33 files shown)


src/Framework/Framework/Routing/LocalizedDotvvmRoute.cs
```diff
@@ -0,0 +1,156 @@
...
+            .ToHashSet();
+
+        private readonly SortedDictionary<string, DotvvmRoute> localizedRoutes = new();
+
+        public override string UrlWithoutTypes => GetRouteForCulture(CultureInfo.CurrentUICulture).UrlWithoutTypes;
+
+        /// <summary>
+        /// Gets the names of the route parameters in the order in which they appear in the URL.
+        /// </summary>
+        public override IEnumerable<string> ParameterNames => GetRouteForCulture(CultureInfo.CurrentUICulture).ParameterNames;
```
exyi, line 35: We should enforce that the parameter names are the same in all languages

src/Framework/Framework/Hosting/Middlewares/DotvvmRoutingMiddleware.cs
```diff
@@ -53,12 +53,35 @@ private static bool TryParseGooglebotHashbangEscapedFragment(string queryString,
...
             foreach (var r in routes)
             {
-                if (r.IsMatch(url, out parameters)) return r;
+                if (r.IsMatch(url, out parameters))
+                {
+                    isPartialMatch = false;
+                    return r;
+                }
+
+                if (partialMatch == null
```
exyi, line 68: I think that we can speed this up by first trying IsMatch on all routes and only if that fails, try the partial matching. You can also add a second list `partialMatchRoutes` filtered only for the `IPartialMatchRoute` type

src/Framework/Framework/Routing/DotvvmRouteTable.cs
```diff
@@ -234,11 +256,22 @@ public void Add(string routeName, RouteBase route)
...
+            ThrowIfFrozen();
+            partialMatchHandlers.Add(handler);
+        }
+
         public bool Contains(string routeName)
         {
             return dictionary.ContainsKey(routeName);
         }
 
+        public bool TryGetValue(string routeName, out RouteBase? route)
```
exyi, line 270:
```suggestion
        public bool TryGetValue(string routeName,  [MaybeNullWhen(false)] out RouteBase? route)
```

src/Framework/Framework/Hosting/HostingConstants.cs
```diff
@@ -28,5 +28,7 @@ public class HostingConstants
         public const string DotvvmFileUploadAsyncHeaderName = "X-DotVVM-AsyncUpload";
 
         public const string HostAppModeKey = "host.AppMode";
+
+        public const string OwinDoNotSetRequestCulture = "OwinDoNotSetRequestCulture";
```
exyi, line 32: This should maybe have a doccomment

PR #1837, opened 2024-07-12
DataPager styling & templating

tomasherceg:
`DataPager` did not have a way to provide custom templates for page numbers, and there was no way to apply CSS classes or other attributes on inner `<li>` and `<a>` / `<span>` elements.

> However, there is a bug here - when I use `ListItemClass` instead of `ListItemclass`, the attribute is rendered as `Class` and is not combined with Knockout JS binding `css: { "class": something }`. 
> 
> It can be reproduced in `ControlSamples/DataPager/DataPagerTemplates` sample by changing the casing.

### DIFF
--- a/src/Framework/Framework/Controls/DataPager.cs
+++ b/src/Framework/Framework/Controls/DataPager.cs
@@ -94,6 +94,20 @@ namespace DotVVM.Framework.Controls
             DotvvmProperty.Register<ITemplate?, DataPager>(c => c.NextPageTemplate, null);
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


tomasherceg:
I tried to track down the issue with `ListItemclass` vs `ListItemClass`, and it seems that `HtmlWriter` doesn't mind when it receives two attributes (`class` and `Class`) that differ only by case. 

It happily renders both of them, and the browser interprets only the first one.

exyi (approved):

PR #1838, opened 2024-07-14
.NET Core projects updated to .NET 6

tomasherceg:
Since everything below .NET 6 is out of support, I upgraded all projects targeting .NET Core to .NET 6.


exyi (approved):

tomasherceg:
The failed test is flaky - works locally.

PR #1839, opened 2024-08-01
Helper methods to build command binding expressions

tomasherceg:
The `bindingCompilationService.Cache` already had methods to build `value` and `staticCommand` bindings. 
I have added methods for creating `command` bindings as well.

Does `resource` binding also make sense here?


exyi (approved):
I'd also include the methods for Resource bindings.

The reason why these methods aren't currently present is that command and resource binding can be created from a plain System.Func, without the need for JS translation.

PR #1840, opened 2024-08-01
Added AlternateCultureLinks control

tomasherceg:
Since we have localizable routes, we are able to automatically generate the list of `<link rel=alternate>` elements, as described at https://developers.google.com/search/docs/specialty/international/localized-versions#html


src/Framework/Framework/Controls/AlternateCultureLinks.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using DotVVM.Framework.Configuration;
+using DotVVM.Framework.Hosting;
+using DotVVM.Framework.Routing;
+
+namespace DotVVM.Framework.Controls
+{
+    public class AlternateCultureLinks : CompositeControl
```
exyi, line 12: I'd include a doccoment with a reference to the Google/MDN documentation for link rel=alternate

exyi (approved):

PR #1843, opened 2024-08-12
view compiler: avoid blocking the ThreadPool in the AfterApplicationStart mode

exyi:
* in all cases, we now limit the paralelism using a semaphore instead of pushing 500 tasks onto the ThreadPool
* in AfterApplicationStart mode, paralelism is limited to cpu_count/2

### DIFF
--- a/src/Framework/Framework/Compilation/DotvvmViewCompilationService.cs
+++ b/src/Framework/Framework/Compilation/DotvvmViewCompilationService.cs
@@ -11,6 +11,7 @@ using DotVVM.Framework.Hosting;
 using DotVVM.Framework.Utils;
 using DotVVM.Framework.Binding.Properties;
 using DotVVM.Framework.Testing;
+using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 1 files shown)


src/Framework/Framework/Compilation/DotvvmViewCompilationService.cs
```diff
@@ -135,11 +148,22 @@ public async Task<bool> CompileAll(bool buildInParallel = true, bool forceRecomp
...
         }
 
-        private async Task ExecuteCompileTasks(Action[] compileTasks, bool buildInParallel)
+        private static async Task ExecuteCompileTasks(Action[] compileTasks, int maxParallelism)
         {
-            if (buildInParallel)
+            if (maxParallelism > 1)
             {
-                await Task.WhenAll(compileTasks.Select(Task.Run));
+                var semafor = new SemaphoreSlim(maxParallelism);
```
tomasherceg, line 155: Rename to `semaphore`

tomasherceg (approved):

PR #1863, opened 2024-09-20
capabilities: fix assignment to default value of non-standard DotvvmP…

exyi:
…roperty

There was an inverted condition when compared to SetValueDirect. In most cases, SetValueDirect is used, except when the property is a DotvvmProperty with overriden GetValue/SetValue or is inherited


tomasherceg (approved):

tomasherceg:
Please fix the tests.

Issue #1868, opened 2024-10-08
ValidationContext is missing ServiceProvider


MichalTichy:
Hello,

Could it be possible to initialize the service provider in the ValidationContext? This would be especially helpful when working with complex validation attributes.

You can find the relevant code here: [ViewModelValidator](https://github.com/riganti/dotvvm/blob/0df511784dbe7160ef4d9554500c6a53689a5e29/src/Framework/Framework/ViewModel/Validation/ViewModelValidator.cs#L76)

The initialization can be done by calling [InitializeServiceProvider](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/ValidationContext.cs#L235)

Thank you!

PR #1869, opened 2024-10-08
Initialized service provider in validation context.

MichalTichy:
This resolves issue reported in #1868.

This change allows users to use complex validation attributes that need access to various services via service provider.


MichalTichy:
I've added simple test.

exyi (approved):
Ok, thanks for the patch.

Note that the service provider will only work with Singleton services, to allow Scoped and Transient services to work correctly, we'd have to use `context.Services` instead of `config.Services`. However, that isn't possible without changing the IViewModelValidator interface, I'd consider doing it later if we want that.

I'll release this one as a 4.3 patch, so you should be able to use the patch soon

PR #1870, opened 2024-10-16
Fix ko.observable handling in Linq translations

exyi:
* FirstOrDefault, LastOrDefault, ... should behave like indexers
    - return observable, if the array contains observables, which then needs to be unwrapped
    - this is a regression in 4.3
* Where, Order, Take, ... return the same structure which the original array has
* Select is weird as it unwraps all observables in the array, but not recursively. That's why patch adds the JsObjectObservableMap object, which can represent this mid-state (and many other combinations)
   - Indexing after Select was apparently always broken in DotVVM

### DIFF
--- a/src/Framework/Framework/Compilation/Javascript/JavascriptTranslator.cs
+++ b/src/Framework/Framework/Compilation/Javascript/JavascriptTranslator.cs
@@ -261,19 +261,25 @@ namespace DotVVM.Framework.Compilation.Javascript
         }
     }
 
+    /// <summary> Specified if properties of the current object are wrapped in knockout observables or not. Observability of the current value is not specified in this annotation, but using <see cref="ResultIsObservableAnnotation" />. </summary>
     public sealed record JsObjectObservableMap: IEquatable<JsObjectObservableMap>
     {
-        /// <summary> The default value for all descendants, unless overriden using the other properties </summary>
+        /// <summary> The default value for all descendants, unless overridden using the other properties. If null, the default nullability for the current expression will be used. </summary>
         public bool? ContainsObservables { get; set; }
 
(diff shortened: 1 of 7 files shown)


src/Framework/Framework/Compilation/Javascript/JavascriptTranslator.cs
```diff
@@ -212,32 +218,230 @@ public ViewModelInfoAnnotation(Type type, bool isControl = false, BindingExtensi
...
+        public bool? ContainsObservables { get; set; }
+
+        /// <summary> Default value for all properties. Null means that `ContainsObservables` is assumed to be the same as the parent's </summary>
+        public JsObjectObservableMap? DefaultChild { get; set; }
+
+        public Dictionary<string, JsObjectObservableMap>? ChildObjects { get; set; }
+
+        public Dictionary<string, bool>? PropertyIsObservable { get; set; }
+
+        public bool? IsPropertyObservable(string? name) => name is {} && PropertyIsObservable?.TryGetValue(name, out var value) == true ? value : ContainsObservables;
```
tomasherceg, line 245: Add doc comment about meaning of `null`

src/Framework/Framework/Hosting/DotvvmPresenter.cs
tomasherceg, line 1: Please move to a separate PR.

src/Samples/Tests/Tests/Control/TextBoxTests.cs
tomasherceg, line 1: Move this and tests-related change to a separate branch so we can cherry-pick it to 4.3 also.

tomasherceg (approved):

PR #1874, opened 2024-10-20
UI tests tolerate whispace changes in datetime format

exyi:
ICU update probably caused different whitespace
being generated before AM/PM (no whitespace,
&nbsp, and simple space are all possible).


exyi:
Seems like github actions got their sh*t together and this is no longer necessary

tomasherceg (approved):

PR #1877, opened 2024-10-26
Fixed ambiguous namespace resolution and matching non-public types

### LINKED ISSUE #1876: Binding translator should ignore non-public classes and report ambiguous imports
I have a resource file called `Strings.resx` and wanted to use it from DotHTML pages:

```
@import MyApp.Resources

{{resource: Strings.SomeResourceKey}}
```

Unfortunately, there is a global import for `System.Linq`, and an internal class called `System.Linq.Strings` is matched by DotVVM.

I believe that bindings should use only public types. Also, we should produce an exception if there are multiple matches.

tomasherceg:
Resolves #1876


src/Framework/Framework/Compilation/Binding/TypeRegistry.cs
```diff
@@ -34,8 +35,13 @@ public TypeRegistry(CompiledAssemblyCache compiledAssemblyCache, ImmutableDictio
             {
                 return expr;
             }
-            expr = resolvers.Select(r => r(name)).FirstOrDefault(e => e != null);
-            if (expr != null) return expr;
+            var matchedExpressions = resolvers.Select(r => r(name)).Where(e => e != null).Distinct(ExpressionTypeComparer.Instance).ToList();
```
exyi, line 38: <span title="unimportant suggestion">:information_source:</span> We have the `WhereNotNull` extension method, which will allow you to make the ExpressionTypeComparer non-nullable

src/Framework/Framework/Compilation/Binding/TypeRegistry.cs
```diff
@@ -56,7 +62,7 @@ public TypeRegistry AddSymbols(IEnumerable<Func<string, Expression?>> symbols)
         [return: NotNullIfNotNull("type")]
         public static Expression? CreateStatic(Type? type)
         {
-            return type == null ? null : new StaticClassIdentifierExpression(type);
+            return type is { IsPublic: true } or { IsNestedPublic: true } ? new StaticClassIdentifierExpression(type) : null;
```
exyi, line 65: IsNestedPublic is annoying, because it will AFAIK return true for `C2` in the following case:

```
internal class C1 {
   public class C2 { }
}
```

I think it would be worth it to also check the chain of declaring types.

exyi (commented):
10 lines, 10 comments :sweat_smile: Anyway, the public type check might be a bit annoying, if someone has all viewmodels internal, which is currently allowed. However, it should be quite simple to migrate and the compilation page will show all the errors, so I think this is a good change for version 5. Alternatively, we could add a list of "InternalsVisibleFrom" assemblies to DotvvmConfiguration, but I guess it is not really worth the added complexity.

I don't know why are the tests failing... more specifically, why  the culture whitespace thing stopped being a problem and why tests which don't even exists are failing? :facepalm:

exyi (approved):

PR #1878, opened 2024-10-26
Binding translation issue in PostBackHandler properties

tomasherceg:
Referencing `_parent` in properties of postback handlers is translated to an incorrect JS:

```
DotHTML:
<dot:SuppressPostBackHandler Suppress="{value: _parent.Enabled}" />

Translated JS:
(c,d) => $parent.Enabled
```

### DIFF
--- a/src/Framework/Framework/Controls/KnockoutHelper.cs
+++ b/src/Framework/Framework/Controls/KnockoutHelper.cs
@@ -355,8 +355,8 @@ namespace DotVVM.Framework.Controls
             {
                 case IValueBinding binding: {
                     var adjustedCode = binding.GetParametrizedKnockoutExpression(handler, unwrapped: true).AssignParameters(o =>
-                        o == JavascriptTranslator.KnockoutContextParameter ? new ParametrizedCode("c") :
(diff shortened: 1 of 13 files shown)


src/Framework/Framework/Controls/KnockoutHelper.cs
```diff
@@ -355,8 +355,8 @@ private static JsExpression TransformOptionValueToExpression(DotvvmBindableObjec
             {
                 case IValueBinding binding: {
                     var adjustedCode = binding.GetParametrizedKnockoutExpression(handler, unwrapped: true).AssignParameters(o =>
-                        o == JavascriptTranslator.KnockoutContextParameter ? new ParametrizedCode("c") :
-                        o == JavascriptTranslator.KnockoutViewModelParameter ? new ParametrizedCode("d") :
+                        o == JavascriptTranslator.KnockoutContextParameter ? CodeParameterAssignment.FromIdentifier("c") :
```
exyi, line 358: FYI: This is just a cosmetic change - the `FromIdentifier` correctly sets operator precedence, so the `c`/`d` variables don't get wrapped in unnecessary parenthesis.

exyi (approved):

PR #1881, opened 2024-11-03
Refactoring & cleanup in routes & route groups + bug fixes

tomasherceg:
We had several issues in route groups when used in combination with URL and route redirections:

```csharp
config.RouteTable.AddGroup(
    "CmsAdmin", urlPrefix, "embedded://Shared.Infrastructure.Cms.Admin", routes =>
{
    routes.AddRouteRedirection(
        "Index", 
        "", // this URL was not concatenated with the group prefix, 
        "CmsAdmin_Articles");
    routes.Add("Preview", 
        "articles/preview", 
        "some_valid_virtual_path", // this cannot be null and requires a valid embedded resource path even though it is not used 
        presenterFactory: _ => new PreviewPopupPresenter());
});
```

### DIFF
--- a/src/Framework/Framework/Routing/RouteBase.cs
+++ b/src/Framework/Framework/Routing/RouteBase.cs
@@ -1,10 +1,6 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
 using DotVVM.Framework.Hosting;
-using System.Reflection;
-using System.Collections.ObjectModel;
 using DotVVM.Framework.Configuration;
 using System.Diagnostics.CodeAnalysis;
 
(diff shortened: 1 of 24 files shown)


exyi (approved):

src/Framework/Framework/Routing/RouteBase.cs
```diff
@@ -203,11 +194,13 @@ public static void AddOrUpdateParameterCollection(IDictionary<string, object?> t
...
         }
 
-
         /// <summary>
         /// Returns a presenter that processes the request.
         /// </summary>
-        public abstract IDotvvmPresenter GetPresenter(IServiceProvider provider);
+        public virtual IDotvvmPresenter GetPresenter(IServiceProvider provider) => PresenterFactory(provider);
+
+        public Dictionary<string, object?> CloneDefaultValues() => new Dictionary<string, object?>(_defaultValues, StringComparer.OrdinalIgnoreCase);
```
exyi, line 202: [nitpick] Could be worth it to push the cloning to FreezableDictionary, then we could use Dictionary.Comparer to make sure we use the same one as the original dict, and calling new Dictionary(dict) is faster if `dict.GetType()==typeof(System.Collection.Generic.Dictionary<...>)`

PR #1886, opened 2024-11-19
Fix Repeater memoization memory leak

exyi:
If the ConditionalWeakTable references itself, it will get leaked by .NET Core GC: https://github.com/dotnet/runtime/issues/12255

Fortunately, we don't really need the ConditinalWeakTable here:
* Repeater should generally be a short-lived object, so strong references will also be collected reasonably soon
* Between Load and PreRender, the viewmodels were pinned in the DataContext of all children and the children are in Children collection. Only after PreRender could those references be useful, but we can as well Clear the dictionary at that point.

### DIFF
--- a/src/Tests/Runtime/RepeaterMemoizationTest.cs
+++ b/src/Tests/Runtime/RepeaterMemoizationTest.cs
@@ -1,0 +1,157 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using AngleSharp.Io;
+using DotVVM.Framework.Binding.Expressions;
+using DotVVM.Framework.Compilation.ControlTree;
+using DotVVM.Framework.Controls;
+using DotVVM.Framework.Controls.Infrastructure;
+using DotVVM.Framework.Hosting;
+using DotVVM.Framework.Testing;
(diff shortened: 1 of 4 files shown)


src/Tests/Runtime/RepeaterMemoizationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+        [DataTestMethod]
+        [DataRow(DotvvmRequestType.Navigate, RenderMode.Client)]
+        [DataRow(DotvvmRequestType.Navigate, RenderMode.Server)]
+        [DataRow(DotvvmRequestType.SpaNavigate, RenderMode.Client)]
+        [DataRow(DotvvmRequestType.SpaNavigate, RenderMode.Server)]
+        [DataRow(DotvvmRequestType.Command, RenderMode.Client)]
+        [DataRow(DotvvmRequestType.Command, RenderMode.Server)]
+        public void Repeater_IsCollectible(DotvvmRequestType requestType, RenderMode renderMode)
+        {
+            var (repeaterRef, contextRef, viewModelRef) = NewMethod(requestType, renderMode);
```
tomasherceg, line 128: I am a bit uncomfortable with the method name `NewMethod` :D

tomasherceg (approved):

PR #1890, opened 2024-12-01
Drop support for 4.1 and 4.0, provide clear support timelines for other versions

exyi:
Update the security policy:

* deprecate 4.0 and 4.1 releases, we will no longer guarantee security fixes
* promise that users will always have at least 1 year to upgrade
* old stable versions now have support expiration date in the table

In practice we will backport fixes of critical vulnerabilities even for unsupported versions, but don't need to promise that.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,7 @@
 
 ## Supported Versions
 
-When it comes to security updates, we always support the current major version together with the latest preview version.
+When it comes to security updates, we always support the current stable version together with the latest preview version.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -3,14 +3,15 @@
 ## Supported Versions
 
 When it comes to security updates, we always support the current major version together with the latest preview version.
```
tomasherceg, line 5:
```suggestion
When it comes to security updates, we always support the current stable version together with the latest preview version.
```

tomasherceg (approved):

PR #1893, opened 2024-12-15
Fix leaking of original object from ko.observableArray

exyi:
ko.observableArray would leak the original object to its 'arrayChanges' subscribers when instance method (such as 'push', 'splice') was used to modify the array. Knockout would pre-compute the diff passed to the subscribers before our validator was called.

* Knockout updated to b7b150a6 to remove an optimization which was the primary cause
* Made FakeObservableObject internal properties non-enumerable so make sure they don't propagate through coercer when assigned into an array.


src/Framework/Framework/Resources/Scripts/tests/stateManagement.test.ts
```diff
@@ -692,3 +692,106 @@ test("state is frozen", () => {
...
+// test("ko.observableArray - clones pushed dotvvm observable taken from elsewhere", () => {
+//     setupValidatingArraySubscriptions((arr: any, initial) => {
+//         arr.push({ Id: 1 })
+//         arr.push(vm.Array().at(-1))
+//         arr().at(-1)().Id(2)
+
+//         expect(arr.state).toEqual([...initial, { Id: 1, $type: "t2" }, { Id: 2, $type: "t2" }])
+//         expect(arr.state).toEqual(ko.toJS(arr))
+//     })
+// })
```
exyi, line 797: This case still doesn't work and will be fixed in a subsequent PR.

I experimented with possible solutions and the one I like the most is to freeze the ko.observableArray underlying array and clone it on every operation.
The implementation in knockout.js is easy, it will slightly simplify our array handling logic, but it is a breaking change:

Currently, it is possible to call `array().push(1)`. The behavior will most likely be very broken, as it does not go through the validator and does not notify subscribers. In some cases, it might however work correctly.

If we make the array frozen, it will throw an exception. That is definitely desired, but I'd prefer to not push this as a 4.3 patch

tomasherceg: My guess is that in 99% of cases, directly modifying the underlying array is a bug and something people do intentionally. I do not insist on having it in 4.3, but I expect it will uncover a few bugs in Business Pack.

src/Framework/Framework/Resources/Scripts/tests/stateManagement.test.ts
```diff
@@ -692,3 +692,106 @@ test("state is frozen", () => {
...
+        arr.push({ Id: 1 })
+        arr.push(vm.Array().at(-1)())
+        arr().at(-1)().Id(2)
+
+        expect(arr.state).toEqual([...initial, { Id: 1, $type: "t2" }, { Id: 2, $type: "t2" }])
+        expect(arr.state).toEqual(ko.toJS(arr))
+    })
+})
+
+// TODO: does not work yet
```
tomasherceg, line 787: Shouldn't we also test that it behaves correctly when accessing the value through the observable? We only look at the contents via the `state` property.

PR #1895, opened 2024-12-20
v5 cleanup: Move HTTP requests to resources, file upload/return under /_dotvvm/

### LINKED ISSUE #1539: Move all dotvvm internal routes to the `/_dotvvm/` prefix
Currently, we internally use a lot of different endpoints (dotvvmResource, something for file upload, returned files, ...). The new things, like configuration page are under `/_dotvvm` prefix, so it would make sense to move everything else into that directory too.

This will make it easier to whitelist all dotvvm requests, avoid conflicts on websites with a `/{username}` route (like github, twitter does), and feel less adhoc in general.

It's breaking, there are definitely people hardcoding dotvvmResource somewhere, but I don't think it's common

exyi:
resolves #1539 (unless I forgot about some requests we do)

### DIFF
--- a/src/Framework/Framework/Hosting/DotvvmRequestContextExtensions.cs
+++ b/src/Framework/Framework/Hosting/DotvvmRequestContextExtensions.cs
@@ -262,7 +262,7 @@ public static class DotvvmRequestContextExtensions
         var downloadAttribute = attachmentDispositionType == "inline" ? null : fileName;
 
         var generatedFileId = await returnedFileStorage.StoreFileAsync(stream, metadata).ConfigureAwait(false);
-        context.SetRedirectResponse(context.TranslateVirtualPath("~/dotvvmReturnedFile?id=" + generatedFileId), downloadName: downloadAttribute);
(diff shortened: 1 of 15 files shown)


src/Framework/Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -262,7 +262,7 @@ public static async Task ReturnFileAsync(this IDotvvmRequestContext context, Str
         var downloadAttribute = attachmentDispositionType == "inline" ? null : fileName;
 
         var generatedFileId = await returnedFileStorage.StoreFileAsync(stream, metadata).ConfigureAwait(false);
-        context.SetRedirectResponse(context.TranslateVirtualPath("~/dotvvmReturnedFile?id=" + generatedFileId), downloadName: downloadAttribute);
+        context.SetRedirectResponse(context.TranslateVirtualPath("~/_dotvvm/returnedFile?id=" + generatedFileId), downloadName: downloadAttribute);
```
tomasherceg, line 265: Please move to `HostingConstants`

tomasherceg (approved):

Issue #1428, opened 2022-06-30
Content is not allowed inside an HtmlLiteral


cafour:
Currently, DotVVM issues an error when I attempt to do this:

```html
<dot:HtmlLiteral>
<script type="template">
  {{ columns }}
</script>
</dot:HtmlLiteral>
```

This is useful in cases when DotVVM is not used for string interpolation using the `{{ value }}` syntax, but some other client-side templating library is.

The error says _"Content is not allowed inside the property 'HtmlLiteral.Html'! (Conflicting node: Node DothtmlElementNode)"_. This is despite the fact that HtmlLiteral has a `[ControlMarkupOptions(DefaultContentProperty = nameof(Html))]` attibute.

IMHO we should allow hard-coded HTML inside an HtmlLiteral and not just a binding.

EDIT: It has occured to me, that maybe I'm not explaining the sample clearly enough. I want for the `{{ columns }}` string to be passed to the browser without any processing (i.e. binding resolution).

WORKAROUND: A resource binding on the `HtmlLiteral.Html` property to a string const in C# land.

exyi:
Will it work then? I think it will then issue another error that binding `{{ columns }}` could not be parser (or that bindings are not allowed in this context). We'd have to instruct the parser that `dot:HtmlLiteral` is somehow special.

1. If something else is processing the bindings, you probably have XSS in your page, so I don't want to explicitly support this use case.
2. If you are really sure that it's fine, or have some other use case for `{{` than inlining javascript, then use CDATA for it.
3. Also... in `<script type=template>` it should be allowed anyway, script tag AFAIK has this special treatment.

cafour:
@exyi

1. Why would I need to have XSS in my page? The page is simply templated by some library, that is not aware of DotVVM or its viewmodel in any way and merely uses the same syntax in its templates. I just need to pass a constant HTML literal without DotVVM processing in it any way.

2. I am not sure if I'm just not using CDATA right, or it's not that well supported. But for me the browser renders not only the contents of CDATA but also the surrounding `<![CDATA[...]>` tag. Therefore does not seem to be a fix.

3. What should be allowed in `<script type=template`? If I try to use it without any escapes, DotVVM simply tells me that requirements for that binding could not be met.

I've found a workaround: passing the string constant into the `<dot:HtmlLiteral> through a resource binding. But this just seems like too much hassle just to escape some dumb `{{`.

exyi:
[...]

PR #1900, opened 2024-12-28
Special-case parsing of <script> and <style> elements, similarly to HTML

### LINKED ISSUE #1445: Fix style / script parsing
`style` and `script` tags have special treatment in HTML - they contain verbatim text, not any nested elements. We should parse the contents as an unencoded text, simply looking for end: `<script\b` (case insensitive).

We need to do this in a major version, since it's a breaking change (for example, someone may rely on {{resource:X}} working in script).

We should consider giving the same parsing exception to `dot:InlineScript`

exyi:
In HTML, the content of <script> and <style> tags should not be parsed as HTML, the parser should simply look for the end tag.
This eliminates the need to HTML-encode all `<` operators (or even HTML inlined in string literals).

To align dothtml and HTML, the patch implements this behavior in dothtml. The change may easily break someone's code, if they already have a script element with entities like &lt;, so it is possible to configure which tags will be parsed as "raw text".
By default, it is script, style and also dot:InlineScript and dot:HtmlLiteral (as suggested in #1428). This setting is up for debate.

resolves #1445

### DIFF
--- a/src/Framework/Framework/Compilation/Parser/Dothtml/DotvvmSyntaxConfiguration.cs
+++ b/src/Framework/Framework/Compilation/Parser/Dothtml/DotvvmSyntaxConfiguration.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using DotVVM.Framework.Configuration;
+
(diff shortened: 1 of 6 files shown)


tomasherceg (approved):
Fix failing tests because of newlines

PR #1902, opened 2025-01-03
Timer control added

tomasherceg:
Usage:

```
<dot:Timer Interval="1000" Command="{command: Value1 = Value1 + 1}" Enabled="{value: Enabled1}" />
```

### DIFF
--- a/src/Framework/Framework/Resources/Scripts/binding-handlers/timer.ts
+++ b/src/Framework/Framework/Resources/Scripts/binding-handlers/timer.ts
@@ -1,0 +1,42 @@
+﻿type TimerProps = {
+    interval: number,
+    enabled: KnockoutObservable<boolean>,
+    command: () => Promise<DotvvmAfterPostBackEventArgs>
+}
+
+ko.virtualElements.allowedBindings["timer"] = true;
+
+export default {
+    "timer": {
+        init: (element: HTMLElement, valueAccessor: () => TimerProps) => {
+            const prop = valueAccessor();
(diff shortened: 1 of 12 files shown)


src/Framework/Framework/Resources/Scripts/binding-handlers/timer.ts
```diff
@@ -0,0 +1,48 @@
...
+
+ko.virtualElements.allowedBindings["dotvvm-timer"] = true;
+
+export default {
+    "dotvvm-timer": {
+        init: (element: HTMLElement, valueAccessor: () => TimerProps) => {
+            const prop = valueAccessor();
+            let timer: number | null = null;
+
+            if (ko.isObservable(prop.enabled)) {
```
exyi, line 15: `enabled` can change even if the value isn't observable. For instance, it will happen if you set it to `Enabled={value: State == 2}`. There are two options to solve it

* Use the `binding.WrappedKnockoutExpression` property for the Enabled property in the DotvvmControl
* Add `update` handler
* Remove the condition here, and use `ko.pureComputed(() => ko.unwrap(valueAccessor().enabled).subscribe(...)`

The ko.pureComputed option is probably the easiest.

exyi: Also, the subscription needs to be disposed. In this case, we cancel the timer, but it would get re-enabled if control is removed, and then Enabled changes.

exyi (approved):

PR #1910, opened 2025-02-19
MultiSelect issue with updating observable array fixed

tomasherceg:
`MultiSelect` was not updating the selections if the number of elements in the observable array remained the same and only its elements were exchanged. I updated Knockout's `selectedOptions` binding to touch all observables in the array (shallow) to ensure the handler is called even when some of the array elements gets updated.


src/Framework/Framework/Resources/Scripts/knockout-latest.debug.js
```diff
@@ -5117,6 +5117,9 @@ ko.bindingHandlers['selectedOptions'] = {
                 previousScrollTop = element.scrollTop;
 
             if (newValue && typeof newValue.length == "number") {
+                // touch all array elements
```
exyi, line 5120:
```suggestion
                // touch all array elements, because the ko.utils.arrayIndexOf bellow unwraps everything with ko.dependencyDetection.ignore
```

exyi (approved):

PR #1922, opened 2025-04-09
Enhancements of localizable routes

tomasherceg:
When using the localizable routes feature, I discovered two unpleasant things:

* Route groups did not have support localized route fragments.

* The `IsPartialMatch` method did not return which culture was matched. I needed that when using ASP.NET Core's Request Localization feature to be able to determine the culture from the URL.


tomasherceg:
I am not sure if we want to merge it in 4.3 because of binary breaking changes (I've added an optional parameter to the 
`AddGroup` method), but I need the feature in one project that may not be ready for migration to DotVVM 5.

exyi (approved):
:+1: 

I'd avoid merging this to 4.3. If we need it in the 4-something version, we can release some 4.4-preview version

tomasherceg:
I changed this PR to go to 5.0 to prevent confusion. 
Please do not merge yet - I want to add tests for the request culture provider I implemented.

PR #1929, opened 2025-04-26
Do not hide SPA page while JS is loading

exyi:
Should improve "visual" loads times, and SEO.

Hiding the page is currently unnecessary
it was added back in 2015 (9c11159), because we used to rely on the URL path fragment (`#/something`), which could be different from the page which was currently being loaded.


tomasherceg:
Seems like this change broke down several SPA-related tests.

PR #1937, opened 2025-06-01
Better support for request cancellation: add context.RequestAborted

exyi:
* New API makes it easier (and faster) to obtain the CancellationToken in user code. Since the token should generally be propagated to all async methods, it can simplify a lot of code.
* The property name is the same as in Asp.Net Core
* DotvvmPresenter will check the RequestAborted token in strategic places and stop executing the request:
   * before each lifecycle event is called on the control tree
   * before command execution
   * in error handling: user exception handling is disabled, except for Dispose methods


tomasherceg:
We should remove the `IRequestCancellationTokenProvider` interface and its implementations.

tomasherceg (approved):

PR #1940, opened 2025-06-13
Fix of several flaky UI tests

tomasherceg:


src/Framework/Framework/Resources/Scripts/postback/http.ts
```diff
@@ -51,26 +51,38 @@ export async function fetchJson<T>(url: string, init: RequestInit): Promise<Wrap
...
+            return await response.text();
+        })();
+
         try {
-            const url = addLeadingSlash(concatUrl(getVirtualDirectory() || "", "_dotvvm/csrfToken"));
-            response = await fetch(url, { signal });
+            token = await csrfTokenFactory;
+            getStateManager().setState({ ...getState(), $csrfToken: token })
         }
         catch (err) {
```
exyi, line 82: this needs to be a `finally` block, we always have to clear the cache. Otherwise, we run into issues when the `$csrfToken` is invalidated due to invalid token response from the server: this cache isn't invalidated, so instead of re-requesting a new token, we keep trying with the old one.

PR #1941, opened 2025-06-29
Generating sitemap.xml from route table

tomasherceg:
I added a mechanism to auto-generate [Sitemaps](https://www.sitemaps.org/) based on metadata in the route table.

1. Turning the sitemap on for all routes (except for DotVVM routes that are always excluded, and also routes without virtualPath - they must be included explicitly):

```
config.RouteTable.WithDefaultSitemapOptions(sitemap =>
{
    sitemap.ChangeFrequency = ChangeFrequency.Daily;
    sitemap.Priority = 1;
});
```

2. The settings may be overridden on route groups. Unfortunately, I had to change the return type of `AddGroup`, otherwise there would be no easy way to access the route group. However, I don't think it will affect many people.

```
routeTable.AddGroup(..., opt =>
    {
        ...
    })
    .WithDefaultSitemapOptions(sitemap =>
    {
        sitemap.Exclude = true;
    });
```

3. You can specify a provider to produce all possible values for parameters. The provider must be registered in `IServiceCollection`. It should implement `IRouteParameterValuesProvider`, and return a list of "dictionaries" containing values for each parameter (for each culture, if route localization is used).

```
config.RouteTable.Add("BlogDetail", "blog/{Slug}", ...)
    .WithSitemapOptions(sitemap =>
    {
        sitemap.UseParameterValuesProvider<BlogDetailParameterValuesProvider>();
    });
```

4. The last modified date is by default inferred from the last modification of the entry assembly (which will be the web application DLL). It is possible to override.

5. The routes may be explicitly included or excluded by setting their `sitemap.Exclude` property. You can also fine-tune the `Priority`, `ChangeFrequency`, and `LastModified`.


tomasherceg:
Abandoning in favor of #1953

PR #1943, opened 2025-06-30
Added DoesNotThrow attribute to methods that throw exceptions

tomasherceg:


src/Framework/Framework/Hosting/DotvvmRequestContextExtensions.cs
```diff
@@ -223,24 +230,29 @@ public static string TranslateVirtualPath(string virtualUrl, IHttpContext httpCo
...
     /// </summary>
     [Obsolete("Use ReturnFileAsync() instead")]
+    [DoesNotReturn]
     public static void ReturnFile(this IDotvvmRequestContext context, Stream stream, string fileName, string mimeType, IEnumerable<KeyValuePair<string, string>>? additionalHeaders = null, string? attachmentDispositionType = null) =>
         context.ReturnFileAsync(stream, fileName, mimeType, additionalHeaders, attachmentDispositionType).GetAwaiter().GetResult();
+
     /// <summary>
     /// Redirects the client to the specified file.
     /// </summary>
+    [DoesNotReturn]
```
exyi, line 248: I could not find whether the attribute is valid on async methods, technically it always returns the task instance. However, it shouldn't matter as long as you await it right away

exyi (approved):

PR #1953, opened 2025-08-10
Generating sitemap.xml from route table [refactoring]

tomasherceg:
Better public API for #1941.

I replaced the "route parameter providers" and "last modified date providers" with a new interface `ISitemapEntryProvider`, which gets a list of all routes that need to resolve entries, and can choose which ones it will resolve by calling `AddSitemapEntry` on the particular route.

It is possible to register multiple providers, and the provider can choose which routes it will handle based on `RouteName`, as shown in the example.

```csharp
public class BlogReferenceSitemapEntryProvider(ICmsContentProvider cmsContentProvider) : ISitemapEntryProvider
{
	public Task TryResolveSitemapEntries(IReadOnlyList<RouteSitemapContext> routes, CancellationToken ct)
	{
		foreach (var route in routes)
		{
			if (route.Route.RouteName == Routes.Public.BlogDetail)
			{
				// list all published articles that are older than now
				var now = DateTime.UtcNow;
				var slugs = cmsContentProvider.GetPublishedArticles("blog-article", "en")
					.Where(a => a.Versions[0].TryGetDateTimeField("date") < now)
					.Select(a => a.Slug);

				foreach (var slug in slugs)
				{
					route.AddSitemapEntry(new { Slug = slug });
				}
			}
			else if (route.Route.RouteName == Routes.Public.ReferenceDetail)
			{
				...
			}
		}

		return Task.CompletedTask;
	}
}
```

If the route is localized and the parameter values don't depend on the culture, it can be handled by a single call to `route.AddSitemapEntry`. There is an optional parameter of `Dictionary<string, object>`, which can supply localized parameter values. It is also possible to override the change frequency, priority, etc., for every added entry.


src/Tools/Sitemap/DotVVM.Sitemap/DotVVM.Sitemap.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
tomasherceg, line 4: Use build variable

src/Tools/Sitemap/DotVVM.Sitemap/DotvvmServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public static class DotvvmServiceCollectionExtensions
+{
+    /// <summary>
+    /// Registers the services required for automatic generation of sitemaps for the applications.
+    /// </summary>
+    public static void AddSitemapServices(this IDotvvmServiceCollection services, Action<SitemapOptions>? configureOptions = null)
+    {
+        services.Services.AddScoped<SitemapPresenter>();
+        services.Services.AddOptions<SitemapOptions>().Configure(configureOptions ?? (_ => { }));
```
tomasherceg, line 17: Add `SitemapRouteName` and `SitemapUrl` to options and add the sitemap route automatically.

src/Tools/Sitemap/DotVVM.Sitemap/Options/SitemapOptions.cs
```diff
@@ -0,0 +1,15 @@
...
+public class SitemapOptions
+{
+    /// <summary>
+    /// Gets the site public URL for the sitemap.
+    /// If not set, it will be detected from the HTTP request.
+    /// </summary>
+    public Uri? SitePublicUrl { get; set; }
+
+    /// <summary>
+    /// Gets or sets the default culture for the sitemap.
```
tomasherceg, line 12: Add link to MDN or some other site where this is documented.

PR #1956, opened 2025-09-13
Change dotvvm resource URL to /_dotvvm/resource-{name}/{name}

exyi:
* this is more backward compatible with v4, as relative references like `../../something`  will keep working
* version hash is moved to query string, which is more conventional (and does not cause compatibility issues for a long time already)

### DIFF
--- a/src/Framework/Framework/ResourceManagement/LocalResourceUrlManager.cs
+++ b/src/Framework/Framework/ResourceManagement/LocalResourceUrlManager.cs
@@ -3,49 +3,57 @@ using DotVVM.Framework.Hosting;
 using DotVVM.Framework.Hosting.Middlewares;
 using DotVVM.Framework.Routing;
 using DotVVM.Framework.Utils;
+using Microsoft.Extensions.Logging;
 using System;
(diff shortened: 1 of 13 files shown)


src/Framework/Framework/ResourceManagement/LocalResourceUrlManager.cs
```diff
@@ -59,38 +67,42 @@ protected virtual string DecodeResourceName(string name)
...
                 return null;
             }
 
-            var name = DecodeResourceName((string)parameters![NameParameterName]!);
-            var hash = (string)parameters[HashParameterName].NotNull();
+            var requestedName = (string)parameters!["name"]!;
+            var name = DecodeResourceName(requestedName);
+            var fileName = (string)parameters["fileName"]!;
+            var hash = (string?)context.Query["v"];
             var type = context.Query.TryGetValue("type", out var x) ? x : null;
```
tomasherceg, line 85: The `type` parameter is not used anymore, I'd remove it.

tomasherceg (approved):

Issue #1911, opened 2025-03-03
Add supprot for .NET 9 Static assets to DotVVM resources


holajan:
In .NET 9 is new feature Static assets.

You can replace UseStaticFiles to MapStaticAssets. But it is only part of this feature.
The main goal is to use fingerprinting  and generated  .gz and .br assets (with .staticwebassets.endpoints.json in debug enviroment).

You can use new @Assets property (In Blazor apps) or link tag helpers (MVC & Razor Pages apps) to resolve the assets fingerprinted file names and <ImportMap /> component or <script type="importmap"></script> for JavaScript modules. There is no such functionality for DotVVM apps.
It needs to be added in DotVVM resources registration on UrlResourceLocation or similar class.

Please add this support.

exyi:
Ok, I like the idea and I tried to look into it. However, I did not find a reasonable way to add the support. As far as I understand it, we need to find the `ResourceAssetCollection` somewhere. Razor pages use an internal API for this: https://source.dot.net/#Microsoft.AspNetCore.Mvc.RazorPages/Builder/PageActionEndpointConventionBuilderResourceCollectionExtensions.cs,90ec282fad908e18, so best I currently have is this:


1. Obtain the `ResourceAssetCollection`:

  ```csharp
  var resolverT = Type.GetType("Microsoft.AspNetCore.Components.Endpoints.ResourceCollectionResolver, Microsoft.AspNetCore.Components.Endpoints")!;
  var resolver = Activator.CreateInstance(resolverT, args: [ ep ]); 
  var anyFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
  var resourceCollection = (ResourceAssetCollection)resolverT.GetMethod("ResolveResourceCollection", anyFlags)!.Invoke(resolver, [null])!;
  
  ```
2. store it somewhere (static variable)
3. Add a resource location which can use this:
  ```
  class AssetResourceLocation(string url): IResourceLocation
  {
      public string GetUrl(IDotvvmRequestContext context, string name)
      {
          ResourceAssetCollection assets = ???;
          var versionedUrl = assets[url];
          if (versionedUrl is null || versionedUrl == url)
          {
              throw new Exception($"Asset {url} does not exist?");
          }
  
          return "/" + versionedUrl;
      }
  }
  ```
4. You can then add resources with `config.Resources.Register("something", new ScriptModuleResource(new AssetResourceLocation(assets, "path/to/something.js")));`


[...]

PR #1974, opened 2025-12-28
Integration with ASP.NET Core static assets

exyi:
see #1911

1. DotVVM automatically translates `~/`-prefixed paths to the static asset path (with the hash in name) when used in attributes such as ``; if you don't like that define it yourself using `StaticAssetResourceLocation`


src/Framework/Hosting.AspNetCore/DotVVM.Framework.Hosting.AspNetCore.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFrameworks>$(DotnetCoreTargetVersion)</TargetFrameworks>
+    <TargetFrameworks>net9.0;$(DotnetCoreTargetVersion)</TargetFrameworks>
```
tomasherceg: Maybe upgrade to .NET 10?

PR #1983, opened 2026-01-02
Added DotVVM.Sitemap to list of publishable packages

tomasherceg:


tomasherceg:
Sorry, I was editing the files without Visual Studio and I didn't notice. Should be fixed now.

PR #2004, opened 2026-02-27
Add *Options to IBaseGridViewDataSet instead of I*eableDataSet

exyi:
The options do support a nil implementation, which means it doesn't add any requirements for IBaseGridViewDataset, but it simplifies usage.

On supported platform, we can even make the No*Options the default implementation, so that Items remains to be the single required property.


tomasherceg (commented):
Let's also merge `IBaseGridViewDataSet` and `IRefreshableGridViewDataSet` into `IGridViewDataSet` - it doesn't have much value to separate those.

tomasherceg (approved):

PR #2041, opened 2026-06-12
Added workflow permissions to write checks

tomasherceg:


tomasherceg:
This is not needed, the checks didn't appear because CodeQL was switched on, and it somehow tampered with GitHub UI filtering.

PR #2062, opened 2026-06-29
Add context.IncludeFileAsync to return file without terminating postback

### LINKED ISSUE #431: Allow to redirect or return file after the successful postback
When you call `Context.Redirect*` or `Context.ReturnFile`, the redirect is performed immediately, so the viewmodel cannot be updated before the browser is instructed to make the download.

We should create another set of methods that would be able to redirect or return a file after the postback is finished.

exyi:
Resolves #431

Implemementation is based on customProperties extension point instead of redirect


tomasherceg:
Could we create and remove the `fakeAnchor` instead of keeping it there all the time?

tomasherceg (approved):

PR #2073, opened 2026-07-10
Bump Nodejs to 24 in Github Actions

exyi:


tomasherceg:
We had to pin npm to 11 because of [Cannot find module 'sigstore'](https://github.com/npm/cli/issues/9722) bug. 
They just closed the bug as completed, so we can wait for a fix in npm v12 and revert the last commit.

tomasherceg (approved):

PR #2107, opened 2026-09-05
Config deserialization for VS extension

tomasherceg:
I want to make DotVVM config deserialization work in the framework so the VS extension can use it.

### DIFF
--- a/src/Framework/Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
+++ b/src/Framework/Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
@@ -31,27 +31,32 @@ namespace DotVVM.Framework.ResourceManagement
 
         public override DotvvmResourceRepository? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            throw new NotImplementedException();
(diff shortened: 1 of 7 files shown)


src/Framework/Framework/ResourceManagement/ResourceRepositoryJsonConverter.cs
```diff
@@ -31,27 +31,32 @@ public override bool CanConvert(Type objectType)
...
+                    ?? UnknownResourceType
+                    ?? throw new JsonException($"Resource collection '{resourceGroup.Name}' is not supported.");
+                if (!typeof(IResource).IsAssignableFrom(resourceType))
+                    throw new JsonException($"Resource collection '{resourceGroup.Name}' has an invalid resource type '{resourceType}'.");
+                if (resourceGroup.Value.ValueKind != JsonValueKind.Object)
+                    throw new JsonException($"Resource collection '{resourceGroup.Name}' must be an object.");
+
+                foreach (var resource in resourceGroup.Value.EnumerateObject())
+                {
+                    var value = JsonSerializer.Deserialize(resource.Value.GetRawText(), resourceType, options) as IResource
```
exyi: We should verify `resourceType` is IResource before deserialization

Issue #2001, opened 2026-02-24
DataPager in markup control


tomasherceg:
I created a repro in the branch `bug/datapager-in-control` for the [issue reported on the forum](https://forum.dotvvm.com/t/ipageablegridviewdataset-in-usercontrol/353).

`DataPager` determines the capabilities of the dataset by checking whether the data context implements `IPageableGridViewDataSet<IPagingFirstPageCapability>` and other `*Capability` interfaces, and renders the UI based on those capabilities.

The problem is that the DataContextStack in the control has type `IPageableGridViewDataSet<IPagingOptions>` from the viewmodel directive, but this type doesn't implement those capabilities. Therefore, the [condition](https://github.com/riganti/dotvvm/blob/main/src/Framework/Framework/Controls/GridViewDataSetBindingProvider.cs#L58) evaluates to `false`.

@exyi Is there a way around this? I tried to set the viewmodel to `IPageableGridViewDataSet<PagingOptions>`, but it throws an invalid cast error.

PR #2118, opened 2026-09-20
Repro for DataPager used in markup control

tomasherceg:
See https://github.com/riganti/dotvvm/issues/2001


tomasherceg:
I updated the sample to try both `IPageableGridViewDataSet<IPagingOptions>` and `IPageableGridViewDataSet<PagingOptions>`, but the bug still persists:
* First option doesn't produce any errors, but the pager doesn't render
* Second option produces an error 
[dotvvm-error-ArgumentException.html](https://github.com/user-attachments/files/32441380/dotvvm-error-ArgumentException.html)

The original issue says the bug is fixed, but I cannot see any commits in the branch except for my commit that adds the repro.

exyi:
Sorry, I forgot to push the commit. We fixed a subsequent problem with missing RequestRefresh method in #2004, although it also needs fix of method resolution (`Expression.Call(x, "Name")` doesn't look into inherited interfaces).

> First option doesn't produce any errors, but the pager doesn't render

That is the correct behavior for empty options.

> Second option produces an error

Should be fixed now

